PR #116, opened 2015-03-05
add option for deserializing binary into Buffer object

msmiley:
**promoteBuffers** {Boolean, default:false}

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,7 @@ The API consists of two simple methods to serialize/deserialize objects to/from 
        * **evalFunctions** {Boolean, default:false}, evaluate functions in the BSON document scoped to the object deserialized.
        * **cacheFunctions** {Boolean, default:false}, cache evaluated functions for reuse.
        * **cacheFunctionsCrc32** {Boolean, default:false}, use a crc32 code for caching, otherwise use the string of the function.
+       * **promoteBuffers** {Boolean, default:false}, deserialize Binary data directly into node.js Buffer object.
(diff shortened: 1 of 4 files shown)


christkv:
would need to be also applied to bson-ext module in c++ to be pulled in and include tests.

lib/bson/bson.js
```diff
@@ -1282,8 +1284,13 @@ BSON.deserialize = function(buffer, options, isArray) {
           if(subType == Binary.SUBTYPE_BYTE_ARRAY) {
             binarySize = buffer[index++] | buffer[index++] << 8 | buffer[index++] << 16 | buffer[index++] << 24;
           }
-          // Slice the data
-          object[name] = new Binary(buffer.slice(index, index + binarySize), subType);
+          if(nativeBuffers) {
+            // Slice the data and create a native Buffer object
+            object[name] = new Buffer(buffer.slice(index, index + binarySize));
```
tellnes: `buffer.slice()` returns a buffer, why not just use that? Now you are doing a memory copy. https://github.com/nodejs/io.js/blob/master/lib/buffer.js#L96-L100

msmiley: you are absolutely right, thanks for catching that!

msmiley:
sorry for taking so long on this, rebased and changed the name of the variable to promoteBuffers to be more in line with the existing promoteLongs option. Also, nativeBuffers may make some users think C++.

Added a couple tests, let me know if I need to add a test somewhere else.

Will make a PR on bson-ext.

PR #126, opened 2015-05-26
Update bson.js

LambertEliott:
Hello,

My update uses splice in deserializationStream.

It allows users to trigger controls before array insertion thanks to Array.prototype.splice.

Thanks


christkv:
refactoring of the bson js parser is completed.

PR #133, opened 2015-07-14
Fix deserializing a bson collection from file

jkresner:
- Remove gate stopping functionality
- Add test + test file


jkresner:
Apologies if the test doesn't run - I'm not familiar with node unit.

PR #143, opened 2015-08-27
serializeWithBufferAndIndex now can serialize >17Mb

vird:


christkv:
I cannot accept this pull request because the single buffer serialization strategy is crucial for the BSON serializer to be competitive with the c++ parser. Maybe you can add a different serialization method for this instead.

PR #151, opened 2015-10-26
Double should have an option to check if convert to Number

flyingfisher:
The problem is:
    Int32 1 -> 1 (number in javascript)
    Double 1.0 -> 1 (number in javascript)
Now there is no way to find if origin value is a double or int32 value. This is good when I just want to read the value, but not good when I want to change value  and set it back to database.

I find that, you add an option 'promoteLongs' to solve kind of problem of Int32 and Int64.
So now I want to add an option like 'promoteDoubles' to solve this problem.

### DIFF
--- a/lib/bson/parser/deserializer.js
+++ b/lib/bson/parser/deserializer.js
@@ -59,6 +59,7 @@ var deserializeObject = function(buffer, options, isArray) {
   var cacheFunctions = options['cacheFunctions'] == null ? false : options['cacheFunctions'];
   var cacheFunctionsCrc32 = options['cacheFunctionsCrc32'] == null ? false : options['cacheFunctionsCrc32'];
   var promoteLongs = options['promoteLongs'] == null ? true : options['promoteLongs'];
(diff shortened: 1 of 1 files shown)


christkv:
The BSON parser exposes a Double type just for this case, no need for additional flags.

flyingfisher:
Yes, I can set double value use 'BSON.Double'. But my question is in Find result.
In find result, double value '1.0' will always shown as javascript number,  which is same to int32 value '1'. I can not tell if this value is an Int32 or a double when I get it from find result.

Now the only way for me to check if a value is Double is to check the number's decimal, Which is not accurate.

Assume there is a value 1.1.
op,  value,  type
find     1.1    double
set      1.0    double  (use BSON.Double)
find      1      int32  <-- here is my problem, should be double. But I can not find a way to tell its type.

christkv:
yeah you can't tell the difference due to the way the javascript number type is used for all numeric values.

flyingfisher:
Thank you for replying. I get your point. I will try to find out workaround for my problem.

PR #161, opened 2016-01-19
When deserializing an arbitary index in a buffer, stay relative to the index

squirrelsama:
In the event that you are providing a `Buffer` and non-zero `startIndex` to `deserialize` or `deserializeStream`, if the `startIndex` points to a valid BSON document, it will incorrectly error out saying `"One object, sized correctly, with a spot for an EOO, but the EOO isn't 0x00"`

This PR fixes that.


christkv:
can you add a test case for this, you can run the the test by doing npm test

PR #172, opened 2016-03-29
Support isValid and new ObjectId accross different versions and instances of the bson package

seriousben:
This PR uses duck-typing to better support `new ObjectID` and `isValid`. It uses logic already present in the `equals` implementation and brings it in the `constructor` and the `isValid`.

The main reason why this is needed is to support the following case:
```
var m1 = require('mongodb')
var m2 = require('./node_modules/MODULE/node_modules/mongodb')
var hash = '56facc09cb986a0051a19c10';

m1.ObjectID.isValid(new m2.ObjectID(hash)) // Returns false! Should return true!
new m1.ObjectId(new m2.ObjectId(hash)) // Throws it should just return an ObjectId
```

### DIFF
--- a/lib/bson/objectid.js
+++ b/lib/bson/objectid.js
@@ -27,7 +27,8 @@ var checkForHexRegExp = new RegExp("^[0-9a-fA-F]{24}$");
 */
 var ObjectID = function ObjectID(id) {
   if(!(this instanceof ObjectID)) return new ObjectID(id);
-  if((id instanceof ObjectID)) return id;
+  // Duck-typing to support ObjectId from different npm packages
+  if((id instanceof ObjectID) || (id && id.toHexString)) return id;
 
   this._bsontype = 'ObjectID';
   var __id = null;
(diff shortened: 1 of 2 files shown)


lib/bson/objectid.js
```diff
@@ -27,7 +27,8 @@ var checkForHexRegExp = new RegExp("^[0-9a-fA-F]{24}$");
 */
 var ObjectID = function ObjectID(id) {
   if(!(this instanceof ObjectID)) return new ObjectID(id);
-  if((id instanceof ObjectID)) return id;
+  // Duck-typing to support ObjectId from different npm packages
+  if((id instanceof ObjectID) || (id && id.toHexString)) return id;
```
vkarpov15: Highly questionable IMO, if you're going to ducktype I'd recommend using a more robust check. With this PR, a [tinycolor](https://github.com/bgrins/TinyColor) would be a valid ObjectId...

seriousben: Thanks @vkarpov15 - It makes a lot of sense, implemented in #173

PR #173, opened 2016-03-30
fix(ObjectID): isValid(new ObjectID()) returns true - Improve duck-typing in constructor

seriousben:
Improving previous PR to take advantage of the validation already done in the `isValid`


seriousben:
@christkv - Don't forget to look at this, it makes the whole ObjectID constructed from another ObjectID more robust.

lib/bson/objectid.js
```diff
@@ -45,6 +44,9 @@ var ObjectID = function ObjectID(id) {
   } else if(id != null && id.length === 12) {
     // assume 12 byte string
     this.id = id;
+  } else if(id != null && id.toHexString) {
```
seriousben: This takes advantage of `var valid = ObjectID.isValid(id)` at line 34.

PR #178, opened 2016-05-13
ObjectID(x) where is an ObjectID-like makes a useless memory allocation

gilles-poirot:
This is a rather common scenario and perf gain is around 3x - 4x faster


christkv:
how is creating an object id with another object id ever a common scenario ?

gilles-poirot:
Hi, 
Thanks for merging. 
Actually, when writing backend services or class library, it is quite common to accept _id parameters to be passed as strings or ObjectIDs (e.g. depending on whether you are coming from http REST layer or from other services). In that case you need to typecast all _id params. 
Regards,
Gilles

PR #195, opened 2016-10-11
fix(deserializer): corrected the check for (size + index) comparison…

JoelParke:
… to buffer.length


JoelParke:
Not sure of the build/test process for js-bson.  Also conventions for pull requests.

PR #202, opened 2016-12-12
simplify and probably improve

caub:


caub:
tests run between 1080ms and 1150ms slighly slower, I guess node's Buffer.from(.. 'hex') isn't so fast

Issue #204, opened 2017-01-03
Hex is not a function on ObjectId creation


thundo:
I'm getting a `TypeError: hex is not a function` using the mongo node driver with
 ```js
new ObjectId('586bc0c7c7a9ab00da14d1d2');
```
It stopped working between 1.0.1 and 1.0.2 of this library. Platform: Linux, Node v. 4.3.2

May be related to changes of #201.

Thanks

**Comment by ahmadferdous** [comment]
I got the same error when I did a fresh npm install today. The stack trace points to bson. Can you please take a look?
```
/xyz/node_modules/mongoose/node_modules/bson/lib/bson/objectid.js:52
    return new ObjectID(Buffer.from(id, 'hex'));
                               ^

TypeError: hex is not a function
    at Function.from (native)
    at Function.from (native)
    at new ObjectID (/xyz/node_modules/mongoose/node_modules/bson/lib/bson/objectid.js:52:32)
    at ObjectID (/xyz/node_modules/mongoose/node_modules/bson/lib/bson/objectid.js:31:42)
```
```
⇒  npm view bson version
1.0.3
```
Node v4.2.4, on Mac.

**Comment by ahmadferdous** [comment]
I have created a pull request to fix this issue: https://github.com/mongodb/js-bson/pull/205

**Comment by ahmadferdous** [comment]
The following outputs true in Nodejs v4.2.4. 
```
var hasBufferType = false;
if(Buffer && Buffer.from) hasBufferType = true;
console.log(hasBufferType);
```

jefarmstrong:
I'm running v4.3.2 in production and (Buffer && Buffer.from) also returns true
but Buffer.from('123', 'hex') fails with 'TypeError: hex is not a function'

alexdonei:
I've the same problem ('TypeError: hex is not a function') with v4.4.6 in production, with v4.6.1 il local env I don't have problem.

JvrBaena:
Same here. This issue is breaking ObjectId validation/generation for everyone using Mongoose >= 4.7.3. The `hex is not a function` error is catched as a CastError when generating any ObjectId from Strings. See here https://github.com/Automattic/mongoose/issues/4860 and here https://github.com/Automattic/mongoose/issues/4867

Would be great if we could get https://github.com/mongodb/js-bson/pull/205 (or another solution for this) merged :)

Thanks!

Kikobeats:
NodeJS API for Buffer is inconsistent between node versions.

[...]

PR #205, opened 2017-01-03
Fix error - hex is not a function

ahmadferdous:
Fixes issue #204.

`Buffer.from` API parameters vary in different Node.js versions. So,
existence of `Buffer.from` function is not enough. Check if the API we use
are defined.

### DIFF
--- a/lib/bson/objectid.js
+++ b/lib/bson/objectid.js
@@ -14,7 +14,13 @@ var hasBufferType = false;
 
 // Check if buffer exists
 try {
-  if(Buffer && Buffer.from) hasBufferType = true;
+  if(Buffer && Buffer.from) {
+    // Buffer.from API parameters vary in different Node.js versions. So,
+    // existence of Buffer.from function is not enough. Check if the API we use
+    // are defined.
+    Buffer.from('1234', 'hex');
+    hasBufferType = true;
+  }
 } catch(err) {};
 
 /**
--- a/test/node/object_id_tests.js
+++ b/test/node/object_id_tests.js
@@ -42,7 +42,12 @@ exports['should correctly create ObjectId from uppercase hexstring'] = function(
  * @ignore
  */
 exports['should correctly create ObjectId from Buffer'] = function(test) {
-  if(!Buffer.from) return test.done();
+  try {
+    Buffer.from('1234', 'hex');
+  } catch (err) {
+    return test.done();
+  }
+
   var a = 'AAAAAAAAAAAAAAAAAAAAAAAA';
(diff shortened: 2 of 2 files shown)


philhardy:
I can confirm this patch fixes the problem for my scenario 👍

Thanks @ahmadferdous !

Gormartsen:
Same here. It fix issue for node v5.4.1 (MacOSX)
mongodb package (mongodb-core@2.1.4)

christkv:
fixed by using new Buffer instead

ahmadferdous:
@christkv , can you please update the following two lines as well? Otherwise, these will confuse.

[the first line](https://github.com/mongodb/js-bson/blob/8dd35ca73c30a2bcab61615ccc5edd053aeb868b/lib/bson/objectid.js#L17) - `Buffer.from` is not used anymore.
```
if(Buffer && Buffer.from) hasBufferType = true;
```
[the second line](https://github.com/mongodb/js-bson/blob/8dd35ca73c30a2bcab61615ccc5edd053aeb868b/lib/bson/objectid.js#L292) - Should be something like `Use Buffer if available`
```
// Use Buffer.from method if available
```

kaankocturk:
Guys I'm new to github and I have the same problem. I realize that this is a fixx but I have no idea how to implement it. Can you go over how to do this step by step?

eschwartz:
@kaankocturk -- are you using `js-bson` directly, or are you seeing this error when using `mongodb` or `mongoose` (because mongodb and mongoose use `js-bson`).

As a general answer, if you see a bug/issue on github that you want to be fixed locally in your code:
- Wait for the issue to be marked as fixed (this issue in bson is fixed, I believe)
- Wait for the fix to be included in a tagged version release (this is sometimes harder to tell... I believe this issue is included in [v1.0.4](https://github.com/mongodb/js-bson/releases/tag/V1.0.4). If you're not sure, you can always ask in a comment on the issue)
- Update your local version of the library (`npm install --save bson@^1.0.4`)

If you're having an issue because this library is a dependency of a dependency, it may be a little tricker to track down. To figure out which of your dependencies is using `bson`, you can run `npm list`, look for `bson`, and then track the dependency graph back up the tree.

Once you find out who is using the broken dependency, you're best best is probably to open a new issue in that dependency's github repo, referencing the original issue.

In this case, you can see there are already a number of issues in `mongodb` and `mongoose` which reference this issue with `bson`. I believe `mongodb` [has already implemented the fix](https://github.com/mongodb/node-mongodb-native/pull/1464) in [v2.2.21](https://github.com/mongodb/node-mongodb-native/releases), but I'm not entirely sure...

PR #206, opened 2017-01-24
fixes assumption that values.bytes is always a Buffer

danielsan:


christkv:
This won't pass on 0.10 and 0.12 so I cannot accept the pr as it stands today.

Issue #212, opened 2017-08-11
In a docker environment, pid is 1.


jiho-kr:
Hi.

Among the values to guarantee the uniqueness of ObjectID is pid.
However, in a docker environment, the pid of most processes has a value of 1.

If pid is 1, I think it is better to use random value.

```
/lib/bson/objectid.js
as is:
var pid = (typeof process === 'undefined' ? Math.floor(Math.random() * 100000) : process.pid) % 0xFFFF;
to be :
var pid = (typeof process === 'undefined' || process.pid === 1 ? Math.floor(Math.random() * 100000) : process.pid) % 0xFFFF;
```

mbroadst:
@jiho-kr SGTM, would you mind putting together a PR for this? Also, you might consider using [crytpo.randomBytes](https://nodejs.org/api/crypto.html#crypto_crypto_randombytes_size_callback) if it's available (rather than always using `Math.random`).

heycalmdown:
Why is the `crypto.randomBytes` preferred at that minimum scale(2 bytes)? Even if there is an alternative implementation [has landed at 2015's](https://v8project.blogspot.kr/2015/12/theres-mathrandom-and-then-theres.html)?

mbroadst:
@heycalmdown not that it's very scientific, but you might find this interesting:
```
> console.time('buffer'); Buffer.from('' + Math.floor(Math.random() * 100000) % 0xFFFF); console.timeEnd('buffer');
buffer: 0.622ms
undefined
> console.time('crypto'); crypto.randomBytes(2); console.timeEnd('crypto');
crypto: 0.071ms
undefined
```

heycalmdown:
make sense

mbroadst:
fixed in #213

Loksly:
If _crypto_ library should be used there, should it be also used for _MachineID_?
https://github.com/mongodb/js-bson/blob/1.0-branch/lib/bson/objectid.js#L9

    var MACHINE_ID = parseInt(Math.random() * 0xFFFFFF, 10);

mbroadst:
@Loksly yes that sounds reasonable. At the end @jiho-kr chose not to use `crypto.randomBytes` so, I didn't feel the need to press further on this.  If you'd like to submit a patch to optionally use it in both cases that would be very welcome!

PR #213, opened 2017-08-16
Maintain uniqueness in a docker environment

jiho-kr:
Among the values to guarantee the uniqueness of ObjectID is pid.
However, in a docker environment, the pid of most processes has a value of 1.

If pid is 1, it uses random value.

https://github.com/mongodb/js-bson/issues/212


mbroadst:
@jiho-kr thanks for the contribution!

PR #217, opened 2017-09-20
style(*): format and lint entire codebase

kmahar:


kmahar:
@mbroadst updated to use globals for `Uint8Array` and `ArrayBuffer`

also the places I had mentioned doing funky stuff because of linter 
- ignoring no control character in regex [here](https://github.com/mongodb/js-bson/blob/877ae8058d1fd8b3e7ece6b9c11dcc640bd2347e/lib/bson/parser/serializer.js#L15) - we use it to verify that input does not contain null bytes. 
- there were some complaints about `while(true)`s. I changed a couple of these ([here](https://github.com/mongodb/js-bson/blob/877ae8058d1fd8b3e7ece6b9c11dcc640bd2347e/lib/bson/long.js#L127), [here](https://github.com/mongodb/js-bson/blob/877ae8058d1fd8b3e7ece6b9c11dcc640bd2347e/lib/bson/timestamp.js#L127)) to be looping on a more specific condition, but one was a loop with many possible termination conditions that seemed annoying to refactor, so I just changed to to `var done = false; while(!done) {...}`.  [here](https://github.com/mongodb/js-bson/blob/877ae8058d1fd8b3e7ece6b9c11dcc640bd2347e/lib/bson/parser/deserializer.js#L80). if it's more preferable to just disable linting for that line instead let me know

PR #218, opened 2017-09-22
Update bson corpus

kmahar:
@mbroadst do you want me to make the travis change a separate PR? (also, do we need all the stuff in there with different Mongo versions, given that we don't actually use Mongo in this module?)


.travis.yml
```diff
@@ -1,8 +1,34 @@
+sudo: required
```
mbroadst: `sudo: false`

.travis.yml
```diff
@@ -1,8 +1,34 @@
...
-  - "5"
-  - "6"
-  - "7"
+
+cache:
+  yarn: true
+  directories:
+    - node_modules
+
+jobs:
```
mbroadst: remove the whole `jobs` section, since we don't actually need to run against mongodb, and replace it with:
```
node_js:
  - 4
  - 6
  - 8
```

lib/bson/decimal128.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 
-var Long = require('./long');
+var Long = require('./long'),
+  f = require('util').format;
```
mbroadst: as long as you're already here, the preferred type for top level includes is `const` since they are not meant to change.

lib/bson/decimal128.js
```diff
@@ -79,13 +80,12 @@ var isDigit = function(value) {
...
   var DIVISOR = Long.fromNumber(1000 * 1000 * 1000);
   var _rem = Long.fromNumber(0);
-  var i = 0;
 
   if (!value.parts[0] && !value.parts[1] && !value.parts[2] && !value.parts[3]) {
     return { quotient: value, rem: _rem };
   }
 
-  for (i = 0; i <= 3; i++) {
+  for (var i = 0; i <= 3; i++) {
```
mbroadst: use `let` here so `i` is lexically scoped to the for statement

lib/bson/decimal128.js
```diff
@@ -245,9 +242,29 @@ Decimal128.fromString = function(string) {
     throw new Error('' + string + ' not a valid Decimal128 string');
   }
 
-  // Check if we have an illegal exponent format
-  if (stringMatch && stringMatch[4] && stringMatch[2] === undefined) {
-    throw new Error('' + string + ' not a valid Decimal128 string');
+  if (stringMatch) {
+    // full_match = stringMatch[0]
```
mbroadst: these are for documentation purposes, not left-over commented out code, correct?

lib/bson/decimal128.js
```diff
@@ -290,21 +307,14 @@ Decimal128.fromString = function(string) {
...
+    if (sawRadix) radixPosition = radixPosition + 1;
 
     nDigitsRead = nDigitsRead + 1;
     index = index + 1;
   }
 
-  if (sawRadix && !nDigitsRead) {
-    throw new Error('' + string + ' not a valid Decimal128 string');
-  }
+  if (sawRadix && !nDigitsRead) throw new Error('' + string + ' not a valid Decimal128 string');
```
mbroadst: perhaps we should wrap the string in quotes?
```
throw new Error(`"${string}" not a valid Decimal128 string`);
```

mbroadst: Also note I used template strings here, it's a convenient way to not need that imported `f` format function above.

lib/bson/parser/deserializer.js
```diff
@@ -384,7 +392,8 @@ var deserializeObject = function(buffer, index, options, isArray) {
         buffer[index + stringSize - 1] !== 0
       )
```
mbroadst: maybe we should wrap this `throw` in braces? It's a little hard to follow

mbroadst: throw new Error('' + string + ' not a valid Decimal128 string');

lib/bson/parser/deserializer.js
```diff
@@ -539,7 +561,7 @@ var deserializeObject = function(buffer, index, options, isArray) {
           elementType.toString(16) +
           ' for fieldname "' +
           name +
-          '", are you using the latest BSON parser'
+          '", are you using the latest BSON parser?'
```
mbroadst: nice ❓

lib/bson/parser/serializer.js
```diff
@@ -354,7 +358,8 @@ var serializeObject = function(
   );
   // Pop stack
   path.pop();
-  // Write size
+
+  // var size = endIndex - index;
```
mbroadst: cruft?

lib/bson/parser/serializer.js
```diff
@@ -629,33 +638,15 @@ var serializeDBRef = function(buffer, key, value, index, depth, serializeFunctio
...
-        $ref: value.namespace,
-        $id: value.oid
-      },
-      false,
-      index,
-      depth + 1,
-      serializeFunctions
-    );
-  }
+  var output = {
```
mbroadst: great refactor ++

test/node/bson_test.js
```diff
@@ -1532,13 +1540,12 @@ exports['Should handle complicated all typed object'] = function(test) {
...
-  assertBuffersEqual(test, serialized_data, serialized_data2, 0);
+
+  expect(serialized_data).to.deep.equal(serialized_data2);
 
   serialized_data2 = createBSON().serialize(doc2, false, true);
 
-  for (var i = 0; i < serialized_data2.length; i++) {
-    require('assert').equal(serialized_data2[i], serialized_data[i]);
-  }
+  expect(serialized_data).to.deep.equal(serialized_data2);
```
mbroadst: man whaaaat was going on in that old version of this

test/node/decimal128_tests.js
```diff
@@ -808,29 +807,31 @@ exports['fromString exponent normalization'] = function(test) {
   );
   test.deepEqual(bytes, result.bytes);
 
+  // this should throw error according to spec.
```
mbroadst: and it's covered by the corpus tests?

lib/bson/decimal128.js
kmahar: yeah, I spent a while figuring out what exactly the regex was giving back, so I figured I'd leave it for future reference.

lib/bson/parser/deserializer.js
kmahar: huh?

lib/bson/parser/deserializer.js
kmahar: thx

test/node/bson_test.js
kmahar: ¯\_(ツ)_/¯

lib/bson/parser/deserializer.js
mbroadst: I have no idea how this got here...

test/node/decimal128_tests.js
kmahar: yep, one of the cases that should throw a parse error here: https://github.com/mongodb/js-bson/pull/218/files?diff=unified#diff-50fc67aa335868ab2ed1e27d8492655bR168

PR #219, opened 2017-09-23
convert tests to mocha

mbroadst:


test/node/bson_compliance_test.js
```diff
@@ -6,130 +6,113 @@ var BSON = require('../..'),
...
-
-    try {
-      // Create a buffer containing the payload
-      var buffer = new Buffer(doc.encoded, 'hex');
-      // Attempt to deserialize
-      bson.deserialize(buffer);
-      test.ok(false);
-    } catch (err) {
-      test.ok(true);
+      try {
```
kmahar: I think this is cleaner and more readable than the try/catch:
```
expect(()=>{
	bson.deserialize(buffer)
}).to.throw();

File: test/node/bson_test.js [hunk 2]
```diff
@@ -38,22 +38,13 @@ BSON.BSON_BINARY_SUBTYPE_UUID = 3;
...
-//   for (var i = 0; i < numberofValues; i++) {
-//     array += String.fromCharCode(parseInt(string[i * 2] + string[i * 2 + 1], 16));
-//   }
-//   return array;
-// };
-
-var assertBuffersEqual = function(test, buffer1, buffer2) {
-  if (buffer1.length !== buffer2.length)
-    test.fail('Buffers do not have the same length', buffer1, buffer2);
+var assertBuffersEqual = function(done, buffer1, buffer2) {
```
kmahar: someday we should probably switch all the times this function is used to just use expect(buffer1).to.deep.equal(buffer2). unless if this is doing something different I'm not seeing?

test/node/bson_compliance_test.js
mbroadst: @kmahar absolutely. This was more of an incremental change, but yes moving forward we should indeed use the new tools we have

test/node/bson_test.js
mbroadst: Also correct.  Chai used to not do deep comparison between buffers, so I believe that's why we ended up with this helper. (It predates even Buffer.compare!)

PR #223, opened 2017-09-28
fix(dbref): correctly upgrade dbrefs to BSON.DBRef

kmahar:
using this module in mongodb-extjson, I realized that my previous updates to the code did not correctly convert { dbref: { $ref: .., $id:... }} to BSON.DBRef objects. the test failed to catch this because it was just checking that fields matched up rather than checking for deep equality, so I have fixed that as well.

### DIFF
--- a/lib/bson/parser/deserializer.js
+++ b/lib/bson/parser/deserializer.js
@@ -404,6 +404,7 @@ var deserializeObject = function(buffer, index, options, isArray) {
         (buffer[index++] << 8) |
         (buffer[index++] << 16) |
         (buffer[index++] << 24);
+
       object[name] = new Timestamp(lowBits, highBits);
     } else if (elementType === BSON.BSON_DATA_MIN_KEY) {
       object[name] = new MinKey();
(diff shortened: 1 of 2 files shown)


lib/bson/parser/deserializer.js
```diff
@@ -173,6 +173,19 @@ var deserializeObject = function(buffer, index, options, isArray) {
         object[name] = deserializeObject(buffer, _index, options, false);
       }
 
+      if (name === 'dbref') {
```
mbroadst: I think I'm confused.  This appears to be checking explicitly for a property called `dbref` on an object, that's not really spec-related is it?  This seems wrong, but maybe I'm misunderstanding the problem here.

kmahar: now that you say this and I look at the documentation on dbrefs I think this is wrong. 

I had just been looking at the test cases in [this file](https://github.com/mongodb/specifications/blob/master/source/bson-corpus/tests/dbref.json) and the fact that the last one is not supposed to be a dbref made me think dbrefs had to explicitly have that key. (since all the previous cases do).
but now that I compare it to the previous test case, I think it's actually because it has a $ key (`$banana`) which isn't part of a DBRef, but a non-$ additional key is ok.

PR #224, opened 2017-10-04
refactor(timestamp):make timestamp subclass of long

kmahar:
This is just copy-paste of @mbroadst's past timestamp refactor in mongodb-extjson, and updated test to no longer assert that a timestamp is not an instance of long.


test/node/bson_test.js
```diff
@@ -1219,7 +1219,7 @@ describe('BSON', function() {
     expect(long instanceof Long).to.be.ok;
     expect(!(long instanceof Timestamp)).to.be.ok;
     expect(timestamp instanceof Timestamp).to.be.ok;
-    expect(!(timestamp instanceof Long)).to.be.ok;
+    // expect(!(timestamp instanceof Long)).to.be.ok;
```
mbroadst: hm what's going on here?

kmahar: this test is "Should Correctly Serialize and Deserialize Long Integer and Timestamp as different types". but since this change just makes Timestamp a wrapper around Long, Timestamp will now be an instance of Long

mbroadst: Yeah, but then shouldn't we just delete the expectation?  Are we anticipating that this will be true again some day?

mbroadst: Or rather update the expectation to assert it _is_ a `Long`?

kmahar: ok good point I'll change it. deleting things entirely just feels more dramatic 😱

PR #226, opened 2017-10-06
refactor(buffers/uint8arrays): polyfill atob and btoa and build into …

kmahar:
…BSON types for use in mongodb-extjson

plus a few other small changes to make the web case in mongodb-extjson tests work


kmahar:
hmmm... any idea why these tests would only be failing for node 4??

mbroadst:
@kmahar looks like you have a warning in there: `(node) Buffer.set is deprecated. Use array indexes instead.`

kmahar:
this should be good to go, and needs to be merged in for extended JSON build to succeed, since it depends on js-bson#2.0.0

lib/bson/objectid.js
```diff
@@ -12,13 +12,6 @@ var MACHINE_ID = parseInt(Math.random() * 0xffffff, 10);
 // Regular expression that checks for hex value
 var checkForHexRegExp = new RegExp('^[0-9a-fA-F]{24}$');
 
-// Check if buffer exists
```
mbroadst: I find all of this a bit confusing (not because of your work, but the original code).  So we have this check to ensure that we know if `Buffer` exists or not, but then later we use a `var _Buffer = Buffer;` and now we're switching to use `if (Buffer == null)` checks.

Is there a way we can clean this up a bit more?  Also, have you guaranteed that the replacement for the `Buffer.from` (using hex constructor) is always available in node 4+?

lib/bson/parser/serializer.js
```diff
@@ -371,7 +371,8 @@ var serializeDecimal128 = function(buffer, key, value, index, isArray) {
   index = index + numberOfWrittenBytes;
   buffer[index++] = 0;
   // Write the data from the value
-  value.bytes.copy(buffer, index, 0, 16);
+  if (value.bytes.copy) value.bytes.copy(buffer, index, 0, 16);
```
mbroadst: This might be premature optimization, but can't we cache this check far earlier? E.g. at the top like `var hasCopy = Buffer.prototype.copy != null`

lib/bson/objectid.js
kmahar: so it looks like `new Buffer(string, encoding)` exists Node 4+, but is deprecated starting with Node 6. meanwhile `Buffer.from(string, encoding)` exists [Node 4.5+](https://nodejs.org/docs/latest-v4.x/api/buffer.html#buffer_class_method_buffer_from_str_encoding). so I guess we should try to use` Buffer.from`, and only fall back on `new Buffer` if `Buffer.from` doesn't exist?

kmahar:
so this doesn't work now but you'll see the kind of errors I've been trying to fix. so the reason this wasn't easily fixable was that a lot of objectID code relies on buffer stuff that not's easily translated to uint8arrays... but I couldn't figure out a way to solve it without writing a ton of if/elses based on buffer vs uint8array

PR #227, opened 2017-11-02
docs: add FAQ re: `undefined`

vkarpov15:
I'd love to have something to link to that explains the rationale behind the `undefined` -> `null` conversion re: issues like https://github.com/Automattic/mongoose/issues/5776 .


mbroadst:
Thanks Val!

PR #228, opened 2017-11-06
Feat/uint array/handle uint array

daprahamian:
Adds a check to wrap all passed in `Uint8Arrays` with a `Buffer`.


mbroadst:
@daprahamian what's going on with this PR?

PR #231, opened 2017-12-22
docs: add FAQ about toBSON()

vkarpov15:
I haven't been able to find any docs about `toBSON()`, so I figure this is as good a place to add it as any.


mbroadst:
thanks val, sorry this took a while to get in for some reason!

PR #232, opened 2018-01-02
Added an option to set the min size of the serialization buffer

xaviergonz:
Added an option to set the min size of the temporary serialization buffer, plus added an optimization so a buffer is not copied after serialization when not needed.

### DIFF
--- a/lib/bson/bson.js
+++ b/lib/bson/bson.js
@@ -24,10 +24,12 @@ var deserialize = require('./parser/deserializer'),
  * @ignore
  * @api private
  */
-// Max Size
+// Default Max Size
 var MAXSIZE = 1024 * 1024 * 17;
-// Max Document Buffer size
+
+// Current Internal Temporary Serialization Buffer
 var buffer = new Buffer(MAXSIZE);
+var currentBufferSize = MAXSIZE;
 
 var BSON = function() {};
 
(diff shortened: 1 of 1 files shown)


lib/bson/bson.js
```diff
@@ -49,6 +52,14 @@ BSON.prototype.serialize = function serialize(object, options) {
     typeof options.serializeFunctions === 'boolean' ? options.serializeFunctions : false;
   var ignoreUndefined =
     typeof options.ignoreUndefined === 'boolean' ? options.ignoreUndefined : true;
+  var minInternalBufferSize =
+    typeof options.minInternalBufferSize === 'number' ? options.minInternalBufferSize : MAXSIZE;
+  
+  // Resize the internal serialization buffer if needed
+  if (currentBufferSize < minInternalBufferSize) {
```
mbroadst: do we need to keep an extra variable around for this? Doesn't `if (buffer.length < minInternalBufferSize)` suffice?

lib/bson/bson.js
```diff
@@ -93,15 +104,14 @@ BSON.prototype.serializeWithBufferAndIndex = function(object, finalBuffer, optio
 
   // Attempt to serialize
   var serializationIndex = serializer(
-    buffer,
+    finalBuffer,
```
mbroadst: this looks like an unrelated change?

lib/bson/bson.js
xaviergonz: good point

lib/bson/bson.js
xaviergonz: yeah, that's an optimization. it saves one copy since you can just flush to the final buffer instead of flushing to a temp buffer and then that to the final buffer

mbroadst: that's fine, but if you don't mind I'd prefer it in another PR so we can clearly call it out in the changelog

mbroadst:
@xaviergonz thanks for the contribution!

xaviergonz:
You're welcome 👍

PR #233, opened 2018-02-22
Fixed browser property in package.json

aydrian:
Updated the `webpack.dist.config.js` to make `npm run build` work.
Updated the `browser` property in `package.json` to point to the webpack output.
Added a `prepublishOnly` script to ensure the transpiled version is built.


hswolff:
It may be additionally desired to gitignore the `browser_build` folder as that seems un-necessary. Can just be included in the npm package.

aydrian:
@daprahamian Would you like me to add that to the `.gitignore` while I have this open?

daprahamian:
@aydrian Yeah, that would be great, thanks!

aydrian:
@daprahamian No problem. Good to go.

PR #237, opened 2018-03-05
Fix misleading type docs link

tunderdomb:


README.md
```diff
@@ -68,7 +68,7 @@ console.log('doc_2:', doc_2)
 
 ### BSON types
 
-For all BSON types documentation, please refer to the documentation for the [MongoDB Node.js driver](https://github.com/mongodb/node-mongodb-native).
+For all BSON types documentation, please refer to the documentation for the [MongoDB Node.js driver](https://docs.mongodb.com/manual/reference/bson-types/).
```
mbroadst: Could we instead say:
```
For all BSON types documentation, please refer to the following sources:
  * [MongoDB BSON Type Reference](https://docs.mongodb.com/manual/reference/bson-types/)
  * [BSON Spec](https://bsonspec.org/)
```
?

tunderdomb: Yes, sure. Do you think a short list of the available types would be helpful?
I wanted to include it but assumed it was linked because if it changes you have to update only one place.

mbroadst: yeah I think just linking to the docs directly is ideal, it doesn't change often but as you can tell when things do change it's people are a bit rusty to update documentation :)

mbroadst:
thanks for the contribution!

PR #157, opened 2015-12-10
use more specific TypeError objects for function parameters errors

hhromic:
Hi,
here is the PR for the proposed changes commented in [here](https://github.com/mongodb/js-bson/issues/156). I went further and modified all other functions in the codebase where this specific error object makes sense (function parameters checking). So the error handling in the lib is consistent.

Cheers and thanks!
Hugo.

PR #243, opened 2018-04-12
use the more specific `TypeError` objects for function parameters errors.

hhromic:
[this PR is a rebase (and update) of PR #157]

When constructing a new `ObjectId` object, the passed id value is validated for conversion to an hex string.
This is just fine, however a generic Error object is thrown when this validation fails.
Tracking the error then becomes unconvenient.
From official [Joyent's design patterns for errors](https://www.joyent.com/developers/node/design/errors), it is suggested that:

> 3. Use the Error's name property to distinguish errors programmatically.
>    When you need to figure out what kind of error this is, use the name property.
>    Built-in JavaScript names you may want to reuse include "RangeError"
>    (an argument is outside of its valid range) and "TypeError" (an argument has the wrong type).
>    For HTTP errors, it's common to use the RFC-given status text to name the error, like
>    "BadRequestError" or "ServiceUnavailableError".

This PR uses the [`TypeError` class](https://nodejs.org/api/errors.html#errors_class_typeerror) instead of the generic `Error` object.
Because `TypeError` is a subclass of `Error`, all existing code should continue to work flawlessly and new code can more easily track an invalid argument error.

### DIFF
--- a/alternate_parsers/bson.js
+++ b/alternate_parsers/bson.js
@@ -526,7 +526,7 @@ BSON.serializeWithBufferAndIndex = function serializeWithBufferAndIndex(
  */
 var serializeObject = function(object, checkKeys, buffer, index, serializeFunctions) {
   if (object.toBSON) {
-    if (typeof object.toBSON !== 'function') throw new Error('toBSON is not a function');
+    if (typeof object.toBSON !== 'function') throw new TypeError('toBSON is not a function');
     object = object.toBSON();
     if (object != null && typeof object !== 'object')
       throw new Error('toBSON function did not return an object');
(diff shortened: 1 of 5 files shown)


mbroadst:
@hhromic hi, sorry one last thing, please don't commit the update to the browser build. That will be updated automatically as part of the build process. Thanks!

hhromic:
Hi, thanks for the clarification! Indeed I was confused if I should have updated the browser build or not.
I reverted that in the PR now. Cheers!

mbroadst:
@hhromic thanks for the contribution!

hhromic:
happy to help!

PR #245, opened 2018-04-16
fix(deserializeStream): allow multiple documents to be deserialized

### LINKED ISSUE #244: deserializeStream throws `corrupt bson message`
Hi,
I was using v2.0.4 of the library for some testing and had an issue reading back data using `deserializeStream`.  It looks like the problem is on [this line](https://github.com/mongodb/js-bson/blob/2.0.0/lib/bson/parser/deserializer.js#L27) specifically the bit which says `buffer.length !== size`.

When you're [deserializing a stream](https://github.com/mongodb/js-bson/blob/2.0.0/lib/bson/bson.js#L196) of objects we'd expect the buffer to be bigger than the current object right?

This change seems to be introduced since the [1.0-branch](https://github.com/mongodb/js-bson/blob/2.0.0/lib/bson/bson.js#L196).

There's also a significant chance that I'm doing something silly / wrong with the library and not understood how the code's working.  Here is an example which shows the behaviour:

```
const BSON = require('bson');
const bson = new BSON();

docs = [
  { foo: "bar" },
  { foo: "baz" },
  { foo: "quux" },
]

// Serialize the test data
serializedDocs = []
for (i = 0; i < docs.length; i++) {
  serializedDocs[i] = bson.serialize(docs[i])
}
buf = Buffer.concat(serializedDocs)

// The following are all fine
bson.deserialize(buf.slice(0, 18))
bson.deserialize(buf.slice(18, 36))
bson.deserialize(buf.slice(36, 55))

parsedDocs = []
// This is not happy: `Error: corrupt bson message`
bson.deserializeStream(buf, 0, docs.length, parsedDocs, 0)
console.log({parsedDocs})
```

It behaves as expected if you remove `buffer.length !== size` from the line mentioned above.  I'm hesitant to open a PR because maybe there's a reason it's been changed and I'm just using it wrong?

There's no rush from my perspective, I've got a workaround but I thought it might be good to point out.  Thanks for your help.

Ben

daprahamian:
deserializeStream was calling deserialize, which required the buffer
size to exactly match the object size per the corpus test. Fixing
this by adding a flag that deserializeStream passes to allow the
buffer to be larger than the object size.

Fixes #244

### DIFF
--- a/lib/bson/bson.js
+++ b/lib/bson/bson.js
@@ -128,6 +128,7 @@ BSON.prototype.serializeWithBufferAndIndex = function(object, finalBuffer, optio
  * @param {Object} [options.promoteValues=false] when deserializing will promote BSON values to their Node.js closest equivalent types.
  * @param {Object} [options.fieldsAsRaw=null] allow to specify if there what fields we wish to return as unserialized raw buffer.
  * @param {Object} [options.bsonRegExp=false] return BSON regular expressions as BSONRegExp instances.
+ * @param {boolean} [options.allowObjectSmallerThanBufferSize] allows the buffer to be larger than the parsed BSON object
  * @return {Object} returns the deserialized Javascript Object.
  * @api public
  */
@@ -183,7 +184,8 @@ BSON.prototype.deserializeStream = function(
   docStartIndex,
   options
 ) {
-  options = options != null ? options : {};
+  options = Object.assign({}, options);
+  options.allowObjectSmallerThanBufferSize = true;
   var index = startIndex;
   // Loop over all documents
   for (var i = 0; i < numberOfDocuments; i++) {
@@ -191,7 +193,7 @@ BSON.prototype.deserializeStream = function(
     var size =
       data[index] | (data[index + 1] << 8) | (data[index + 2] << 16) | (data[index + 3] << 24);
     // Update options with index
-    options['index'] = index;
+    options.index = index;
     // Parse the document at this point
     documents[docStartIndex + i] = this.deserialize(data, options);
     // Adjust index by the document size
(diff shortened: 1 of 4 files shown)


lib/bson/bson.js
```diff
@@ -183,15 +184,16 @@ BSON.prototype.deserializeStream = function(
   docStartIndex,
   options
 ) {
-  options = options != null ? options : {};
+  options = Object.assign({}, options);
```
mbroadst: `options = Object.assign({ allowObjectSmallerThanBufferSize: true }, options);` ?

lib/bson/bson.js
```diff
@@ -128,6 +128,7 @@ BSON.prototype.serializeWithBufferAndIndex = function(object, finalBuffer, optio
  * @param {Object} [options.promoteValues=false] when deserializing will promote BSON values to their Node.js closest equivalent types.
  * @param {Object} [options.fieldsAsRaw=null] allow to specify if there what fields we wish to return as unserialized raw buffer.
  * @param {Object} [options.bsonRegExp=false] return BSON regular expressions as BSONRegExp instances.
+ * @param {boolean} [options.allowObjectSmallerThanBufferSize] allows the buffer to be larger than the parsed BSON object
```
mbroadst: woof, what a property name.. either way let's document that it has a default here as well

lib/bson/parser/deserializer.js
```diff
@@ -23,8 +23,11 @@ var deserialize = function(buffer, options, isArray) {
     (buffer[index + 2] << 16) |
     (buffer[index + 3] << 24);
 
+  const sizeCheck = options.allowObjectSmallerThanBufferSize
+    ? buffer.length < size
+    : buffer.length !== size;
   // Ensure buffer is valid size
-  if (size < 5 || buffer.length !== size || size + index > buffer.length) {
+  if (size < 5 || sizeCheck || size + index > buffer.length) {
```
mbroadst: I feel like we can do better here and emit a different `TypeError` for each condition, this would be far more helpful for our users?

jdelman:
Could you guys please merge this in and publish a new version? :)

PR #247, opened 2018-05-16
fix(binary): add type checking for buffer

kvwalker:
Fixes [NODE-1423](https://jira.mongodb.org/browse/NODE-1423)

### DIFF
--- a/lib/bson/binary.js
+++ b/lib/bson/binary.js
@@ -28,6 +28,14 @@ if (typeof global !== 'undefined') {
 function Binary(buffer, subType) {
   if (!(this instanceof Binary)) return new Binary(buffer, subType);
 
+  if (
+    buffer != null &&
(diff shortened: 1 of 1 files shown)


lib/bson/binary.js
```diff
@@ -28,6 +28,14 @@ if (typeof global !== 'undefined') {
 function Binary(buffer, subType) {
   if (!(this instanceof Binary)) return new Binary(buffer, subType);
 
+  if (
+    buffer != null &&
+    !(typeof buffer === 'string') &&
```
daprahamian: 1. We should use `Buffer.isBuffer` instead of `instanceof Buffer`
2. This appears to only check if the buffer is non-null and either string or buffer. Can we add checks for `Uint8Array` and `Array`?

PR #250, opened 2018-05-20
fix(test): actual and expected size diverged

BridgeAR:
This makes sure the actual and expected size are correctly tested
and it fixes failing tests.

I can not verify if this is actually good to accept as is or if there is a underlying bug.

### DIFF
--- a/test/node/bson_test.js
+++ b/test/node/bson_test.js
@@ -1886,7 +1886,7 @@ exports['Should correctly calculate the size of a given javascript object'] = fu
     serializeFunctions: true
   });
   // Validate the correctness
-  test.equal(37, size);
+  test.equal(36, size);
   test.done();
 };
 
@@ -1914,7 +1914,7 @@ exports[
     serializeFunctions: true
   });
   // Validate the correctness
-  test.equal(37, size);
+  test.equal(36, size);
   test.done();
 };
 
@@ -1959,8 +1959,8 @@ exports['Should correctly serializeWithBufferAndIndex a given javascript object'
     index: 0
   });
   // Validate the correctness
-  test.equal(37, size);
-  test.equal(36, index);
+  test.equal(36, size);
+  test.equal(35, index);
   test.done();
 };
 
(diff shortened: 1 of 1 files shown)


BridgeAR:
Interesting... how come the value is calculated differently on travis than on my local machine and on [CITGM](https://github.com/nodejs/citgm)?

mbroadst:
@BridgeAR not sure! These definitely run fine on my machine without change, can you describe the hardware you're running on?

BridgeAR:
It is a [T450s](https://www.notebookcheck.net/Lenovo-ThinkPad-T450s-Ultrabook-Review.137248.0.html) with different memory and a different SSD (but the latter two should not have any impact).

BridgeAR:
@mbroadst please have a look at https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/1449/#showFailuresLink. There are multiple runs machines that fail just as they do on mine.

mbroadst:
@BridgeAR I just tried `boron` and `carbon` on both osx 10.13.3 as well as ubuntu 16.04, and they passed all tests without issue. If you can reliably reproduce on your local machine, maybe you could provide the raw bytes? Maybe we can reverse engineer what's going on here

mbroadst:
@BridgeAR also are you experiencing this only with a bleeding edge version of node?

mbroadst:
@BridgeAR ah, just confirmed this with v10.3.0!

mbroadst:
So it looks like this is a difference between versions of node in how `Function.prototype.toString` formats its output. Specifically we are not normalizing the output [here](https://github.com/mongodb/js-bson/blob/1.0-branch/lib/bson/parser/serializer.js#L446), so in all of the tests you fixed the strings were differing between `function () {}` and `function() {}`.

Interesting bug! Thanks for pointing this out to us, I'll get the fix in shortly and push new versions of the 1.x and 2.x branches

BridgeAR:
@mbroadst thanks a lot!

PR #251, opened 2018-05-20
fix(bson): fix custom inspect property

BridgeAR:
The currently used custom inspect way is deprecated and was removed
in Node.js 11.0.0-pre. This should fix that.


DominusVilicus:
Great

MylesBorins:
ping @daprahamian

Issue #211, opened 2017-08-04
Feature Request: Custom object to deserialize data to


rahbari:
For now adding method from an ES6 class to a deserialized object can be done using Object.assign() or Object.setPrototypeOf() which both have poor performance. It's nice if we can create the object from class (or even the class itself), pass it to BSON deserialize, and data is added to this object (or object instantiated from class) instead of `var object = isArray ? [] : {};` at the first level of deserializing. 

The PDO of PHP provide such functionally:

> PDO::FETCH_CLASS: returns a new instance of the requested class, mapping the columns of the result set to named properties in the class, and calling the constructor afterwards, unless PDO::FETCH_PROPS_LATE is also given.

The API must become available in Core Driver as well.

mbroadst:
@rahbari perhaps you can follow up with some code examples of your desired experience, I think it would greatly assist discussion.

rahbari:
@mbroadst if we want add some functions to a fetched mongodb object we have 4 options:

1. Create the object from class and add the fields to it using `Object.assign(this, fetchedObj)` in the constructor.
2. Use Object.setPrototypeOf(fetchedObj, desiredPrototype) which breaks every optimization in javascript compiler:
> Changing the [[Prototype]] of an object is, by the nature of how modern JavaScript engines optimize property accesses, a very slow operation, in every browser and JavaScript engine.
3. Add the functions to the fetched object using `Object.assign(fetchedObj, desiredPrototype)`
4. Add the fields directly to the desired object which is the fastest but BSON parser and MongodbCoreDriver must provide.

I made a simple benchmark test for the four solution https://jsperf.com/assign-setprototypeof-direct. You can see how faster the fourth approach is with just 5 simple fields.

For now deserialize method use this line to create an empty object to populate data to:
`var object = isArray ? [] : {};`

It's nice the query object get an object with some functions and add the fetched data to it instead of empty object. 

Or when there are multiple documents it can accept a Class, instantiate it and data to this instantiated object instead of and empty object.

mbroadst:
I am quite familiar with how slow changing prototypes can be on objects, and you still have not provided an example of what exactly you want here (even with a benchmark provided). It _sounds_ like what you want is to be able to override [this code](https://github.com/mongodb/js-bson/blob/1.0-branch/lib/bson/parser/deserializer.js#L69) such that you could pass your own object in:

```
const Long = BSON.Long;
let bson = new BSON();
const doc = { long: Long.fromNumber(100) }
[...]

PR #253, opened 2018-06-07
Deserializing to custom targets (classes)

rahbari:
This pull request tries to cover this feature request https://github.com/mongodb/js-bson/issues/211

I thought of two options for nested targeting:

1- one (which current code is based on) is **using static props of the top target**, for example consider following document in mongodb:

```
{
    _id: 1,
    comments: [
          {_id: 1, content: ""},
          {_id: 2, content: ""}
    ]
}
```

Assigning the top document to an Article object:
```
class Article {}
articles.find({}, {target: Article});
```

Assigning articles to an extended Array:
```
class Article {
     static $comments = class Comments extends Array {}
} 
articles.find({}, {target: Article});
```

assigning articles items to a Comment class:
```
class Article {
     static $comments = class Comments extends Array {
            static $ = class Comment {}
     }
} 
articles.find({}, {target: Article});
```

I used $ at the begging of field names to prevent confusion with Class self properties, and used $ for array items.

2- **Using a custom data structure** like this:
`articles.find({}, {targets: {root: Article, Comments: {root: Comments, item: Comment}}});`

I believe first one is nicer but it's more confusing and the whole hierarchy of data must be defines.
If the maintainers accepts the general idea then it can be discussed.

For now I used **firstBatch** and **nextBatch** to detect documents parsing,  until someone suggest a better approach.

if you want to test this code please use my modified mongodb core driver (I've added necessary code to commands.js).

### DIFF
--- a/lib/bson/parser/deserializer.js
+++ b/lib/bson/parser/deserializer.js
@@ -38,7 +38,7 @@ var deserialize = function(buffer, options, isArray) {
   return deserializeObject(buffer, index, options, isArray);
 };
 
-var deserializeObject = function(buffer, index, options, isArray) {
+var deserializeObject = function(buffer, index, options, isArray, target) {
   var evalFunctions = options['evalFunctions'] == null ? false : options['evalFunctions'];
   var cacheFunctions = options['cacheFunctions'] == null ? false : options['cacheFunctions'];
   var cacheFunctionsCrc32 =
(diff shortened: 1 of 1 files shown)


j:
Hey @rahbari,

I too feel like it'd be nice to have this feature where (de)serialization happens.

This is cool but I still would hope for some sort of way of controlling the object instantiation.

Some potential ideas (typescript):

```typescript
class Author {
  public name: string;
}

class Comment {
  public static $bson = {
    targets: {
      author: Author
    }
  };

  public author: Author;
  public body: string;
}

class Article {
  public static $bson = {
    targets: {
      author: Author,
      comments: [Comment]
    }
  };

  public comments: Comment[] = [];
  public author: Author;
  public body: string;
  public createdAt: Date;
}

articles.find<Article>({}, { target: Article })
```

The only thing I see that may be potentially missing is if people wanted to do field name mapping so that "createdAt" is the class property but is stored as "ca" in the database.

None-the-less, this would be a joy to have.  :)

I slapped this together and have to go, so hopefully it all makes sense.

j:
Playing with this code isn't intuitive right now.  It's not passing the target option to the deserializer.

j:
Also, the document detection question you have is the problem I had when trying to implement mapping into BSON.  :(.. So I'm not sure the best way.  I wish ONLY documents went through (de)serialize functions.

agolin95:
https://jira.mongodb.org/browse/NODE-3012

bajanam:

PR #254, opened 2018-06-07
Update serializer.js

AramilRey:
Keep compatibility with older node.js versions


mbroadst:
Support for legacy versions of node is provided through our dist bundles.

PR #257, opened 2018-06-29
feat(bson): test bson in browser

rweinberger:
Add support for running bson tests in the browser via Karma.

Fixes NODE-1504


shsaskin:
Node 4 is failing from an error loading the karma-rollup-preprocessor because node 4 does not support this version of rollup. We plan to not have karma run on version 4, and we will push again soon with that implemented.

lib/bson/parser/utils.js
```diff
@@ -5,7 +5,7 @@
  * @param {Function} fn The function to stringify
  */
 function normalizedFunctionString(fn) {
-  return fn.toString().replace('function(', 'function (');
+  return fn.toString().replace(/function(.*)\(/, 'function (');
```
daprahamian: 1. This is a good fix
2. I think we should submit this as a separate change. It kind of gets buried here.

package.json
```diff
@@ -51,12 +58,15 @@
     "node": ">=4.0.0"
   },
   "scripts": {
-    "test": "mocha ./test/node",
+    "test": "./tools/run_tests.sh",
+    "karma-test": "./tools/run_tests.sh karma",
```
daprahamian: Let the bike-shedding commence!!!

I personally think we should name these `test-browser` and `test-node`.

rollup.config.js
```diff
@@ -1,21 +1,33 @@
...
 
-export default {
+const commonjs = require('rollup-plugin-commonjs');
+const nodeGlobals = require('rollup-plugin-node-globals');
+const nodeBuiltins = require('rollup-plugin-node-builtins');
+const nodeResolve = require('rollup-plugin-node-resolve');
+const jsonPlugin = require('rollup-plugin-json');
+const babel = require('rollup-plugin-babel');
+
+const onwarn = warning => {
```
daprahamian: What is this doing?

tools/loadTestFiles.js
```diff
@@ -0,0 +1,25 @@
...
+const bugDataObj = { data: gsWeirdBugData };
+
+const scenarios = fs
+  .readdirSync(path.join(__dirname, '../test/node/specs/bson-corpus'))
+  .filter(x => x.indexOf('json') !== -1)
+  .map(x =>
+    JSON.parse(fs.readFileSync(path.join(__dirname, '../test/node/specs/bson-corpus', x), 'utf8'))
+  );
+
+fs.writeFile('./tools/scenarios.json', JSON.stringify(scenarios, null, 2), function(err) {
```
daprahamian: If there is an error with these, do you propagate an error code? It might be good to `process.exit(errCode);` if one of these writes fails.

tools/deleteTestFiles.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+const fs = require('fs');
+const paths = ['./tools/scenarios.json', './tools/gsWeirdBugData.json'];
+
+paths.forEach(path => {
+  fs.stat(path, function(err) {
```
daprahamian: See comment about `process.exit` below

tools/run_tests.sh
```diff
@@ -0,0 +1,36 @@
...
+
+fi
+
+exit_code=$?
+
+# run cleanup script no matter what
+node ./tools/deleteTestFiles.js
+
+# exit with code if not ok
+if [[ $exit_code != 0 ]] ; then exit $exit_code ; fi
```
daprahamian: what does `exit 0` do? Could we just always exit with `$exit_code`?

tools/run_tests.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+ # run setup script
+node ./tools/loadTestFiles.js &&
```
daprahamian: I'd store the executables for `node` in a variable so that if we need an absolute path on some system, we can use `$node`

tools/run_tests.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+ # run setup script
+node ./tools/loadTestFiles.js && 
+
+# if no arguments run both tests
+if [ $# -eq 0 ]; then
+
+    # run node tests
+    mocha ./test/node
```
daprahamian: Can you execute this via `npx`?

tools/run_tests.sh
```diff
@@ -0,0 +1,36 @@
...
+
+    node_exit_code=$?
+
+    # if node exit status not ok, run cleanup and exit with code
+    if [[ $node_exit_code != 0 ]] ; then node ./tools/deleteTestFiles.js && exit $node_exit_code ; fi
+
+    # run karma tests
+    ./node_modules/karma/bin/karma start karma.conf.js
```
daprahamian: Can this be executed via `npx`?

karma.conf.js
```diff
@@ -0,0 +1,39 @@
...
+const rollupConfig = require('./rollup.config.js');
+delete rollupConfig.input;
+
+// Karma configuration
+// Generated on Thu Jun 28 2018 14:24:01 GMT-0400 (EDT)
+
+module.exports = function(config) {
+  config.set({
+    basePath: '',
+
```
daprahamian: Can we remove the spaces between lines?

tools/run_tests.sh
shsaskin: the original script used "mocha ./test/node"
Node modules doesn't have a mongodb-test-runner in js-bson. I could do "npx mocha ./test/node", but I don't know if that's what you're looking for.

rollup.config.js
rweinberger: The tests generated a lot of noisy & repetitive warnings that we wanted to silence - specifically, warnings for circular dependencies and invocations of `eval`. 

When we ran the tests, we were getting an entire pageful of repetitive Circular Dependency warnings originating from within `node_modules` that looked like:

`Circular dependency: node_modules/chai/lib/chai.js -> node_modules/chai/lib/chai/utils/index.js -> node_modules/chai/lib/chai/utils/addMethod.js -> node_modules/chai/lib/chai.js`

These circular dependencies were within modules like chai, so we deemed them non-concerns. Similarly, the following warning about using `eval` was printed over 20 times:

`lib/bson/parser/deserializer.js (506:4) Use of eval is strongly discouraged, as it poses security risks and may cause issues with minification`

We decided to silence this one as well. Are the concerns about using `eval` valid for in-browser use? Should we be concerned?

daprahamian: Ok, that makes sense. Can you move this to the `karma.conf.js`, so it is only added for the tests, and not for the build step?

tools/run_tests.sh
daprahamian: Yes, I'd like to see all of the scripts that are installed with `npm` use `npx` to invoke them.

shsaskin: I updated the node and karma tests to use npx, but now the node 4 and 6 tests fail in travis because they don't have npx.

test/node/bson_test.js
```diff
@@ -19,7 +18,8 @@ var Buffer = require('buffer').Buffer,
   MinKey = BSON.MinKey,
   MaxKey = BSON.MaxKey,
   BinaryParser = require('../binary_parser').BinaryParser,
-  vm = require('vm');
+  vm = require('vm'),
```
daprahamian: Since we are here, let's make these `const`, and one import per statement.

tools/scenarios-plugin.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+
+const scenariosPlugin = () => {
```
daprahamian: This is wonderful. I love this.

rollup.config.js
```diff
@@ -1,21 +1,29 @@
...
+module.exports = {
   input: 'index.js',
   output: {
     file: 'dist/bson.js',
     format: 'umd',
-    name: 'bson',
+    name: 'bson'
   },
   plugins: [
+    scenariosPlugin(),
```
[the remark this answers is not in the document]
rweinberger: Makes sense! We made a modification so that the scenariosPlugin is only added to the plugins array in the karma config, so the rollup config is essentially left unchanged.

rollup.config.js
```diff
@@ -1,16 +1,20 @@
...
+const nodeBuiltins = require('rollup-plugin-node-builtins');
+const nodeResolve = require('rollup-plugin-node-resolve');
+const babel = require('rollup-plugin-babel');
 
 const input = 'index.js';
 const plugins = [
   nodeResolve(),
   commonjs(),
   nodeBuiltins(),
+  nodeGlobals(),
```
edaniels: It seems like this shouldn't be added back in. This is going to polyfill things that don't need to be there based on latest changes.

dist/bson.bundle.js
```diff
@@ -3002,6 +3002,225 @@ var timestamp = Timestamp;
 var Timestamp_1 = Timestamp;
 timestamp.Timestamp = Timestamp_1;
 
+var global$1 = typeof global !== "undefined" ? global : typeof self !== "undefined" ? self : typeof window !== "undefined" ? window : {};
```
edaniels: I wouldn't expect these to show up or be wanted by users in this bundle file. Is it a result of testing?

rollup.config.js
```diff
@@ -1,16 +1,18 @@
...
+
+const pkg = require('./package.json');
+const commonjs = require('rollup-plugin-commonjs');
+const nodeBuiltins = require('rollup-plugin-node-builtins');
+const nodeResolve = require('rollup-plugin-node-resolve');
+const babel = require('rollup-plugin-babel');
 
 const input = 'index.js';
 const plugins = [
+  nodeBuiltins(),
```
edaniels: Just a reminder to remove these since they are only needed in karma

rweinberger: Just wanted to confirm, both nodeGlobals and nodeBuiltins are karma-only? We hesitate because we see nodeBuiltins in the rollup config on master at the moment

edaniels: Ah yes. Builtins should be lower, my bad. nodeGlobals isn't needed and you don't have it here anyway so that's good. This file should be unmodified from master and by assocation so should dist/bson.bundle.js.

rweinberger: Thanks for clarifying, will re-order to match master!

test/node/bson_array_test.js
```diff
@@ -57,30 +57,30 @@
...
- * instance of the db client
- *
- * @ignore
- */
-exports.setUp = function(callback) {
-  _Uint8Array = global.Uint8Array;
-  delete global['Uint8Array'];
-  callback();
-};
+// /**
```
daprahamian: these guys are still commented out. Can we either uncomment them or determine if we need them to begin with?

PR #260, opened 2018-07-08
fix(rollup): fixup dist and bundle in Buffer

edaniels:


.gitignore
```diff
@@ -13,4 +13,7 @@ npm-debug.log
 builderror.log
 
 bson.sublime-project
-bson.sublime-workspace
\ No newline at end of file
+bson.sublime-workspace
+
+dist/*
```
daprahamian: Do we want to permanently remove the `dist` directory and only have the files in the published version?

README.md
```diff
@@ -18,7 +18,7 @@ npm run build
 A simple example of how to use BSON in the browser:
 
 ```html
-<script src="./dist/bson.js"></script>
+<script src="./dist/bson.bundle.js"></script>
```
daprahamian: Can the `README.md` include more documentation about what each `dist` file is?

lib/bson/bson.js
```diff
@@ -383,6 +384,7 @@ module.exports.Symbol = Symbol;
 module.exports.BSON = BSON;
 module.exports.DBRef = DBRef;
 module.exports.Binary = Binary;
+module.exports.ObjectId = ObjectID;
```
daprahamian: This seems like an unrelated change.

lib/bson/parser/deserializer.js
```diff
@@ -27,17 +28,18 @@ var deserialize = function(buffer, options, isArray) {
     throw new Error(`bson size must be >= 5, is ${size}`);
   }
 
-  if (options.allowObjectSmallerThanBufferSize && buffer.length < size) {
-    throw new Error(`buffer length ${buffer.length} must be >= bson size ${size}`);
+  if (options.allowObjectSmallerThanBufferSize && Buffer.byteLength(buffer) < size) {
```
daprahamian: Why `Buffer.byteLength`?

lib/bson/parser/deserializer.js
```diff
@@ -27,17 +28,18 @@ var deserialize = function(buffer, options, isArray) {
...
+    throw new Error(`buffer length ${Buffer.byteLength(buffer)} must be >= bson size ${size}`);
   }
 
-  if (!options.allowObjectSmallerThanBufferSize && buffer.length !== size) {
-    throw new Error(`buffer length ${buffer.length} must === bson size ${size}`);
+  if (!options.allowObjectSmallerThanBufferSize && Buffer.byteLength(buffer) !== size) {
+    console.log(buffer, buffer.length, Buffer.byteLength(buffer));
```
daprahamian: `console.log`

.gitignore
edaniels: I think the bundle is the one good file to keep here

README.md
edaniels: I don't think it's necessary since only 1 is committed. It should be self documenting from package.json.

lib/bson/bson.js
edaniels: It's related in removing confusing duplication in index.js. This is the one diffed thing not present.

lib/bson/parser/deserializer.js
edaniels: `length` is not a real property on `Buffer`. It comes from the underlying Uint8Array which is not always consistent depending on the environment. `byteLength` will always produce the correct result which will be the same as `length`.

lib/bson/parser/deserializer.js
edaniels: whoops

edaniels:
Changes addressed
@daprahamian AFAICT, Buffer does not get globally assigned if you look at the output of each dist. It appears to be used locally. Even if it did, a good citizen polyfill would be the same API and functionality and would only polyfill in the event it's already present.

edaniels:
I've also loaded both bson and mongodb-extjson in the same page with 0 errors.

README.md
daprahamian: Since I don't fully understand it, I think it would be helpful to document what each bundle is in the readme.

edaniels: Done.

PR #261, opened 2018-07-11
test(windows): make tests work on windows again

mbroadst:
This removes the `run-tests.sh` script in favor of executing tests
through package.json scripts. By default only mocha tests are run
locally, and the `test-browser` target now builds the bundles and
runs karma tests against them. Checks for node v4 have been moved
to the travis configuration, soon to be removed outright due to
v4 deprecation.


mbroadst:
@daprahamian why rewrite `run_tests.sh` in node?

mbroadst:
Okay, the default `test` target now tests both

PR #265, opened 2018-07-16
NODE-1568 modernize and clean entire module

mbroadst:
**NOTE**: this pull request is rather difficult to parse holistically, I recommend taking a look at the changes on a per-commit basis. 

This is a rather dramatic departure from the original design of `js-bson` in anticipating of the next breaking major release. General themes include

- modernization (use es6 everywhere possible, `var` => `let`, `class` etc)
- code deduplication (remove vendored-in old long.js implementation, add dependency on community module. dedupe use of constants)
- organization (the top level exports of the module are now very well defined, and shouldn't cause problems for bundling)
- BSON is no longer a function, there is no need to `const bson = new BSON([ .. list of types .. ])`. You simply do:
```
const BSON = require('bson');
const long = BSON.Long.fromNumber(10);
const data = BSON.serialize({ llamas: 'are cool' });
const size = BSON.calculateObjectSize({ another: 'test' });
```

This is the major batch of changes required for the new version, but subsequent PRs will be required in order to update documentation where API has changed.

/cc @edaniels


lib/bson.js
```diff
@@ -176,212 +191,266 @@ BSON.prototype.calculateObjectSize = function(object, options) {
...
+  BSON_BINARY_SUBTYPE_USER_DEFINED,
+
+  // wrapped types
+  Code,
+  Map,
+  Symbol,
+  DBRef,
+  Binary,
+  ObjectID,
+  ObjectId: ObjectID,
```
edaniels: If we're going to bump the major version, maybe now's a good time to drop this alias?

lib/bson.js
```diff
@@ -213,222 +214,7 @@ function deserializeStream(data, startIndex, numberOfDocuments, documents, docSt
...
-  BSON_DATA_MIN_KEY,
-  BSON_DATA_MAX_KEY,
-  BSON_BINARY_SUBTYPE_DEFAULT,
-  BSON_BINARY_SUBTYPE_FUNCTION,
-  BSON_BINARY_SUBTYPE_BYTE_ARRAY,
-  BSON_BINARY_SUBTYPE_UUID,
-  BSON_BINARY_SUBTYPE_MD5,
-  BSON_BINARY_SUBTYPE_USER_DEFINED,
-
+module.exports = Object.assign(constants, {
```
edaniels: This _may_ mess with bundlers trying to read the exported names.

lib/bson.js
mbroadst: Yeah good idea

lib/bson.js
mbroadst: Yeah I thought maybe that would be the case, if we actually drop Node 4 support we can just use the spread operator here which might make it nicer

daprahamian: This can be fixed with `Object.assign({}, constants, {...})`.

.gitignore
```diff
@@ -13,4 +13,6 @@ npm-debug.log
 builderror.log
 
 bson.sublime-project
-bson.sublime-workspace
\ No newline at end of file
+bson.sublime-workspace
+
+dist
```
daprahamian: Be careful with this. This will make npm ignore the dist folder when publishing, so you have to explicitly include the dist folder in the package.json

lib/parser/deserializer.js
```diff
@@ -49,71 +51,71 @@ var deserialize = function(buffer, options, isArray) {
...
-  var cacheFunctions = options['cacheFunctions'] == null ? false : options['cacheFunctions'];
-  var cacheFunctionsCrc32 =
+function deserializeObject(buffer, index, options, isArray) {
+  const evalFunctions = options['evalFunctions'] == null ? false : options['evalFunctions'];
+  const cacheFunctions = options['cacheFunctions'] == null ? false : options['cacheFunctions'];
+  const cacheFunctionsCrc32 =
     options['cacheFunctionsCrc32'] == null ? false : options['cacheFunctionsCrc32'];
 
-  if (!cacheFunctionsCrc32) var crc32 = null;
+  // if (!cacheFunctionsCrc32) var crc32 = null;
```
daprahamian: Commented out code

lib/db_ref.js
```diff
@@ -1,44 +1,47 @@
...
-  var o = {
-    $ref: this.collection,
-    $id: this.oid
-  };
+  /**
+   * @ignore
+   * @api private
+   */
+  toJSON() {
+    const o = Object.assign(
```
daprahamian: This changes the order of assignment such that `this.db` comes after `this.fields`. Is that expected?

lib/bson.js
```diff
@@ -1,36 +1,48 @@
...
  * @ignore
  * @api private
  */
 // Default Max Size
-var MAXSIZE = 1024 * 1024 * 17;
+const MAXSIZE = 1024 * 1024 * 17;
 
 // Current Internal Temporary Serialization Buffer
-var buffer = new Buffer(MAXSIZE);
+let buffer = new Buffer(MAXSIZE);
```
daprahamian: Are we allocating 17 MB of data here?

.gitignore
mbroadst: Nice call, that seems like non ideal default behavior of npm but I do indeed see it in the docs that `.gitignore` will be used in the absence of a `.npmignore` file. I'll remove it here, but I just want to make sure that we're not committing the bundles to the git repo anymore - they should all only be packaged with the distributed tarball on npm.

lib/db_ref.js
mbroadst: It's different, but I'm having trouble figuring out when you would want to do this or why order of assignment matters? Do you have something in mind?

lib/bson.js
mbroadst: We are! This is actually in line with the server code, and generally is over 16MB so we can accommodate the over head of message envelopes to the server, particular bulk cases

lib/binary.js
```diff
@@ -0,0 +1,377 @@
...
+      buffer != null &&
+      !(typeof buffer === 'string') &&
+      !Buffer.isBuffer(buffer) &&
+      !(buffer instanceof Uint8Array) &&
+      !Array.isArray(buffer)
+    ) {
+      throw new Error('only String, Buffer, Uint8Array or Array accepted');
+    }
+
+    if (buffer instanceof Number) {
```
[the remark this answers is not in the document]
mbroadst: Nice catch! This looks like a shortcut for creating a buffer by subtype, but it is never tested and the throw before it has been there for some time. I'll remove the case.

lib/db_ref.js
daprahamian: I honestly don't expect it to be any different, I just wanted to note that it is. Hypothetically, someone could have `this.fields.$db` that they expect to overrride the value of `this.db`. But that is unlikely.

lib/bson.js
daprahamian: It seems like  a lot to preallocate. Especially on the web, when even the heaviest web pages are usually around 40Mb of heap space.

mbroadst: That's a good point, especially considering this is brought in as a dependency for e.g. stitch's js SDK. I've created [NODE-1599](https://jira.mongodb.org/browse/NODE-1599) to track this.

.gitignore
daprahamian: So I've confirmed that as long as the `files` array in `package.json` includes `"dist"`, it will still pack up the `dist` folder. It should be safe to add this back in.

lib/bson.js
mbroadst: okay fixed

lib/bson.js
mbroadst: I did the work to rename everything internally to use the preferred `ObjectId`, but I just can't quite bite the bullet dropping the alias. It's so cheap for us, and it has such potential to break everyone's code. If everyone feels very strongly that this is the right path I'd reconsider, but otherwise it feels like we're begging for bug reports.

daprahamian: Add a deprecated getter to return it, so people get a deprecation message?

PR #267, opened 2018-08-02
fix(serializer): map insert expects only string keys

shsaskin:
Checks that the key is a string before checking that
the key is a valid string to prevent type error.

Fixes NODE-1572


lib/parser/serializer.js
```diff
@@ -785,7 +785,7 @@ function serializeInto(
       const type = typeof value;
 
       // Check the key and throw error if it's illegal
-      if (key !== '$db' && key !== '$ref' && key !== '$id') {
+      if (key !== '$db' && key !== '$ref' && key !== '$id' && typeof key === 'string') {
```
mbroadst: it looks like this needs to be merged with #269

mbroadst: also requires modification to L888

lib/parser/serializer.js
```diff
@@ -786,7 +786,7 @@ function serializeInto(
       const type = typeof value;
 
       // Check the key and throw error if it's illegal
-      if (!ignoreKeys.has(key)) {
+      if (!ignoreKeys.has(key) && typeof key === 'string') {
```
mbroadst: prefer to check type first, before iterating through the set. This is a case where we can ensure monomorphic access to the `has` function.

PR #269, opened 2018-08-03
fix(serializer): do not use checkKeys for $clusterTime

kvwalker:
Fixes NODE-1409


lib/parser/serializer.js
```diff
@@ -10,6 +10,7 @@ const constants = require('../constants');
 const normalizedFunctionString = require('./utils').normalizedFunctionString;
 
 const regexp = /\x00/; // eslint-disable-line no-control-regex
+const ignoreKeys = ['$db', '$ref', '$id', '$clusterTime'];
```
daprahamian: Can we use a `Set` for this instead?

PR #270, opened 2018-08-08
fix(serializer): do not use checkKeys for $clusterTime

kvwalker:
Fixes NODE-1409


lib/bson/parser/serializer.js
```diff
@@ -15,6 +15,7 @@ var normalizedFunctionString = require('./utils').normalizedFunctionString;
 // }
 
 var regexp = /\x00/; // eslint-disable-line no-control-regex
+const ignoreKeys = new Set(['$db', '$ref', '$id', '$clusterTime']);
```
mbroadst: can't use `Set` in the 1.x branch, since it supports versions of node older than 4. Use an array instead, and convert the `has` calls below to `indexOf` checks

PR #273, opened 2018-08-30
fix(serializer): only check $ for inner keys

kvwalker:
Fixes NODE-1645

### DIFF
--- a/lib/parser/serializer.js
+++ b/lib/parser/serializer.js
@@ -794,7 +794,7 @@ function serializeInto(
         }
 
         if (checkKeys) {
-          if ('$' === key[0]) {
+          if ('$' === key[0] && depth === 0) {
             throw Error('key ' + key + " must not start with '$'");
           } else if (~key.indexOf('.')) {
             throw Error('key ' + key + " must not contain '.'");
@@ -894,7 +894,7 @@ function serializeInto(
         }
 
         if (checkKeys) {
-          if ('$' === key[0]) {
+          if ('$' === key[0] && depth === 0) {
             throw Error('key ' + key + " must not start with '$'");
           } else if (~key.indexOf('.')) {
             throw Error('key ' + key + " must not contain '.'");


BastienAr:
I think you can also drop

```javascript
else if (~key.indexOf(".")) {
   throw Error("key " + key + " must not contain '.'");
}
```

Since the [3.6](https://docs.mongodb.com/v3.6/reference/limits/) version, fields with dot are allowed

codepushr:
Has this not been merged? The 4.0.0 version still has a check for dots although the official documentation states that it is allowed already.

daprahamian:
@codepushr unfortunately this feature will not be incorporated at this time. From [the official documentation on Documents](https://docs.mongodb.com/manual/core/document/#field-names)

> The MongoDB Query Language cannot always meaningfully express queries over documents whose field names contain these characters (see [SERVER-30575](https://jira.mongodb.org/browse/SERVER-30575)).
>
>Until support is added in the query language, the use of $ and . in field names is not recommended and is not supported by the official MongoDB drivers.

Since this can be a major footgun, we won't be supporting keys with `$` and `.`. That said, if you need to work around this restriction in the `bson` library, you can set the option `checkKeys` to `false`:

```js
bson.serialize({ foo: { "$bar": 1 }, "fizz.buzz": 2 }, { checkKeys: false });
```

This is also possible in the `mongodb` driver:

```js
const { MongoClient } = require("mongodb");
const client = new MongoClient('mongodb://localhost:27017/', { useNewUrlParser: true });
const inner$ = { a: { $b: 1 } };

client.connect().then(async function() {
  ...
  const collection = client.db('db').collection('col');
  await collection.insertOne(inner$, { checkKeys: false })
  ...
});
```

Keep in mind that if you do so, you run the risk of not being able to query on these fields.

For more context, you can refer to the following tickets:
+ [SERVER-30575](https://jira.mongodb.org/browse/SERVER-30575)
+ [SERVER-30417](https://jira.mongodb.org/browse/SERVER-30417)
+ [NODE-1645](https://jira.mongodb.org/browse/NODE-1645?focusedCommentId=1990165&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-1990165)
+ [JAVA-2810](https://jira.mongodb.org/browse/JAVA-2810?focusedCommentId=2117856&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-2117856)

codepushr:
Thanks for the explanation.

PR #276, opened 2018-09-29
feat(ext-json): merge extended JSON into js-bson

mbroadst:
This is a big code merge, but I believe it would dramatically reduce our maintenance load and we are already making a major version bump here. Generally the code changes would go from:
```
const EJSON = require('mongodb-extjson');
EJSON.stringify(...);
EJSON.parse(...);
```
to
```
const { EJSON } = require('bson');
EJSON.stringify(...);
EJSON.parse(...);
```

### DIFF
--- a/lib/binary.js
+++ b/lib/binary.js
@@ -274,6 +274,32 @@ class Binary {
   toString(format) {
     return this.buffer != null ? this.buffer.slice(0, this.position).toString(format) : '';
   }
+
+  /**
+   * @ignore
+   */
+  toExtendedJSON() {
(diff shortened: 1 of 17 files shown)


mbroadst:
@daprahamian for starters not at all. `bson-ext` would have to bump its dependency on js-bson to 4.x in order to bring in any changes, but once it does it will get extended JSON support for free. Since `bson-ext` now just pulls in all of `js-bson`, and overwrites the serialization methods, this change will have no impact on it other than to provided more functionality.

PR #280, opened 2018-10-30
fix(deserialize): fix deserialization of 0xFFFD

### LINKED ISSUE #277: Error 'Invalid UTF-8 string in BSON document' when deserializing a previously serialized BSON document
The following code throws an error on the `bson.deserialize(...)` call:

```javascript
const BSON = require("bson");
const bson = new BSON();

const offendingString = String.fromCharCode(0x41, 0x42, 0xFFFD, 0x43, 0x44); // "AB�CD"

const serialized = bson.serialize({ "value": offendingString });
const deserialized = bson.deserialize(serialized);
```

The error that is thrown is 'Invalid UTF-8 string in BSON document'. From [the source](https://github.com/mongodb/js-bson/blob/v3.0.2/lib/bson/parser/deserializer.js#L130) it appears that the `deserialize()` function specifically checks for the `0xFFFD` character and, if present, throws this error.

Why is a UTF-8 string containing this character considered invalid? And why doesn't `serialize()` already throw when serializing it in the first place? This seems like inconsistent behavior, in the sense that the BSON library allows you to serialize invalid data, but not deserialize it later, causing issues down the line when reading previously stored data.

daprahamian:
When deserializing a string, we previously relied on the appearance
of 0xFFFD to denote an invalid unicode character. However, 0xFFFD
is a valid unicode character if that is what was originally
input.

Fixes NODE-1718
Fixes #277

### DIFF
--- a/lib/validate_utf8.js
+++ b/lib/validate_utf8.js
@@ -1,0 +1,48 @@
+'use strict';
+
+const FIRST_BIT = 0x80;
+const FIRST_TWO_BITS = 0xc0;
+const FIRST_THREE_BITS = 0xe0;
+const FIRST_FOUR_BITS = 0xf0;
+const FIRST_FIVE_BITS = 0xf8;
+
+const TWO_BIT_CHAR = 0xc0;
+const THREE_BIT_CHAR = 0xe0;
+const FOUR_BIT_CHAR = 0xf0;
+const CONTINUING_CHAR = 0x80;
+
+/**
+ * Determines if the passed in bytes are valid utf8
+ * @param {Buffer|Uint8Array} bytes An array of 8-bit bytes. Must be indexable and have length property
+ * @param {Number} start The start index to start validating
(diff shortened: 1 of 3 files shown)


lib/validate_utf8.js
```diff
@@ -0,0 +1,48 @@
...
+const TWO_BIT_CHAR = 0xc0;
+const THREE_BIT_CHAR = 0xe0;
+const FOUR_BIT_CHAR = 0xf0;
+const CONTINUING_CHAR = 0x80;
+
+/**
+ * Determines if the passed in bytes are valid utf8
+ * @param {Buffer|Uint8Array} bytes An array of 8-bit bytes. Must be indexable and have length property
+ * @param {Number} start The start index to start validating
+ * @param {Number} end The end index to start validating
```
kvwalker: `start index` should be `end index`

PR #282, opened 2018-11-02
ObjectId.isValid should check buffter length

clarkttfu:
```
const buff = Buffer.from([1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13])

// returns true
ObjectId.isValid(buff) 
// TypeError: Argument passed in must be a single String of 12 bytes or a string of 24 hex characters
new ObjectId(buff)
```

### DIFF
--- a/lib/objectid.js
+++ b/lib/objectid.js
@@ -328,7 +328,7 @@ class ObjectId {
       return true;
     }
 
-    if (id instanceof _Buffer) {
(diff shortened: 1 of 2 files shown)


mbroadst:
Hi @clarkttfu, this looks good but would you be willing to provide a test as well?

clarkttfu:
Hi @mbroadst, test case committed, please review.

PR #299, opened 2019-01-13
Fix: deprecation warnings for buffers and util.inspect

karimsa:
Node has deprecated the `inspect` property on objects for use in `util.inspect()`. Doing so on an ObjectID on Node v10 prints a deprecation warning:

```
(node:16945) [DEP0079] DeprecationWarning: Custom inspection function on Objects via .inspect() is deprecated
```

This fixes it to use the new symbol that is exposed by `util.inspect` instead - and fallback to the string property for old node versions.

### DIFF
--- a/lib/binary.js
+++ b/lib/binary.js
@@ -297,7 +297,7 @@ class Binary {
    */
   static fromExtendedJSON(doc) {
     const type = doc.$binary.subType ? parseInt(doc.$binary.subType, 16) : 0;
-    const data = new Buffer(doc.$binary.base64, 'base64');
+    const data = Buffer.from(doc.$binary.base64, 'base64');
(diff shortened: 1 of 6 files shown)


mbroadst:
Hi @karimsa, thanks for the contribution! Would you mind including a test here verifying that this is fixing something?

karimsa:
This doesn't really cause any type of error - just a warning (unless `--throw-deprecation` is used with node). But I can't add that to the test runner because 702 other tests fail due to other forms of deprecations (like `new Buffer()`).

karimsa:
I've added a fix for the buffer constructor deprecation and added the `--throw-deprecation` flag to the test running - without the fixes, the tests fail.

PR #300, opened 2019-02-06
Fix EJSON.stringify if replacer is array (not function)

justingrant:
Currently, passing an array form of the `replacer` parameter of `EJSON.stringify` will not work, despite the documentation saying that it should work.  The cause is that the current implementation checks whether `replacer` is of type `object`, and if it is then it assumes that the `replacer` parameter is really the `options` parameter (meaning that `replacer` and `space` have been omitted by the user).  This logic breaks when `replacer` is an array (which is an `object` too). 

To fix this problem I added a check using `Array.isArray` so the replacement above will only happen if `replacer` is an object that's not an array. 

While I was editing this method, I also changed the weird (and eslint-defying) use of assignment expressions like this: 
`if (space != null && typeof space === 'object') (options = space), (space = 0);` 
to a more conventional statement form of assignment.

This PR doesn't conflict or overlap with two other PRs I filed today: #304 and #305.

### DIFF
--- a/lib/extended_json.js
+++ b/lib/extended_json.js
@@ -251,16 +251,6 @@ function serializeValue(value, options) {
     return { $numberDouble: value.toString() };
   }
 
-  if (value instanceof RegExp) {
-    let flags = value.flags;
-    if (flags === undefined) {
-      flags = value.toString().match(/[gimuy]*$/)[0];
-    }
-
-    const rx = new BSONRegExp(value.source, flags);
-    return rx.toExtendedJSON();
-  }
-
   if (value != null && typeof value === 'object') return serializeDocument(value, options);
   return value;
 }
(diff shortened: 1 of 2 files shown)


mbroadst:
Hey @justingrant! Thanks for the fix, would you mind contributing a test case exemplifying the problem  and proving it's fixed?

justingrant:
Added tests but made a mistake and test is failing. Will fix shortly. Stay tuned.

justingrant:
OK @mbroadst - tests are checked in and working

justingrant:
Added another checkin that fixes #302 and fixes #303.  Also added tests to validate.  If you want me to split those fixes out into a separate PR, that's fine, just let me know.

mbroadst:
Hey @justingrant, yeah can you please split those two out into their own PR(s)? It will make it much easier to review and merge in, it looks like I would have some questions about some of the added on work here.

justingrant:
@mbroadst - Yep, not a problem. I reverted those changes and opened #304 and #305. Let me know if you want to see changes in any of these three PRs.

mbroadst:
@justingrant thanks! This one looks good, merging now

PR #304, opened 2019-02-11
Enable EJSON serialization of ObjectID (capital "D")

justingrant:
This PR fixes #303 - ObjectID (capital "D") instances can't be serialized to Extended JSON. 
It should also fix #294 as well as https://jira.mongodb.org/browse/NODE-1779 which look like the same issue as #303.  (Sorry I didn't notice #294 before opening 303.  I only looked for dupes among open issues which was a mistake.)

The EJSON serializer: in 4.0.1 doesn't correctly serialize instances of ObjectID (capital "D") used by the official`'mongodb'` package. For example, running `EJSON.serialize` on the output of `db.coll.findOne()` will throw an exception.

BTW, this only affects the Extended JSON serializer. The BSON serializer was covered in #290.

For testing, I leveraged existing test cases for `ObjectID` and extended them with side-by-side comparisons of all three possible variations of object IDs to make sure that all three work OK:
* `BSON.ObjectId` - this is the "main" object ID class going fwd. 
* `BSON.ObjectID` - this is an alias for ObjectId.  Testing this guarantees that the alias is working.  Using this alias was also how the tests were written previously so I wanted to avoid removing working tests.
* `require('mongodb').ObjectID` - this validates the code changes in this PR. Because the `mongdb` package isn't available in the browser, browser tests fall back to `BSON.ObjectId` instead. 

I also noticed a potential problem where unrecognized `_bsontype` isn't being handled. I marked this with a TODO comment for dealing with in a later PR. 

The changes in this PR should be compatible (no merge conflicts) with the other PRs I opened recently.

PR #305, opened 2019-02-11
Stop running Extended JSON serializer twice for each nested document

### LINKED ISSUE #302: Extended JSON serializer runs twice for every nested document
A missing `else` is causing EJSON serialization to run 2x for every nested document. The first result is being thrown away, and the code whose results are ignored is duplicated elsewhere. Here's the code. Note that `_doc[name]` is being assigned twice. 

```js
  // the document is an object with nested BSON types
  const _doc = {};
  for (let name in doc) {
    let val = doc[name];
    if (Array.isArray(val)) {
      _doc[name] = serializeArray(val, options);
    } else if (val != null && typeof val.toExtendedJSON === 'function') {
      if (val._bsontype === 'Code' && val.scope) {
        val.scope = serializeDocument(val.scope, options);
      } else if (val._bsontype === 'DBRef' && val.oid) {
        val.oid = serializeDocument(val.oid, options);
      }

      _doc[name] = val.toExtendedJSON(options);
    } else if (val instanceof Date) {
      _doc[name] = serializeValue(val, options);
    } else if (val != null && typeof val === 'object') {
      _doc[name] = serializeDocument(val, options);
    }
    _doc[name] = serializeValue(val, options);
    if (val instanceof RegExp) {
      let flags = val.flags;
      if (flags === undefined) {
        flags = val.toString().match(/[gimuy]*$/)[0];
      }

      const rx = new BSONRegExp(val.source, flags);
      _doc[name] = rx.toExtendedJSON();
    }
  }
```

Here's the 2017 commit where it was introduced: https://github.com/mongodb-js/mongodb-extjson/commit/66fe3d98aba1659ba7c3ca44cc63ab6100c1b9ce#diff-b87f4aa0da50c93cafec959192135a40.

It seems likely that fixing this will speed up serialization for large nested documents. 

I'm working on a PR.

justingrant:
This PR fixes #302 (Extended JSON serializer runs twice for every nested document) by refactoring EJSON serialization to remove the ignored first serialization and to remove the duplicated code that was producing the ignored first serialization.

Although all existing tests still pass, I didn't add any new tests because the only externally-visible change from this PR should be faster serialization-- the library itself should work the same.

I also noticed two other potential problems that I didn't know how to fix, so I added TODO comments that can be addressed in later PRs.

This PR's code changes overlap with #304, but the overlapping lines are the same in both--and neither PR is dependent on the other--so they should merge together or individually without conflicts.

### DIFF
--- a/lib/extended_json.js
+++ b/lib/extended_json.js
@@ -282,8 +282,7 @@ function serializeDocument(doc, options) {
   // Recursively serialize this document's property values. 
   const _doc = {};
   for (let name in doc) {
-    const val = doc[name];
-    _doc[name] = serializeValue(val, options);
+    _doc[name] = serializeValue(doc[name], options);
   }
 
   return _doc;
--- a/lib/extended_json.js
+++ b/lib/extended_json.js
@@ -245,6 +245,16 @@ function serializeValue(value, options) {
     return { $numberDouble: value.toString() };
   }
 
+  if (value instanceof RegExp) {
+    let flags = value.flags;
+    if (flags === undefined) {
+      flags = value.toString().match(/[gimuy]*$/)[0];
+    }
+
+    const rx = new BSONRegExp(value.source, flags);
+    return rx.toExtendedJSON();
+  }
+
   if (value != null && typeof value === 'object') return serializeDocument(value, options);
   return value;
 }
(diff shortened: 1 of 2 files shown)


lib/extended_json.js
```diff
@@ -245,53 +245,45 @@ function serializeValue(value, options) {
...
-        val.oid = serializeDocument(val.oid, options);
-      }
-
-      _doc[name] = val.toExtendedJSON(options);
-    } else if (val instanceof Date) {
-      _doc[name] = serializeValue(val, options);
-    } else if (val != null && typeof val === 'object') {
-      _doc[name] = serializeDocument(val, options);
-    }
+    const val = doc[name];
```
mbroadst: this begs the question whether we need to pull this value at all anymore? I believe it was pulled out originally because it was used so often in the code you are removing here.

mbroadst:
Thanks @justingrant!

justingrant:
@mbroadst - Happy to help!  Sorry I'm bombarding you with so many issues and PRs-- and thanks for the fast turnaround on getting these merged.

PR #309, opened 2019-02-16
Fix getTimestamp for times beyond 2038

autopulated:
Avoid treating timestamps with the most significant bit set as negative. The mongodb documentation describe the timetsamp portion of objectids as 'seconds since the unix epoch', so this seems to be the correct behaviour, although I cannot find a formal specification for it.

However this is technically a breaking change if people were relying on timestamp values before 1970.

### DIFF
--- a/lib/objectid.js
+++ b/lib/objectid.js
@@ -254,7 +254,7 @@ class ObjectId {
     // 2038) we must be careful to avoid creating a negative time. To do that,
     // use only the high 31 bits of the 32 bit timestamp, then add the least
     // significant bit after converting to a number:
-    const time = 2 * ((this.id[3] >>> 1) | (this.id[2] << 7) | (this.id[1] << 15) | (this.id[0] << 23)) + (this.id[3] & 1);
+    const time = this.id.readUInt32BE(0);
     timestamp.setTime(Math.floor(time) * 1000);
     return timestamp;
   }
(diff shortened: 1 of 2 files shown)


lib/objectid.js
```diff
@@ -249,7 +249,12 @@ class ObjectId {
    */
   getTimestamp() {
     const timestamp = new Date();
-    const time = this.id[3] | (this.id[2] << 8) | (this.id[1] << 16) | (this.id[0] << 24);
+    // javascript performs bitwise operations as if the values are signed 32
+    // bit integers - so if the highest bit of this.id[0] is set (times beyond
+    // 2038) we must be careful to avoid creating a negative time. To do that,
+    // use only the high 31 bits of the 32 bit timestamp, then add the least
+    // significant bit after converting to a number:
+    const time = 2 * ((this.id[3] >>> 1) | (this.id[2] << 7) | (this.id[1] << 15) | (this.id[0] << 23)) + (this.id[3] & 1);
```
daprahamian: We can achieve the same effect by running the signed int32 through `>>> 0` like this:

```js
let time = (this.id[3] | (this.id[2] << 8) | (this.id[1] << 16) | (this.id[0] << 24));
time >>>= 0;
```

IMO that is a little clearer.

Honestly, I'd prefer to use the built-in Buffer method here if possible. Could we leverage `Buffer.prototype.readUInt32LE`?

autopulated: 👍 It's big endian, so I used `readUInt32BE`, but that is much neater yes!

mbroadst:
Thanks for the fix @autopulated !

Issue #310, opened 2019-02-17
Library 4.x can't send MinKey to MongoDB


justingrant:
The 4.x library is not capable of sending a MinKey instance to MongoDB, because of a bug in the 1.x library that I don't think we can work around on the 4.x side.  The user-facing result is that filters for MinKey will be sent to the server as MaxKey instead, which could cause unexpected query results, unexpected deletion, etc.

The problem is in this 1.x code: 

https://github.com/mongodb/js-bson/blob/39215f038b6787d015bdd1525e4b9db060603350/lib/bson/parser/serializer.js#L250-L258

The problem is that `instanceof MinKey` will only match the 1.x `MinKey` class. When you send a 4.x `MinKey` instance, it will be treated as MaxKey and presumably the query will return unexpected results.

To fix this I think we'd need to patch the 1.x library for this as well as patch `mongodb-core` and `mongodb`.

I'll build a PR. 

BTW I found this while writing tests for the PR for #306.  I'll comment out MinKey tests from that PR until this bug is fixed in 1.x.

justingrant:
BTW this issue was fixed by #313. Not sure why this issue wasn't auto-closed when that PR was merged, so closing it now.

PR #312, opened 2019-02-18
Improve 4.x<->1.x compatibility (includes revert _bsontype to ObjectID)

### LINKED ISSUE #306: Object IDs created by BSON v4 are silently ignored by Node driver 3.x
The latest Node driver 3.1.3 ignores all `ObjectId` instances created by the v4 BSON library, including Object IDs put in filters and Object IDs used as values to be stored.  The Object ID instance(s) will be silently ignored by the Node driver and will never be sent to the server.  No errors are generated and the server operation will complete successfully as if the `ObjectId` was not included in the input. 

This bug essentially makes the 4.x library incompatible with MongoDB, because any non-trivial MongoDB app will involve sending `ObjectId` instances to the server.  Using `ObjectId` inside a filter document is particularly dangerous. For example I found a case where, in a `findOneAndUpdate`, a 4.x `ObjectId` was in a filter for a `$pull` operation.  Instead of removing only one item from the array, MongoDB removed all of them!

Repro is trivial: 
* Create a document containing an BSON 4.x `ObjectId` instance, e.g. `{a: 1, b: new ObjectId()}`
* Try to insert that document using the Node 3.1.3 driver or use it as a query filter
* Expected: entire document is inserted / entire document is used a filter
* Actual: 
  * (insert case) only `{a: 1}` is inserted. `b` is not inserted, but the insert succeeds and no errors are returned. 
  * (filter case) all documents matching `{a: 1}` are included in the filter. The `ObjectId` filter is ignored.  This can lead to unexpected data loss for filters on delete or update operations.

The root cause is the switch of `_bsontype` from `ObjectID` to `ObjectId` (lowercase "d") in BSON v4.  When the Node driver 3.1.3 serializes commands before sending them to the server, the BSON v1 serializer used by Node driver 3.1.3 doesn't recognize `ObjectId` instances created by BSON V4. The serializer lacks a default "fall-through" case so`ObjectId` instances (and any other objects with an unrecognized `_bsontype`) are silently ignored by the serializer without any errors.  From the client's perspective, this just looks like data loss.

I can see a few possible fixes for this: 
* Option 1: go back to using `{_bsontype: 'ObjectID'}` like BSON v1 used.  Unless there's a compelling reason to leave it as `ObjectId`, this seems like the cleanest solution because it allows interop between BSON v4 and current Node drivers (or any other library that creates Object IDs that are duck-type compatible with BSON v1 ObjectID instances).
* Option 2: Patch the BSON 1.x branch to recognize `_bsontype: 'ObjectId'` too, and then patch the 3.x node driver to depend on the patched BSON 1.x. This is essentially the reverse of #290 which added `ObjectID` support to BSON 4.x.
* Option 3: Patch the 3.x Node driver to depend on BSON 4.x, and document the requirement for users to permanently avoid mixing BSON v4 with older versions of the Node driver that use BSON 1.x.  I assume that this would be the riskiest option given that there may be other potential bugs or breaking changes that are lurking in BSON v4 that haven't been found yet, so it might be too risky to bump this dependency without a minor or major release of the Node driver. 

@mbroadst - I'm happy to help with PRs for any and/or all of these options but could use your guidance. Do you have a preference for which option(s) should be implemented?  Option 1 seems like the least risky and impactful because there are few users of BSON v4 but many of BSON v1, but I don't understand the context behind the shift to `_bsontype: 'ObjectId'` so I may be wrong. Why was `_bsontype` changed in the first place?  I understand the logic of changing the class name to `ObjectId`, but why did `_bsontype` have to change too?  Is there any harm in changing it back?

The code below (from the BSON 1.x branch) is where the problem happens. If an object looks like `{_bsontype: 'ObjectId', ... }`, then execution falls through to the end of the chain of `if`/`else` statements without being serialized but also without generating any errors.  (BTW, the lack of a fall-through case for _bsontype is also a bug in 4.x. I'll file a new issue & PR to address this problem in 4.x.)

https://github.com/mongodb/js-bson/blob/39215f038b6787d015bdd1525e4b9db060603350/lib/bson/parser/serializer.js#L919-L982

### LINKED ISSUE #307: Invalid _bsontype objects silently fail serialization
As part of investigating #306, I noticed that objects with invalid `_bsontype` are incorrectly handled by both the BSON serializer and the EJSON serializer: 
* The BSON serializer will silently ignore these objects-- not throwing an error but not sending them to the server either.
* The EJSON serializer will serialize these objects as regular JS objects, which in turn can cause other problems downstream when the objects are deserialized. 

Silently failing is bad, but it's especially bad when the input data is completely lost like in the case of the BSON serializer. The right behavior in both cases should be to throw an error if an object with an invalid `_bsontype` is attempted to be serialized.

PR is coming soon!

justingrant:
This PR improves interop and compatibility between 1.x and 4.x versions of `js-bson`.  Goals:
1. Enable documents created by 4.x to be parsed/serialized/deserialized by 1.x BSON (e.g. the current Node driver) and vice-versa
2. Fix Object ID interop with the current Node driver and with existing duck-typing code that depends on `ObjectId._bsontype === 'ObjectID'`
3. Add wider test coverage for interop use-cases, and fix problems these tests uncover

_This is a fairly big PR so feel free give feedback if parts of this need work or are off-base. Thanks so much for reviewing!_

Changes include: 

**Tests**
* Added generic (every BSON type) interop tests between 1.x and 4.x libraries. Tests are 2x2 (Old/Old, Old/New, New/Old, New/New) and for both the BSON and EJSON serializers. The process behind this PR was to try every possible combination for every BSON type, and then fix the problems that were exposed.
* Adds `mongodb-core 3.1.11` as a DEV dependency in order to run tests against the 1.x library. Ideally we could just add `bson 1.1.0` as a DEV dependency instead, but npm prohibits self-referring packages.
* Changes `Decimal128` tests to undo global changes after these tests run, because those global changes were breaking EJSON tests.
* Adds a .vscode/launch.json file for running tests inside the VSCode debugger.  It took a while to get this debugging config working right, so wanted to help others trying to do the same thing.

**Fixes #306 - Object IDs created by BSON v4 are silently ignored by Node driver 3.x**
* Reverts the 4.x `_bsontype` change. After this PR, `ObjectId._bsontype` will be `ObjectID` again.  This will enable 4.x users to create `ObjectId`s and send them to the 3.x Node driver (today they're silently dropped).
* Includes support for 4.0.0 and 4.0.1 `_bsontype: 'ObjectId'` too.  Tests include mocks for the 4.0.1 `ObjectId` class to verify interop.
* Fixes typos in JSDoc comments in `objectid.js`
* Fixed a problem with `ObjectID` interop in the size calculator.
* Tests to validate ObjectID interop

**Fixes #307 - Invalid _bsontype objects silently fail serialization**
* Adds error handling that throws if unrecognized `_bsontype` values are passed to the EJSON serializer, which should prevent #306-like bugs in the future. Also adds tests to cover this behavior.
* Adds similar error handling for the BSON serializer + tests
* In the EJSON serializer, adds a whitelist of `_bsontype` values mapped to constructors or factories of 4.x BSON-typed objects. Every duck-typed or 1.x BSON object input to the EJSON serializer will get replaced with a 4.x equivalent, which replaces the special-case interop code for Object IDs that didn't work for other BSON types. 
* Ensures that BSON types are either serialized successfully or they'll throw.  No fall-through.

**Other interop fixes**
The generic interop tests above exposed a few type-specific bugs which are also fixed in this PR. In most cases, these fixes are covered by the same generic all-types tests noted above. In a few places (e.g. `MinKey`) there are new type-specific tests:
* Fixes half of `MinKey` serialization interop. This PR unblocks BSON-serializing a 1.x `MinKey` using the 4.x serializer. Unfortunately that's not nearly as useful as #310 which is the corresponding 1.x fix, because that one will enable 4.x clients to send `MinKey`s to the 3.1 Node driver.
* Enables `DBRef` interop via a new `DBRef.namespace` undocumented property that's an alias for the existing 4.x `collection` property (it was called `namespace` in 1.x)
* Fixes `DBRef` and `Code` serialization of embedded objects. As a side-effect of this fix, we no longer mutate the input to the EJSON serializer.
* Replaces `instanceof Long` with `Long.isLong()` to prevent 4.x/1.x interop problems with `Long` and `Timestamp`.


justingrant:
@mbroadst - Let me know if you have feedback about this PR. Thanks!

mbroadst:
Hey @justingrant, we're working on getting out a beta next week for some upcoming transactions work, so that will take priority right now unfortunately. I'm going to try to review this either this afternoon or Monday at the latest. Thanks for your patience!

.vscode/launch.json
```diff
@@ -0,0 +1,36 @@
+{
```
mbroadst: Can you please remove your vscode folder from the PR?

package.json
```diff
@@ -36,6 +36,7 @@
     "karma-mocha-reporter": "^2.2.5",
     "karma-rollup-preprocessor": "^6.1.0",
     "mocha": "^5.2.0",
+    "mongodb-core": "3.1.11",
```
mbroadst: I would prefer not to make `mongodb-core` a dependency of this project, in particular because the plan is to use this version of bson in that module eventually. I think a better approach might be for you to create an object that "looks" like the old `ObjectID`, or even just copy the basic parts of the old implementation into the tests.

lib/parser/calculate_size.js
```diff
@@ -94,7 +94,7 @@ function calculateElement(name, value, serializeFunctions, isArray, ignoreUndefi
         value['_bsontype'] === 'MaxKey'
       ) {
         return (name != null ? Buffer.byteLength(name, 'utf8') + 1 : 0) + 1;
-      } else if (value instanceof ObjectId || value['_bsontype'] === 'ObjectId') {
+      } else if (value instanceof ObjectId || value['_bsontype'] === 'ObjectId' || value['_bsontype'] === 'ObjectID') {
```
mbroadst: I think we should remove all `instanceof` checks for BSON types, and depend strictly on the `_bsontype` field.

lib/objectid.js
```diff
@@ -410,5 +410,5 @@ ObjectId.prototype.inspect = ObjectId.prototype.toString;
  */
 ObjectId.index = ~~(Math.random() * 0xffffff);
 
-Object.defineProperty(ObjectId.prototype, '_bsontype', { value: 'ObjectId' });
+Object.defineProperty(ObjectId.prototype, '_bsontype', { value: 'ObjectID' });
```
mbroadst: Can you please leave a note here for future readers as to why we have reverted this?

lib/extended_json.js
```diff
@@ -265,38 +265,68 @@ function serializeValue(value, options) {
...
-        doc.oid = serializeDocument(doc.oid, options);
+    return _doc;
+
+  } else if (typeof bsontype === 'string') {
+
+    // the "document" is really just a BSON type object
+    let _doc = doc;
+    if (typeof _doc.toExtendedJSON !== 'function') {
+      // There's no EJSON serialization function on the object. It's probably an
+      // object created by a previous version of this library or (another library
```
mbroadst: nit: in the comment the `or (another library` looks like the or should be within the parens

lib/extended_json.js
```diff
@@ -265,38 +265,68 @@ function serializeValue(value, options) {
   return value;
 }
 
+const bsonTypeMappings = {
```
mbroadst: nit: since this is a constant can we use caps (`BSON_TYPE_MAPPINGS`)

justingrant: OK done.

lib/extended_json.js
justingrant: fixed

lib/objectid.js
justingrant: Yep, done.

lib/parser/calculate_size.js
justingrant: Done.

package.json
justingrant: Yep, done. Note that the new tests look at interop for all BSON types (not just ObjectId), and also they test serialization code interop (4.x objects serialized by 1.x). So unless we want to remove those tests, the easiest solution here (what I just checked in) was to copy/paste the entire js-bson 1.x codebase into the tests folder. This will be easy to update if there are more changes to the 1.x branch-- simply install the new 1.x from npm and copy/paste.

.vscode/launch.json
justingrant: Yep, done. Getting the launch configs right was non-trivial, so once the dust settles I can PR a change to the docs with an example for other folks who want to run tests in VSCode.

justingrant:
OK I think I've addressed all review comments-- holler if you see anything else that needs fixing. Thx!

justingrant:
Hi @daprahamian - I'm sure you guys are super-busy, but is there an ETA for review on this PR?  Thanks!

justingrant:
Hi @daprahamian and @mbroadst - I'm happy to follow whatever you guys think is best-- lemme know what you want to do and I'll make changes.  That said, I don't think Option 3 is a great idea because the interop tests all go through the serializer and deserializer, and the deserializer code `require`s all BSON types... otherwise it couldn't deserialize all of them! This means that any Option 3 solution would require major surgery on the serializer and deserializer code, which itself could be a maintenance hassle if there are later patches to the 1.x serializer like #310.

It would probably be pretty easy, however, to reduce the 1.x code to just the `lib` folder which would limit code bloat to code that's actually used by tests and not include various 1.x config files that aren't needed for tests.

Another option would simply be to comment out the tests that use 1.x code and remove the 1x folder completely. Those tests were really valuable to run twice--once to find bugs and again to verify fixes--but probably won't be nearly as important going fwd, especially once 4.x replaces 1.x in `mongodb-core`.

Another possibility would be to try something similar to your option 2, but only install the 1.x code from npm when running tests. This could probably be done programmatically inside the test code itself (there are npm packages that make it easy to install npm packages), but the easiest way would probably be something like this in `package.json`:

`"test-node": "npm i --no-save --force bson@1.1.0 && mocha ./test/node",`
 or, with cleanup: 
`"test-node": "npm i --no-save --force bson@1.1.0 && mocha ./test/node && npm uninstall --no-save --force bson@1.1.0",`

I prototyped this and verified that `require('bson')` inside a unit test points to the right file, and also verified that tests pass with this setup.

Let me know which option you guys would like to do and I'll make it happen.

slimee:
Hi!

Thanks you for this work!

I love to build my event sourcing app with bson on amqp :)

justingrant:
OK @daprahamian @mbroadst - 1.x code is removed. Instead of commenting out interop tests, I realized that it's easy to dynamically skip them if `require('bson')` fails, which makes it easy to turn them on (or off) later without changing any test code. This is what's implemented in my last checkin. I verified that the tests work both ways. But I'm also happy to fully comment out these tests, just let me know.

Also let me know if there's anything else needed before this PR can be merged. Will be good to retire my fork! ;-)

I added this comment to the top of `extended_json_tests.js` to explain how to turn on the interop tests.

```
// Several tests in this file can test interop between current library versions and library version 1.1.0, because
// between 1.1.0 and 4.0.0 there was a significant rewrite. To minimize maintenance issues of a hard dependency on
// the old version, these interop tests are inactive by default. To activate, edit the test-node script in package.json:
//   "test-node": "npm i --no-save --force bson@1.1.0 && mocha ./test/node && npm uninstall --no-save --force bson@1.1.0"
```

justingrant:
Hi @mbroadst - I rebased and squashed commits into a single clean commit. Holler if you spot any problems.  

P.S. - This was my first large open-source PR anywhere, so it was cool to learn the process while helping with something hopefully useful.  Sorry if I made any newbie github mistakes. ;-) Thanks for the reviews and getting this over the finish line!

daprahamian:
Just merged this in @justingrant. Thank you so much for all the work you put in to this!

PR #313, opened 2019-02-27
Patch 1.x branch to fix 4.x->1.x interop for MinKey and ObjectId

### LINKED ISSUE #310: Library 4.x can't send MinKey to MongoDB

justingrant:
This PR for the 1.0 branch fixes #310 (which prevents 4.x from sending MinKey values to the current Node driver).  Unlike every other interop issue I found while building #312, `MinKey` didn't have an obvious workaround without changing the 1.x library (and bumping deps and re-releasing `mongodb-core` and `mongodb` packages).

While fixing the `MinKey` issue above, I also added interop support 4.0.0 and 4.0.1 `ObjectId._bsontype` so that either `ObjectID` or `ObjectId` values for `_bsontype` will be accepted by the 1.x library.  This doesn't remove the need to merge #312 (which contains fixes for duck-typed libraries and other BSON types) but does make things less broken for existing 4.0 users before they upgrade to a 4.x release that includes #312.

Tests for the fixes above are included. Testing the MinKey changes was easy. For testing ObjectId I copied the 4.x ObjectId implementation into the Tests folder and converted it to pre-ES6 code so that the Node 0.12 tests would run.

### DIFF
--- a/browser_build/bson.js
+++ b/browser_build/bson.js
@@ -16447,11 +16447,10 @@ return /******/ (function(modules) { // webpackBootstrap
 
 	var writeIEEE754 = __webpack_require__(353).writeIEEE754,
 	    Long = __webpack_require__(334).Long,
-	    MinKey = __webpack_require__(347).MinKey,
 	    Map = __webpack_require__(333),
 	    Binary = __webpack_require__(350).Binary;
 
-	const normalizedFunctionString = __webpack_require__(354).normalizedFunctionString;
+	var normalizedFunctionString = __webpack_require__(354).normalizedFunctionString;
 
 	// try {
 	//   var _Buffer = Uint8Array;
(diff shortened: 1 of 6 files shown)


mbroadst:
Thanks for your patience @justingrant. I'm just back from vacation, and we are now discussing releases for these fixes.

PR #317, opened 2019-03-07
Don't use new Buffer

daprahamian:


lib/bson/bson.js
```diff
@@ -56,7 +59,7 @@ BSON.prototype.serialize = function serialize(object, options) {
   
   // Resize the internal serialization buffer if needed
   if (buffer.length < minInternalBufferSize) {
-    buffer = new Buffer(minInternalBufferSize);
+    buffer = util.allocBuffer(minInternalBufferSize);
```
kvwalker: `util` => `uitls`

lib/bson/bson.js
```diff
@@ -28,10 +29,12 @@ var deserialize = require('./parser/deserializer'),
 var MAXSIZE = 1024 * 1024 * 17;
 
 // Current Internal Temporary Serialization Buffer
-var buffer = new Buffer(MAXSIZE);
+var buffer = utils.allocBuffer(MAXSIZE);
 
 var BSON = function() {};
 
+var utils = require('./parser/utils');
```
kvwalker: Is this being required twice?

daprahamian: it is! lemme fix that

PR #322, opened 2019-06-04
fix(Decimal128): throw error when passing a string to Decimal128 constructor

vkarpov15:
You can get some nasty surprises when you pass a string to the Decimal128 constructor:

```javascript
const bson = require('bson');

const v = new bson.Decimal128('14.92');

// {"$numberDecimal":"8.740930561E-6167"}
console.log(JSON.stringify(v));
```

Would be handy to at least throw an error. Although is there some reason why we can't just do `fromString()` if the user passes a string instead of a buffer to the `Decimal128` constructor?

### DIFF
--- a/lib/decimal128.js
+++ b/lib/decimal128.js
@@ -154,6 +154,9 @@ function invalidErr(string, message) {
  * @return {Double}
  */
 function Decimal128(bytes) {
+  if (typeof bytes === 'string') {
+    throw new Error('Decimal128 constructor expects a Buffer parameter. Use `Decimal128.fromString()` to load a Decimal128 from a string'); 
+  }
   this.bytes = bytes;
 }
 


vkarpov15:
Re: https://github.com/Automattic/mongoose/issues/7815

mbroadst:
I can't think of a reason we wouldn't support a string in the constructor, other than nobody thought to do it when they were originally implementing the module. How about adding that here with a test?

daprahamian:
I'm on board with 

```js
function Decimal128(bytes) {
  if (typeof bytes === 'string') {
    return Decimal128.fromString(bytes);
  }
  this.bytes = bytes;
}
```

agolin95:
https://jira.mongodb.org/browse/NODE-3018

bajanam:

PR #325, opened 2019-07-03
feat(extended_json): EJSON.to/fromBuffer

spal1:
NODE-2008

Two methods are being added to `bson.EJSON`: `toBuffer(str)` and `fromBuffer(buffer)`

### DIFF
--- a/lib/extended_json.js
+++ b/lib/extended_json.js
@@ -327,9 +327,21 @@ function serializeDocument(doc, options) {
   }
 }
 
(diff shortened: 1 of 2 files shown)


jimmywarting:
Hmm, this did go a tiny bit about my recommendation about stop depending on buffer and making bson more cross friendlier for both deno and browsers.
if anything then i think you should have to/from uint8array or ArrayBuffer

bajanam:

PR #328, opened 2019-10-01
docs: ObjectId#isValid has a parameter

arosequist:


lib/objectid.js
```diff
@@ -315,6 +315,7 @@ class ObjectId {
    * Checks if a value is a valid bson ObjectId
    *
    * @method
+   * @param {object} id ObjectId instance to validate.
```
daprahamian: Since it is valid to pass any type in to the function, could you change the type field from `object` to `*`?

arosequist: Sure thing, thanks!

PR #333, opened 2019-10-22
fix(_bsontype): only check bsontype if it is a prototype member.

daprahamian:


lib/bson/parser/serializer.js
```diff
@@ -898,6 +903,9 @@ var serializeInto = function serializeInto(
       // Check the type of the value
       type = typeof value;
 
+      var _objPrototype;
```
mbroadst: these two lines look a little funky to me, why are you doing object assignment as part of a long conditional chain? It also looks like `_bsontype` could end up being some falsey value like false/undefined/null. Let's make the check for `_bsontype` on the prototype more explicit?

You already grab the type of `value` above, so you can confirm if this thing is `null` or `undefined` without an extra check, and at that point we should be able to check the prototype. Also, it doesn't seem like we need to store the prototype, I can't see it being used below for anything.

daprahamian: if `_bsontype` is a falsey value, then it will never `===` a string that we check against, so that is fine.

The fully expanded version of the check is:

```js
var _objPrototype = type === 'object' && value && Object.getPrototypeOf(value);
var _bsontype =  _objPrototype && objPrototype._bsonvalue;
```

The reason for these checks are to handle the following cases:
1. any non-object value will throw for `Object.getPrototypeOf(value)`, so we do `type === 'object'`
2. `null` will pass `type === 'object'`, and will throw for `Object.getPrototypeOf(value)`, so we also do a truthy check on `value`
3. `Object.create(null)` will create an object with a null prototype, so we need to make sure that `_objPrototype` is truthy before picking the value off of it.

I condensed this into one line b/c I was being hacky. I can expand it back into two lines.

mbroadst: OK cool, yeah I would prefer two lines. Also, can you please include tests for all these cases? Let's prove your assumptions.

daprahamian:
Updated with tests, splitting the check to two lines, and also throwing if an unknown `_bsontype` is somehow detected.

PR #336, opened 2019-11-14
fix: throw if invalid _bsontype is detected

daprahamian:


mbroadst:
@daprahamian it looks like there is a real failure here on the 0.12x Travis CI run

PR #338, opened 2019-11-29
src(objectid): cleanup and simplification, perf improvement

zbjornson:
I have another PR prepared that trivially speeds up the `equals` method a ton, but wanted to clean the surrounding code up first.

---

* Since Node.js v6.9.0 is the minimum version, `Buffer.from` is always available. Allows removing `decodeLookup` stuff, poor duck-typing of 'id' (`this.id.copy` was used as a test for it being a buffer).

* Indicate that 12-byte string is allowed as constructor arg.

* Simplify logic in constructor. Performance improvement from not having to call `isValid` in as many cases and reordering of branches. **Creating an ObjectId from a hex string is now ~19% faster.**

* Remove a hidden Buffer constructor usage (aliased as `new _Buffer`)

* Add a TODO regarding an existing bug in `getTimestamp`. (I want to convert 12-char strings to Buffers in the ctor to fix this and clean other code up, but that would be a breaking change because it would make the constructor throw instead of `toHexString`. Throwing in the ctor seems preferable...)

* Simplify `createFromHexString`. (This is the same as the constructor but requires the input be a hex string.)

* Simplify logic (remove redundant and unnecessary code) in `equals` method.

### DIFF
--- a/lib/objectid.js
+++ b/lib/objectid.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const Buffer = require('buffer').Buffer;
-let randomBytes = require('./parser/utils').randomBytes;
+const randomBytes = require('./parser/utils').randomBytes;
 const util = require('util');
 const deprecate = util.deprecate;
 
@@ -10,30 +10,11 @@ const PROCESS_UNIQUE = randomBytes(5);
 
 // Regular expression that checks for hex value
 const checkForHexRegExp = new RegExp('^[0-9a-fA-F]{24}$');
-let hasBufferType = false;
-
-// Check if buffer exists
-try {
-  if (Buffer && Buffer.from) hasBufferType = true;
-} catch (err) {
-  hasBufferType = false;
-}
 
 // Precomputed hex table enables speedy hex string conversion
 const hexTable = [];
 for (let i = 0; i < 256; i++) {
   hexTable[i] = (i <= 15 ? '0' : '') + i.toString(16);
-}
-
-// Lookup tables
-const decodeLookup = [];
-let i = 0;
-while (i < 10) decodeLookup[0x30 + i] = i++;
-while (i < 16) decodeLookup[0x41 - 10 + i] = decodeLookup[0x61 - 10 + i] = i++;
-
-const _Buffer = Buffer;
-function convertToHex(bytes) {
-  return bytes.toString('hex');
 }
 
 function makeObjectIdError(invalidString, index) {
@@ -52,49 +33,40 @@ class ObjectId {
   /**
    * Create an ObjectId type
    *
-   * @param {(string|Buffer|number)} id Can be a 24 byte hex string, 12 byte binary Buffer, or a Number.
+   * @param {(string|Buffer|number)} id Can be a 24 character hex string, 12 byte string, 12 byte Buffer, or a Number.
    * @property {number} generationTime The generation time of this ObjectId instance
    * @return {ObjectId} instance of ObjectId.
    */
   constructor(id) {
-    // Duck-typing to support ObjectId from different npm packages
     if (id instanceof ObjectId) return id;
 
-    // The most common usecase (blank id, new objectId instance)
     if (id == null || typeof id === 'number') {
-      // Generate a new id
       this.id = ObjectId.generate(id);
-      // If we are caching the hex string
-      if (ObjectId.cacheHexString) this.__id = this.toString('hex');
-      // Return the object
-      return;
-    }
-
-    // Check if the passed in id is valid
-    const valid = ObjectId.isValid(id);
-
-    // Throw an error if it's not a valid setup
-    if (!valid && id != null) {
-      throw new TypeError(
-        'Argument passed in must be a single String of 12 bytes or a string of 24 hex characters'
-      );
-    } else if (valid && typeof id === 'string' && id.length === 24 && hasBufferType) {
-      return new ObjectId(Buffer.from(id, 'hex'));
-    } else if (valid && typeof id === 'string' && id.length === 24) {
-      return ObjectId.createFromHexString(id);
-    } else if (id != null && id.length === 12) {
-      // assume 12 byte string
+    } else if (typeof id === 'string') {
+      if (id.length === 24 && checkForHexRegExp.test(id)) {
+        this.id = Buffer.from(id, 'hex');
+        this.__id = id;
+      } else if (id.length === 12) {
+        // TODO convert to buffer now? Would allow cleaning up 'equals' method
+        // and fix the bug in 'getTimestamp'.
+        this.id = id;
+      }
+    } else if (id instanceof Buffer && id.length === 12) {
       this.id = id;
-    } else if (id != null && id.toHexString) {
+    } else if (id.toHexString) {
       // Duck-typing to support ObjectId from different npm packages
-      return ObjectId.createFromHexString(id.toHexString());
-    } else {
+      this.id = Buffer.from(id.toHexString(), 'hex');
+    }
+
+    if (!this.id) {
       throw new TypeError(
         'Argument passed in must be a single String of 12 bytes or a string of 24 hex characters'
       );
     }
 
-    if (ObjectId.cacheHexString) this.__id = this.toString('hex');
+    if (ObjectId.cacheHexString && !this.__id) {
+      this.__id = this.toString('hex');
+    }
   }
 
   /**
@@ -106,7 +78,6 @@ class ObjectId {
   toHexString() {
     if (ObjectId.cacheHexString && this.__id) return this.__id;
 
-    let hexString = '';
     if (!this.id || !this.id.length) {
       throw new TypeError(
         'invalid ObjectId, ObjectId.id must be either a string or a Buffer, but is [' +
@@ -115,12 +86,13 @@ class ObjectId {
       );
     }
 
-    if (this.id instanceof _Buffer) {
-      hexString = convertToHex(this.id);
+    if (this.id instanceof Buffer) {
+      const hexString = this.id.toString('hex');
       if (ObjectId.cacheHexString) this.__id = hexString;
       return hexString;
     }
 
+    let hexString = '';
     for (let i = 0; i < this.id.length; i++) {
       const hexChar = hexTable[this.id.charCodeAt(i)];
       if (typeof hexChar !== 'string') {
@@ -189,7 +161,7 @@ class ObjectId {
    */
   toString(format) {
     // Is the id a buffer then use the buffer toString method to return the format
-    if (this.id && this.id.copy) {
+    if (this.id && this.id instanceof Buffer) {
       return this.id.toString(typeof format === 'string' ? format : 'hex');
     }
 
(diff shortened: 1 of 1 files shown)


zbjornson:
Any chance this can be reviewed please? The 19% performance improvement is significant, and this removes a lot of dead code.

nbbeeken:
Hi 👋  I created this ticket to track the issue, we'll be reviewing it
https://jira.mongodb.org/browse/NODE-2492

marcj:
Is this project dead, or why does it take so long to review those PRs? I rewrote the whole JS BSON parser for my project and got a 3x speed-up, and for certain cases 7.5x speep-up (which makes it faster than JSON.parse) and I thought about providing a pull-request, however it seems a lot of PRs are not even reviewed and progress stalled, so I wonder if it's even worth it. @nbbeeken @mbroadst

mbroadst:
Hi @marcj, no the project is most definitely not dead, but it is used by millions of projects every day so we try to take a lot of care in making changes given past issues with subtle breaks that are highly disruptive to our users. We're currently in the process of [rewriting the module in TypeScript](https://jira.mongodb.org/browse/NODE-2647) in order to provide higher quality type information, as well as focusing on web users of the library as first-class citizens. We're aware there is room for improvement in performance, but that will be our focus after ensuring stability. 

Regarding the performance work you've already done, a pull request would be welcome but please be aware that our focus will be on the above project in the short term.

marcj:
@mbroadst thanks for your quick response. Cool to hear it's not dead! :) 
Yeah, so I wrote it in Typescript as well. Your linked Jira issue can't be viewed as anonymous user so I don't see the progress of it. If you want I can take this task over (in rewriting the lib itself as Typescript), should be done in a few days only. I'd integrate then the performance already and leave all js tests as is, adding new TS tests on top of it to make sure now BC in public API is introduced. If you tell me this will realistically merged so everybody can enjoy better typings and at least a 3x perf improvement I'd start right away.

mbroadst:
@marcj I really appreciate the offer! We've internally scoped and designed the project, and it's already been started so I think it's better to continue to run it as we expected to. Once the TypeScript changes have landed, I imagine we could apply your performance changes on top of it. I updated the link above (sorry I forgot epics aren't public, stupid jira nuance) to link to the individual ticket for converting the source to TypeScript, so you can watch that or I can notify you on there if you make a comment.

marcj:
@mbroadst fair enough, I don't want to take away work scheduled for @nbbeeken :) 
Ok, then I'll keep my parser in my repo and link it when published so you can take it as template/reference. For me it's not important if it is in the official js-bson package or not, because I'll use my JIT BSON parser/serializer anyways, but I also implemented a schemaless version which yielded also 3x improvements that can be easily integrated into this package. It's honestly a game-changer for me because now it's actually faster and lighter to use BSON for server<->client communication, whereas we were stuck for years with JSON. So I think it's for other people also a game changer and should be published ASAP - preferably as new major version, to get rid of all the legacy stuff.

mbroadst:
@marcj are you talking about work outside of @zbjornson s work in `bson-ext` (https://github.com/zbjornson/bson-to-json)? Is your JIT code posted somewhere I can take a look at?

marcj:
@mbroadst I don't talk about `bson-ext` or `bson-to-json`. I also rewrote the parser in C++, but fundamentally the public v8 API itself limits the performance and I wasn't able to get anywhere near the JS implementation with moderate investments (my implementation actually was in the same ballpark of the `bson-ext` implementation). In my c++ implementation I also tried like the official JSON.parse implementation to use custom stack management so it doesn't recursively create function call stacks, but didn't help in BSON's use-cases. The reason is that BSON is heavily utf8 string parsing dependent (object keys) and the public v8 API doesn't allow you to easily reuse already allocated strings in the heap, so what both JS-bson and bson-ext do is to create ten thousands new strings on the heap, although they are all the same. The official c++ v8 JSON.parse implementation is so fast because they work around that and use the internal heap factory, which partially re-uses generated strings - like the v8 JIT does with string literals. This makes a huge difference when working with many objects in an array (classic db scenario) or objects in general. In the Marshal JS JIT version (see below), I exploited the fact that the v8 JIT reuses string literals and thus you save a lot of memory allocations, also the comparison of object keys is done by look-ahead using the ArrayBuffer directly instead of first converting to costly utf8 strings.

Here are some benchmark outputs:

```
Start benchmark BSON Parser array with 10,000 objects
 🏎 x 494.85 ops/sec ±1.35% 0.0020208309195804 sec/op 	▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ Marshal JS JIT
 🏎 x 156.40 ops/sec ±2.18% 0.0063939557061404 sec/op 	▇▇▇▇▇▇▇▇▇▇▇▇▇▆▇▇▇▇▇ Marshal JS generic
 🏎 x  38.47 ops/sec ±0.77% 0.0259920129411765 sec/op 	▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ official js-bson parser
 🏎 x  31.68 ops/sec ±2.53% 0.0315616399666667 sec/op 	▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ official bson-ext, c++ native
 🏎 x 198.80 ops/sec ±1.62% 0.0050302004969697 sec/op 	▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ JSON.parse()
 🏁 Fastest Marshal JS JIT

```

```
Start benchmark BSON Parser array with 1 objects
 🏎 x 2,324,617.95 ops/sec ±4.56% 0.0000004301782152 sec/op 	▇▇▇▇▇▇▆▇▇▆▆▆▇▇▇▆▆▆ Marshal JS JIT
 🏎 x   838,576.54 ops/sec ±2.00% 0.0000011924969898 sec/op 	▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ Marshal JS generic
 🏎 x   284,910.04 ops/sec ±1.95% 0.0000035098797714 sec/op 	▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ official js-bson parser
 🏎 x   188,435.56 ops/sec ±0.99% 0.0000053068540811 sec/op 	▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ official bson-ext, c++ native
 🏎 x   769,308.96 ops/sec ±1.34% 0.0000012998678769 sec/op 	▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ JSON.parse()
```

`Object` in question:

```
{
        id: i,
        name: 'Peter',
        ready: true,
        priority: 0,
        tags: ['a', 'b', 'c'],
}
```

Source code is not yet committed, but I will today or tomorrow, then I'll link it here.

marcj:
Please note that performance degrades massively when dealing with bigger strings (as Buffer.toString is then required which is costly).

Another thought about optimizing it further: We could improve the JS non-JIT variant by implementing speculative property key parsing. This is also used in v8 JSON.parse and allows to not to always (costly) create and convert the same utf8 bytes. I haven't implemented it yet tho, but correctly done it should bring the op/s near to the JIT parser making it faster than JSON.parse without the need to write a fix schema definition (that you would need with the JIT parser).

marcj:
> We could improve the JS non-JIT variant by implementing speculative property key parsing

I just did that and now the generic (non-JIT) parser is faster than JSON.parse.

```
Start benchmark BSON Parser array with 10000 objects
 🏎 x 467.74 ops/sec ±1.69% 0.0021379485127273 sec/op 	▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ Marshal JS JIT
 🏎 x 147.54 ops/sec ±1.88% 0.0067779742625 sec/op 	▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ Marshal JS generic
 🏎 x 246.57 ops/sec ±0.73% 0.0040556664395604 sec/op 	▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ Marshal JS generic speculative property key
 🏎 x  35.27 ops/sec ±2.44% 0.0283524985 sec/op 	▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ official js-bson parser
 🏎 x  32.10 ops/sec ±1.91% 0.0311545368 sec/op 	▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ official bson-ext, c++ native
 🏎 x 195.37 ops/sec ±2.02% 0.0051183712272727 sec/op 	▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ JSON.parse()
```

I implemented a very basic not per-document but per buffer property cache which increased performance greatly. This approach shines particularly when object keys are relatively constant. If they are completely random (which is rarely the case) it's a bit slower than the current generic parser. This approach can also easily be adopted by js-bson.

marcj:
I just published the code for both the parser and serializer. Let me know if you need anything or have any questions. Please note that they support only a subset of BSON, as I don't need certain types like SYMBOL, CODE_W_SCOPE, TIMESTAMP, etc, but adding those into the parse function statement won't hurt performance. You can also ping me once you are done rewriting js-bson to TS, then I can adjust the parser to reflect the perf adjustments if it wasn't possible to extract it from my code.

You find the [generic parser here](https://github.com/super-hornet/super-hornet.ts/blob/master/packages/marshal-bson/src/bson-parser.ts) and the [JIT version here](https://github.com/super-hornet/super-hornet.ts/blob/master/packages/marshal-bson/src/bson-jit-parser.ts): 

In the [README](https://github.com/super-hornet/super-hornet.ts/tree/master/packages/marshal-bson) you find more information.

The parser benchmark is located [here](https://github.com/super-hornet/super-hornet.ts/blob/master/packages/marshal-bson/benchmarks/bench-bson-parser.ts) and serializer benchmark is located [here](https://github.com/super-hornet/super-hornet.ts/blob/master/packages/marshal-bson/benchmarks/bench-bson-serializer.ts).

billouboq:
Instead of updating js-bson(which could possibly break in userland) would it be possible to plug a different parser in mongodb native driver ?

marcj:
@billouboq this was possible in `mongodb-core`, but isn't anymore in `node-mongodb-native`. Unfortunately I have to say, because I wanted to use my serializer also for commands send to the mongodb server. Currently it's hard coded in `node-mongodb-native`, see https://github.com/mongodb/node-mongodb-native/blob/master/src/bson.ts. This file is imported directly everywhere.

zbjornson:
@billouboq if you use yarn instead of npm, you can (probably) use dependency resolutions to override the package used for "bson": https://classic.yarnpkg.com/en/docs/selective-version-resolutions/. I haven't tried it with this package to know if it works for sure though.

Uzlopak:
What is speculative property key parsing?

Uzlopak:
And how did you implement it? There is no commit in this PR?

marcj:
@zbjornson

> I don't think can actually change package names, just versions.

No worries, I meanwhile [reimplemented](https://github.com/deepkit/deepkit-framework/tree/master/packages/mongo/src/client) the whole mongodb client a while ago but not only to use my own bson serializer and get an overall performance boost (up to 3x), but also to fix many shortcomings that are in my opinion not acceptable in 2020. Performance, missing BigInt support, and stack trace loss just to name a few.

Uzlopak:
@marcj
Do you have way to use your client under the hood of mongoose?

marcj:
@Uzlopak no, not possible since the API is different. However, it also includes an ORM that is much faster than Mongoose with more modern schema definition API and features. It's actually a mystery to me that people still use Mongoose.

agolin95:
Tracked in Jira: https://jira.mongodb.org/browse/NODE-3024

bajanam:

PR #342, opened 2019-12-28
fix: Support Number object in Int32 and Double constructors

stieg:
Numbers objects deserve the same love and support as their primitive
counterparts. So handle them appropriately without introducing
regressions into behaviors of other types that may be passed in as
values

Fixes NODE-2384


mbroadst:
Thanks for the contribution @stieg !

Issue #331, opened 2019-10-17
Incorrect binary data EJSON.stringify


JeromeDeLeon:
Basically I wanted to stringify the binary data but doing so results in an invalid one.

Suppose EJSON is imported;

```js
function getEJSONString(data) {
  return EJSON.stringify(data);
}
```

That gives me wrong binary format and whenever I imported it in another database of MongoDB, the files is nowhere to be found.

I manage to do it using old library, but tried several things using the current library like using `serialize`, `deserialize` or combination of both, but still the the problem persists.

Using `mongodb-extended-json` library solves the problem.

```js
function getJSONString(data) {
  return EJSON.stringify(ejson.serialize(data));
}
```

Still this is an outdated library and it would be better if this can be solve using the current one.

Where did I go wrong here?

daprahamian:
@JeromeDeLeon can you give an example of what you mean by "wrong binary format"?

JeromeDeLeon:
I tried to export the MongoDB data manually with NodeJS.
I was using EJSON stringify only to stringify all documents from collections; however, I was getting binary string differs from the binary string in the database.

from the database:
`/9j/4AAQSkZJRgABAgAAZABkAAD/`...

What I got:
`w7/CxAMDBAXDw0PFxsUEBAUGxW`....

But somehow, I managed to successfully stringify it, using the following:

```js
function getJSONString(data) {
  const { EJSON } = BSON;

  const serializedData = BSON.serialize(data);
  const deSerializedData = BSON.deserialize(serializedData);
  const serializedEJSON = EJSON.serialize(deSerializedData);

  return EJSON.stringify(serializedEJSON);
}
```

Am I doing it wrong?

JeromeDeLeon:
The setup is here:

I'm using a library that encapsulates handling mongodb.

MongoDB: 3.3.1
OS: Windows 10
bson: 4.0.2

I just connect to the mongdb and fetch that data and stringify it using EJSON.
The result is the same as above.

JeromeDeLeon:
I just don't get it why I need to do the following:
BSON S > BSON DS > EJSON S > EJSON Stringify just to get the right result

daprahamian:
@JeromeDeLeon Thank you for the system information. Everything I have done on my end does not surface this bug, so I need to see exactly what you are doing to figure out where the discrepancy is. Please provide a code snippet I can run so I can reproduce what you are seeing.

JeromeDeLeon:
```js
[...]

PR #348, opened 2020-01-14
fix: incorrect binary data

kambing86:
fix the issue https://github.com/mongodb/js-bson/issues/331


zandaqo:
I guess this is the issue that gives an error here:
```javascript
const binary = new BSON.Binary([1,2,3]);
const bson = BSON.serialize({ a: binary });
//=> TypeError: data.copy is not a function
```

bajanam:

PR #350, opened 2020-03-25
NODE-2518: preserve type information when round tripping symbols

mbroadst:
A decision made early in the refactoring of this module for v2 (now v4) lead us to always upgrade `BSONSymbol` to a JavaScript `String`. This is problematic because the type can no longer be round tripped to and from raw BSON bytes. This changeset puts the automatic upgrading behind an existing flag, `promoteValues`, which is already used in a number of places to do this type of lossy conversion.

The last commit is the real change, but a number of other changes were made along the way to fix the bson corpus test runner.

### DIFF
--- a/lib/double.js
+++ b/lib/double.js
@@ -1,4 +1,5 @@
 'use strict';
+
 /**
  * A class representation of the BSON Double type.
  */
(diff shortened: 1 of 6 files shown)


lib/double.js
```diff
@@ -41,16 +42,22 @@ class Double {
     if (options && (options.legacy || (options.relaxed && isFinite(this.value)))) {
       return this.value;
     }
-    return { $numberDouble: this.value.toString() };
+
+    if (Object.is(Math.sign(this.value), -0)) {
+      return { $numberDouble: `-${this.value.toFixed(1)}` };
```
emadum: TIL there's difference between -0 and +0. I think a comment here could be useful to explain this check, it seems pretty obscure.

mbroadst: _definitely_ obscure! Okay, I added a comment

lib/parser/deserializer.js
```diff
@@ -414,8 +415,8 @@ function deserializeObject(buffer, index, options, isArray) {
         buffer[index + stringSize - 1] !== 0
       )
         throw new Error('bad string length in bson');
-      // symbol is deprecated - upgrade to string.
-      object[name] = buffer.toString('utf8', index, index + stringSize - 1);
+      const symbol = buffer.toString('utf8', index, index + stringSize - 1);
+      object[name] = promoteValues ? symbol : new BSONSymbol(symbol);
```
nbbeeken: nit: cant we benefit from short circuiting here and not have to do toString on the buffer if we might just toss it?

mbroadst: hm, how would we short circuit here? At this point we are definitely going to use the string, it's just a matter of whether we return just the string or wrap it in a `BSONSymbol` instance.

nbbeeken: I'm blind 🤦‍♂ nvm LGTM

PR #355, opened 2020-04-16
fix(object-id): make sure new instance of ObjectID returns an ObjectID

davidfou:
No credit, the code comes from master:
https://github.com/mongodb/js-bson/blob/c307ca8e81b73ccfcac0b54e5012123773d30ddb/lib/objectid.js#L90

I'm not sure if that's the way to contribute to this repo :sweat_smile:

### DIFF
--- a/lib/bson/objectid.js
+++ b/lib/bson/objectid.js
@@ -68,7 +68,7 @@ var ObjectID = function ObjectID(id) {
     this.id = id;
   } else if (id != null && id.toHexString) {
     // Duck-typing to support ObjectId from different npm packages
-    return id;
+    return ObjectID.createFromHexString(id.toHexString());
   } else {
     throw new Error(
       'Argument passed in must be a single String of 12 bytes or a string of 24 hex characters'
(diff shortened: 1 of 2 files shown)


davidfou:
Can someone retrigger a build? For an unknown reason, `npm install` failed on node@6 https://travis-ci.org/github/mongodb/js-bson/jobs/675839512

Uzlopak:
This code seems wrong. As the comment on this line says, that the object gets ducktyped and returned, as it is then expected to be valid

davidfou:
@Uzlopak do you suggest changing the code on master? The code I wrote comes from there.

It sounds weird to me to have a case where `new ObjectId()` doesn't return an instance of `ObjectId`.

In my use case, the bug comes from much more `mongoose@5.0.17`. On this [notebook](https://runkit.com/dfournier/fix-bson-lib/1.0.0), the problem is reproduced. Updating to `5.0.18` will avoid falling to this error given this [change](https://github.com/Automattic/mongoose/compare/5.0.17...5.0.18#diff-e1bbd4f15e3b63427b4261e05b948ea8R12).

mbroadst:
This has been fixed in #382

PR #359, opened 2020-05-14
[NODE-2613] fix: parseInt value of Int32 in constructor

reggi:
This update adds `parseInt` to the `Int32` constructor. 

Pre-Fix:

> [Here's a repl.it page with examples.](https://repl.it/@thomasreggi/NODE-2613-without-fix)

```js
> bson.EJSON.parse(`{"$numberInt":"1"}`, { relaxed: false });
Int32 { value: '1' }
> new bson.Int32('1')
Int32 { value: '1' }
```

Post-Fix

> [Here's a repl.it page with the changes from this branch introduced.](https://repl.it/@thomasreggi/NODE-2613-with-fix)

```js
> bson.EJSON.parse(`{"$numberInt":"1"}`, { relaxed: false });
Int32 { value: 1 }
> new bson.Int32('1')
Int32 { value: 1 }
```

[NODE-2613](https://jira.mongodb.org/browse/NODE-2613)

### DIFF
--- a/lib/int_32.js
+++ b/lib/int_32.js
@@ -14,11 +14,7 @@ class Int32 {
       value = value.valueOf();
     }
 
-    if (typeof value === 'string') {
-      value = parseInt(value, 10);
-    }
-
-    this.value = value;
+    this.value = +value;
   }
 
   /**
--- a/lib/int_32.js
+++ b/lib/int_32.js
@@ -6,7 +6,7 @@ class Int32 {
   /**
    * Create an Int32 type
    *
-   * @param {number|Number|string} value the number we want to represent as an int32.
+   * @param {*} value the number we want to represent as an int32.
    * @return {Int32}
    */
   constructor(value) {
--- a/lib/int_32.js
+++ b/lib/int_32.js
@@ -6,12 +6,16 @@ class Int32 {
   /**
    * Create an Int32 type
    *
-   * @param {number|Number} value the number we want to represent as an int32.
+   * @param {number|Number|string} value the number we want to represent as an int32.
    * @return {Int32}
    */
   constructor(value) {
     if (value instanceof Number) {
       value = value.valueOf();
+    }
+
+    if (typeof value === 'string') {
+      value = parseInt(value, 10)
     }
 
     this.value = value;
--- a/lib/int_32.js
+++ b/lib/int_32.js
@@ -15,7 +15,7 @@ class Int32 {
     }
 
     if (typeof value === 'string') {
-      value = parseInt(value, 10)
+      value = parseInt(value, 10);
     }
 
     this.value = value;
--- a/test/node/int_32_tests.js
+++ b/test/node/int_32_tests.js
@@ -6,9 +6,9 @@ const expect = require('chai').expect;
 
 describe('Int32', function() {
   describe('Constructor', function() {
-    var hexValue = '0x2A';
-    var octalValue = 0o52;
-    var value = 42;
+    const hexValue = '0x2A';
+    const octalValue = 0o52;
+    const value = 42;
 
     it('Primitive number', function(done) {
       expect(new Int32(value).valueOf()).to.equal(value);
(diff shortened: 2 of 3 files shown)


test/node/int_32_tests.js
```diff
@@ -17,5 +19,49 @@ describe('Int32', function() {
...
+      expect(new Int32(hexValue).valueOf()).to.equal(value);
+      done();
+    });
+
+    it('Octal', function(done) {
+      expect(new Int32(octalValue).valueOf()).to.equal(value);
+      done();
+    });
+
+    it('Cases covering Int32 value conversion using +', function(done) {
```
reggi, line 33: @nbbeeken I've added all the tests we discussed here, let me know if I forgot any, ran this before and after PR and all work.

reggi:
@mbroadst @emadum Per @nbbeeken's recco changed this to use + over `parseInt` which has some limitations. Requesting re-review.

test/node/int_32_tests.js
nbbeeken: Coverage here looks good 👍 🚀

test/node/int_32_tests.js
```diff
@@ -17,5 +19,53 @@ describe('Int32', function() {
       expect(new Int32(new Number(value)).valueOf()).to.equal(value);
       done();
     });
+
+    it('Hex', function(done) {
```
mbroadst, line 23: nit: while we're here why not switch to using `context`?

test/node/int_32_tests.js
```diff
@@ -17,5 +19,53 @@ describe('Int32', function() {
...
+    });
+
+    it('Octal', function(done) {
+      expect(new Int32(octalValue).valueOf()).to.equal(value);
+      done();
+    });
+
+    it('Cases covering Int32 value conversion using +', function(done) {
+      const string = BSON.serialize({ key: new Int32('fortyTwo') });
+      const plus = BSON.serialize({ key: new Int32(+'fortyTwo') });
```
mbroadst, line 35: nit: looks like you could dramatically decrease the code dupe here:
```
[ '42fortyTwo', hexValue, +octalValue, 42, ...].forEach(value => {
  const serialized = BSON.serialize({ key: new Int32(value) });
  expect(serialized).to.equal(zero);
});
```

test/node/int_32_tests.js
```diff
@@ -17,5 +19,53 @@ describe('Int32', function() {
...
+      expect(plusOctal.equals(fortyTwo)).to.be.true;
+      expect(plusZero.equals(zero)).to.be.true;
+      expect(positiveZero.equals(zero)).to.be.true;
+      expect(negativeZero.equals(zero)).to.be.true;
+      expect(inf.equals(zero)).to.be.true;
+      expect(strInf.equals(zero)).to.be.true;
+      expect(plusStrInf.equals(zero)).to.be.true;
+      expect(posInf.equals(zero)).to.be.true;
+      expect(negInf.equals(zero)).to.be.true;
+      done();
```
mbroadst, line 68: no need for a call to `done` in a test case which is not async

lib/int_32.js
```diff
@@ -6,15 +6,15 @@ class Int32 {
   /**
    * Create an Int32 type
    *
-   * @param {number|Number} value the number we want to represent as an int32.
+   * @param {number|Number|string} value the number we want to represent as an int32.
```
mbroadst, line 9: kind of looks like we accept any type here now? Since you're casting, you can convert any JSValue into a numeric

test/node/int_32_tests.js
```diff
@@ -6,7 +6,9 @@ const expect = require('chai').expect;
 
 describe('Int32', function() {
   describe('Constructor', function() {
-    var value = 42;
+    const hexValue = '0x2A';
```
emadum, line 9: Should be ` const hexValue = 0x2A;` (no quotes)

reggi: @emadum We're using string because the test is checking to see if the `+` properly converts the string hex value to base 10, which it does. Try for yourself in REPL:

```js
> +"0x2a"
42
```

PR #361, opened 2020-05-14
[NODE-2613] fix: int32 parseInt in toExtendedJSON

reggi:
This is an extra PR showing how we can change `toExtendedJSON` to abide spec.

BSON 1.1 Type or Convention | Canonical Extended JSON Format | Relaxed Extended JSON Format
-- | -- | --
Int32 | {"$numberInt": <32-bit signed integer as a string>} | integer

### DIFF
--- a/lib/int_32.js
+++ b/lib/int_32.js
@@ -38,7 +38,7 @@ class Int32 {
    * @ignore
    */
   toExtendedJSON(options) {
-    if (options && (options.relaxed || options.legacy)) return this.value;
+    if (options && (options.relaxed || options.legacy)) return parseInt(this.value, 10);
     return { $numberInt: this.value.toString() };
   }
 


lib/int_32.js
```diff
@@ -38,7 +38,7 @@ class Int32 {
    * @ignore
    */
   toExtendedJSON(options) {
-    if (options && (options.relaxed || options.legacy)) return this.value;
+    if (options && (options.relaxed || options.legacy)) return parseInt(this.value, 10);
```
mbroadst, line 41: If we make a change in https://github.com/mongodb/js-bson/pull/359 to ensure that `Int32` is _always_ wrapping a `number`, then why do we ever need to `parseInt` again?

reggi: I think this was clarified in meeting, as picking one of the solutions. Sorry for the confusion. Closing this PR.

PR #364, opened 2020-05-26
fix bug when add toBSON to BigInt type and serialize a 0n value.

moriyalb:
when you use a BigInt type, and add toBSON method to the prototype
```javascript
BigInt.prototype.toBSON = function() {
	return this.toString()
}
```

When use mongodb native you make call like this:
```javascript
db.col.bulkWrite([updateOne:{$set:{value: 0n}])
```
then you will got an error something like $set is empty.

all because when the bson/serializer use toBSON custom method, it's nil value checking is wrong.

### DIFF
--- a/lib/parser/serializer.js
+++ b/lib/parser/serializer.js
@@ -686,7 +686,7 @@ function serializeInto(
       let value = object[i];
 
       // Is there an override value
-      if (value && value.toBSON) {
+      if (value != undefined && value != null && value.toBSON) {
         if (typeof value.toBSON !== 'function') throw new TypeError('toBSON is not a function');
         value = value.toBSON();
       }
@@ -880,7 +880,7 @@ function serializeInto(
     for (let key in object) {
       let value = object[key];
       // Is there an override value
-      if (value && value.toBSON) {
+      if (value != undefined && value != null && value.toBSON) {
         if (typeof value.toBSON !== 'function') throw new TypeError('toBSON is not a function');
         value = value.toBSON();
       }


moriyalb:
> i assume you want to check if it is an object and not null. But Writing `value != undefined &&` is equal to `value && `
> 
> If you actually want to have a useful check then at least type check if it is an object. Also use triple equal. Compare with the null check method in lodash.
> 
> So better
> 
> `if (typeof value === "object" && value !== null && value.toBSON)`

Thank you for pointing that. My modifications just make sure any `Type` should be checking whether has a `toBson` method.

In the origin verion of bson code

```
if (value && value.toBson) 
```

will failed when passing `0n`, however we want the call this

```
BigInt.prototype.toBSON = function() {
	return this.toString()
}
```

nbbeeken:
Hey @moriyalb Thanks for submitting this, I think as @Uzlopak suggested you'll want to strictly check 
`value !== null && value !== undefined` 
before checking for the potential toBSON property:
`value !== null && value !== undefined && typeof value.toBSON === 'function'`

As for working with BigInts specifically we have added helpers on the Long class `toBigInt` and `fromBigInt` to assist with working with 64bit bigints if that's your use case. We also think it would be nice to add helpers to the Decimal128 class, we welcome any community contributions to that end.

moriyalb:
> Hey @moriyalb Thanks for submitting this, I think as @Uzlopak suggested you'll want to strictly check
> `value !== null && value !== undefined`
> before checking for the potential toBSON property:
> `value !== null && value !== undefined && typeof value.toBSON === 'function'`
> 
> As for working with BigInts specifically we have added helpers on the Long class `toBigInt` and `fromBigInt` to assist with working with 64bit bigints if that's your use case. We also think it would be nice to add helpers to the Decimal128 class, we welcome any community contributions to that end.

I'm glad to be notified when `BigInt` type is supported

PR #366, opened 2020-06-01
fix: support using `bson.ObjectId` with webpack build

vkarpov15:
For v1.x, you can use `require('bson').ObjectId` in Node, but when you use that in the browser `require('bson').ObjectId` is undefined.


bajanam:

PR #368, opened 2020-06-09
fixes #324 - needs to merge it in another branch `3.0.3`

### LINKED ISSUE #324: bson.Objectid issue in node 12.*+
I am using node 12.* verison, and i found issue by using os modules

in line 3. just calling hostname and create random 3-byte value.

```
'use strict';

const hostname = require('os').hostname;
const fnv1a24 = require('./fnv1a').fnv1a24;

/**
 * Machine id.
 *
 * Create a random 3-byte value (i.e. unique for this
 * process). Other drivers use a md5 of the machine id here, but
 * that would mean an asyc call to gethostname, so we don't bother.
 * @ignore
 */
const MACHINE_ID = fnv1a24(hostname);
```

but it's wrong.

[please look at this docs.](https://nodejs.org/dist/latest-v8.x/docs/api/os.html#os_os_hostname)

The `os.hostname` is not default string, it's a function.

ggondim:
Updates `objectid.js` to the version in branch `3.0` (https://github.com/mongodb/js-bson/blob/3.0/lib/bson/objectid.js), because the `3.0.2` tag doesn't have it.

This fixes #324 that is happening also with mongodb-extjson when using Node v12.

I propose a `3.0.3` package release when the tests are OK.

### DIFF
--- a/lib/bson/objectid.js
+++ b/lib/bson/objectid.js
@@ -1,8 +1,4 @@
 'use strict';
-
-const hostname = require('os').hostname;
-const fnv1a24 = require('./fnv1a').fnv1a24;
-
 /**
  * Machine id.
  *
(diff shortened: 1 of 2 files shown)


bajanam:

PR #369, opened 2020-07-07
NODE-2431: Reduce floating point precision required of extended json implementations

nbbeeken:
https://jira.mongodb.org/browse/NODE-2431

Here's the fix for the tests sync'ed from bson-corpus and an update to the toExtendedJSON function for the Double type to output the correct precision.
All of our bson-corpus json files were marked as executable, that got changed when I ran the sync script which is why every file will show up with a change. (We'll see how GH ui shows this)


mbroadst:
@nbbeeken we'll need to backport the fix to the 1.0-branch as well

PR #374, opened 2020-07-17
chore(ci): evergreen integration

emadum:
### Description

Defines a testing matrix across linux,windows,mac platforms and node version 6,8,10,12,14 runtimes.

[NODE-886](https://jira.mongodb.org/browse/NODE-886)


nbbeeken:
Can you manually schedule the windows and mac ones just for this pr? since they're new it'd be interesting to see the results
Is it possible to use only the platforms axis for the browser tests? there's no change between the node tests for the karma runs (I'm pretty sure?)

emadum:
> Can you manually schedule the windows and mac ones just for this pr? since they're new it'd be interesting to see the results

[patch](https://spruce.mongodb.com/version/5f12509157e85a256ff7b7f2/tasks)

> Is it possible to use only the platforms axis for the browser tests? there's no change between the node tests for the karma runs (I'm pretty sure?)

Yeah, I moved the browser tests into their own matrix so they can target different platforms. There were some issues on windows and mac so I'm just running those on ubuntu with node v10.

Unfortunately `nvm-windows` doesn't support the `--lts` flag, making it difficult to properly test against LTS versions of node on windows. Since this isn't critical for this project, I commented out that platform. There's a [stale PR](https://github.com/coreybutler/nvm-windows/pull/422) up to add that functionality; if that gets merged in we can enable the windows tests.

mbroadst:
@emadum were you going to investigate [generate.tasks](https://github.com/evergreen-ci/evergreen/wiki/Project-Commands#generate-tasks) in this project? I think the evergreen configuration you've chosen makes it easier to maintain from a developer persepective, but ultimately makes our evergreen visual feedback very cluttered. I know that seems like an "evergreen problem", but we chose to generate tasks in the driver configuration in such a way that patch builds were better organized for visualization on EVG.

emadum:
> @emadum were you going to investigate [generate.tasks](https://github.com/evergreen-ci/evergreen/wiki/Project-Commands#generate-tasks) in this project? I think the evergreen configuration you've chosen makes it easier to maintain from a developer persepective, but ultimately makes our evergreen visual feedback very cluttered. I know that seems like an "evergreen problem", but we chose to generate tasks in the driver configuration in such a way that patch builds were better organized for visualization on EVG.

I was considering using that but it seemed a lot simpler to use the matrix approach, so I went with what seemed more straightforward. I can take another stab at it with generating tasks; agreed that the visual arrangement produced by this approach is less than ideal.

emadum:
> I can take another stab at it with generating tasks; agreed that the visual arrangement produced by this approach is less than ideal.

@mbroadst I tried using `generate.tasks` but I found the interface less than ideal - you can only schedule the "generating" task; you can't further select subsets of the generated tasks - I don't think it's a good fit for our use case.

However, I think creating a different task for each node runtime works well. I've updated the config.yml accordingly, PTAL!

cc @nbbeeken / @reggi

.evergreen/run-tests.sh
```diff
@@ -0,0 +1,34 @@
...
+  exit 1
+fi
+
+NODE_ARTIFACTS_PATH="${PROJECT_DIRECTORY}/node-artifacts"
+
+if [ "$OS" == "Windows_NT" ]; then
+    export NVM_HOME=`cygpath -w "$NODE_ARTIFACTS_PATH/nvm"`
+    export NVM_SYMLINK=`cygpath -w "$NODE_ARTIFACTS_PATH/bin"`
+    export PATH=`cygpath $NVM_SYMLINK`:`cygpath $NVM_HOME`:$PATH
+    NODE_TEST="test-node-windows"
```
nbbeeken, line 16: Whats the reason for not using `test-node` here?

emadum: The script `test-node` references `node_modules/.bin`, which doesn't exist on windows. I could probably just change `test-node` to what I put in for `test-node-windows` so I don't need the extra logic; will try that out.

nbbeeken: Wow that `.` is easy to miss, it is what it is if there's a different directory shape in windows, up to you.

PR #388, opened 2020-08-19
NODE-2716: Convert to Typescript

nbbeeken:
Converted the codebase to typescript. In order to get us up and running with testing I also did the leg work on bringing our tooling in sync with the driver, (for example we have to update prettier to get `import type` syntax). I brought in ts-node and got that working with mocha so tests are still being run and this works for both Node ~~and the in browser karma tests~~ (if the bundle is broken there's no karma tests! spoke too soon). This initial run is minimal changes needed to get us a compiling code base, unlike when we did this for the driver I opted to initially disable typescript's `strict` rule which allows me to omit type annotations on functions mainly, which in a subsequent fix up PR would all be changed from `any` to something more meaningful. 

I also brought in our gulp file from the driver, along with the api-extractor config so we are ready to go on distributing typings.
Another pass will be to add the ts-doc information about access level and correctly formatting the `@param` tags etc.

Currently this PR will break our rollup config, I'm working on getting that working with the new TS code but figured I get the ball rolling on this initial step.


nbbeeken:
Ah so something I realized after the fact, currently when you `npm install bson` you get a library that has a lib folder and a dist folder. The dist folder contains the bundles intended for the web and the lib folder was just our source, I assume we want to keep that form of distribution? I got mixed up I guess because we're now generating the lib folder. And I can generate the bundles from the generated lib folder instead of directly from the ts which works without any changes to the current rollup setup. (At least I'm pretty sure, I can revert to make sure, but that's why karma originally passed my tests without complaint, cus it was using the compiled lib folder!)
OR we can make sure our bundles are generated directly from typescript. This is fine and I'm expecting it to be preferred.

In either case:
There's the issue that was found with the es module formatted bundles, this issue is unfixable with the current method that the 'long' dependency uses to export the class `module.exports = Long` is incompatible with `import Long from 'long'`. And if you attempt to use `import * as Long from 'long'`  the result is:
```js
import * as Long from 'long';
import { prototype, fromString as fromString$1, fromNumber, ZERO, isLong, fromInt, fromBits } from 'long';
```
Which the first line give you hope for a solution but alas the second line shows you that pulling apart the class into its individual components like `prototype` is not what we desire.


As an aside, rollup (at least the rollup of today), is designed to take in esm style code and translate that to the other formats (commonjs, umd, iife) however we have a codebase written in commonjs and rely on rollup to translate that to esm. That's not in itself an issue its just the inverse of what the bundler tool is designed for today.

Gulpfile.ts
```diff
@@ -0,0 +1,104 @@
...
+    await run('npx rimraf lib/*.d.ts lib/**/*.d.ts');
+    await run('npx prettier types/bson.d.ts --write');
+  } catch (err) {
+    console.error('encountered an error:');
+    console.error((err.stdout as string).trim());
+    console.error((err.stderr as string).trim());
+  }
+});
+
+gulp.task('compile', () => {
```
mbroadst, line 91: should be able to rewrite this as:
```
gulp.task('compile', () => {
  return gulp.src('./src/**/*.ts').pipe(ts(tsConfig.compilerOptions)).pipe(gulp.dest('./lib'));
});
```

Gulpfile.ts
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  const docOptions = generateTypeDocConfig();
+
+  writeFileSync('./typedoc.json', JSON.stringify(docOptions, undefined, 2), {
+    encoding: 'utf8'
+  });
+
+  try {
+    await run('npx typedoc');
```
mbroadst, line 65: we shouldn't have to run `npx` in the gulp file, there are likely plugins for all of these things or at least programmatic apis. For using `typedoc` you can just use [gulp-typedoc](https://www.npmjs.com/package/gulp-typedoc) which will allow you to do all of this in a more "gulpy" way without having to write a temporary file and unlink it on error

Gulpfile.ts
```diff
@@ -0,0 +1,104 @@
...
+    unlinkSync('./typedoc.json');
+  }
+});
+
+gulp.task('api-extractor', async () => {
+  try {
+    const { stdout, stderr } = await run('npx api-extractor run --local');
+    console.log(stdout);
+    console.log(stderr);
+    await run('npx rimraf lib/*.d.ts lib/**/*.d.ts');
```
mbroadst, line 82: [gulp-rimraf](https://www.npmjs.com/package/gulp-rimraf)

Gulpfile.ts
```diff
@@ -0,0 +1,104 @@
...
+    await run('npx prettier types/bson.d.ts --write');
```
mbroadst, line 83: see the `ts-pipeline` for examples of programmatically using `prettier` in a gulpfile

src/db_ref.ts
```diff
@@ -1,9 +1,23 @@
-'use strict';
+import type { ObjectId } from './objectid';
+
+export interface DBRefShape {
```
mbroadst, line 3: nit: `DBRefLike` (for consistency across both codebases)

src/document.ts
```diff
@@ -0,0 +1,3 @@
+export interface Document {
```
mbroadst, line 1: maybe we can just move this to `bson.ts`? I'm loathe to add even more files to this project..

src/map.ts
```diff
@@ -0,0 +1,10 @@
+// We have an ES6 Map available, return the native instance
+
+let exportedMap: MapConstructor;
+if (typeof global.Map !== 'undefined') {
+  exportedMap = global.Map;
+} else {
+  exportedMap = require('core-js/es/map');
```
mbroadst, line 7: hm shouldn't this otherwise be polyfilled by TypeScript for us?

src/symbol.ts
```diff
@@ -1,16 +1,13 @@
...
+export class BSONSymbol {
   /**
    * Create a Symbol type
    *
    * @param {string} value the string representing the symbol.
    */
-  constructor(value) {
-    this.value = value;
-  }
+  constructor(public value: string) {}
```
mbroadst, line 10: why the `public` here?

src/timestamp.ts
```diff
@@ -1,15 +1,13 @@
...
 /**
  * @class
  * @param {number} low  the low (signed) 32 bits of the Timestamp.
  * @param {number} high the high (signed) 32 bits of the Timestamp.
  * @return {Timestamp}
  */
-class Timestamp extends Long {
-  constructor(low, high) {
+export class Timestamp extends Long {
+  constructor(low: number | Long, high?: number) {
```
mbroadst, line 10: looks like some overloads would be helpful here

Gulpfile.ts
nbbeeken: The gulp typedoc plugin didn't work for our project configuration, its not the plugins fault, I wrestled with it for quite a bit, the problem is actually that typedoc's programatic API differs from its config file API so this was my hacky solution, I do still have some clean up I can do here to make this more readable.

Gulpfile.ts
nbbeeken: I actually want to drop this behavior in both here and the driver gulp file, we should be turning on declaration maps and prettier-ing the output will ruin the maps.

src/map.ts
nbbeeken: Typescript doesn't do polyfills, that's a babel thing which is strange since we use babel to shim Buffer in browsers, we could have used babel to polyfill Map, that might still be a possible solution here instead of manually importing it (I loathe the directly added dependency). Although on the other hand perhaps we shouldn't be doing this logic of if Map exists on global and instead export the polyfill regardless since the point here is to use it on platforms where Map is not supported.

nbbeeken:
I have not found a solution to using `Long` without vendoring it and modifying it to use esm syntax, (I think I could also just modify it to do `module.exports.Long = Long`, key change being the Long key on the exports object, but I figured if I'm vendoring anyway, take it the full mile).

I still need to fix up the gulp file to use plugins correctly and the rollup config still has issues with correctly shimming Buffer.

src/map.ts
reggi: Not sure if this is covered in the bundling but browser would need `window.Map`.

src/validate_utf8.ts
```diff
@@ -18,7 +16,7 @@ const CONTINUING_CHAR = 0x80;
  * @param {Number} end The index to end validating
  * @returns {boolean} True if valid utf8
  */
-function validateUtf8(bytes, start, end) {
+export function validateUtf8(bytes, start, end) {
```
reggi, line 19: missed typings?

reggi:
Just noticed some places for code / logic improvement we can hold off on.

I also saw some untyped arguments, are we leaving these untyped and doing more incremental progress?

src/validate_utf8.ts
nbbeeken: I wasn't able to get the convert script we used on the driver to work here, so this is all done by hand, I figured that script just populated most of the codebase with `any` anyway so its fine to omit here and we will type properly in the next iteration.

nbbeeken:
There's an issue with testing, I can confirm locally that tests pass on node 6.8 but tsc supports node 10+ so I can work on updating the EVG shell files to get both versions of node, and compile with one and test with the other but that might be best to leave to another pr, let me know what's best

rollup.config.js
```diff
@@ -1,97 +1,76 @@
...
-      format: 'es',
-      name: defaultName,
-      exports: 'named'
-    },
-    plugins: browserPlugins,
-    external
+    ],
+
+    plugins: [
+      nodeResolve({ preferBuiltins: false }),
```
mbroadst, line 33: in both cases you are using `preferBuiltins: false`, and there is a removed `nodeBuiltins()` plugin. Can you explain what's going on here?

src/map.ts
```diff
@@ -0,0 +1,2 @@
+const myMap = Map;
```
mbroadst, line 1: This looks strange, can you explain why this is needed? At very least it looks like we probably don't need a whole file for this

src/symbol.ts
```diff
@@ -1,14 +1,14 @@
-'use strict';
 /**
  * A class representation of the BSON Symbol type.
  */
-class BSONSymbol {
+export class BSONSymbol {
```
mbroadst, line 4: how much of a breaking change is it to remove this type alltogether and depend on the `Symbol` type with a possible built-in polyfill? (similar to `Map`)

.evergreen/install-dependencies.sh
```diff
@@ -41,14 +41,16 @@ path: $NVM_SYMLINK
 EOT
 
   nvm install ${NODE_VERSION}
+  nvm install 10.22.0 # install lts for compilation only
```
mbroadst, line 44: should we use `nvm install --lts=dubnium` instead?

.evergreen/install-dependencies.sh
```diff
@@ -60,4 +62,5 @@ registry=https://registry.npmjs.org
 EOT
 
 # install node dependencies
-npm install
\ No newline at end of file
+npm install
```
mbroadst, line 65: hm, I don't think this will cut it. You would need to do something like:
```
nvm install --no-progress --lts=dubnium
npm install # implicitly does an `npm build:ts` or whatever
git clean -dfx
nvm install --no-progress ${NODE_VERSION}
npm install
```

Also, can you please document this more carefully, this actually somewhat tricky how it's working right now. I suspect this is working because of the `prepare` script checking for the presence of the `lib` folder and bailing from compilation otherwise. It's not clear from this source that `prepare` is being used this way, nor is it even clear that `npm install` is doing the initial compilation.

tools/scenarios-plugin.js
```diff
@@ -1,5 +1,3 @@
-'use strict';
```
mbroadst, line 1: hm, this isn't a ts file so why are we removing the `'use strict';`? Did we lose a linter rule to enforce this?

tools/scenarios-plugin.js
```diff
@@ -15,4 +13,4 @@ const scenariosPlugin = () => {
   };
 };
 
-module.exports = scenariosPlugin;
+module.exports.scenariosPlugin = scenariosPlugin;
```
mbroadst, line 16: nit: this is `6.9.2+` so you can use the shorthand property form:
```
module.exports = { scenariosPlugin };
```

test/register-bson.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+// All test files import BSON from one place here
+// This was born out of debugging our karma configuration
+// But it can also be a convenient way to run the tests against an alternative implementation
+// If you want to run the tests against the typescript directly, change the 'lib' to 'src'
+// and make sure you run mocha using our .mocharc.json or with --require ts-node/register
+
+// This should be done by mocha --require, but that isn't supported until mocha version 7+
+require('chai/register-expect');
```
mbroadst, line 10: Is this file really about code dupe? If we want to test against typescript then we should just use `ts-node` for everything, right?

nbbeeken: I wasn't able to get rollup to compile our ts and correctly generate the bundles so the tests make use of the compiled js from typescript because karma relies on rollup to bundle all of mocha into a browser ready bundle

tools/scenarios-plugin.js
nbbeeken: It's because its falling under the settings of the root eslintrc which disallows 'use strict'; vs in the test subdirectory the eslint config requires it. I brought over the config.

.evergreen/install-dependencies.sh
nbbeeken: I added a comment about npm install doing the compilation with prepare.
If you run `git clean` you will delete the compiled lib folder and the downloaded dependencies. The prepare script doesn't bail if there's a lib folder, it checks for a src directory which is only present if you cloned the repo. It does also bail if the node version is too old, but it will warn you if you lack a lib folder, and even further warn you that you can't compile the lib on older versions of node. Here's the prepare script:
```js
#! /usr/bin/env node
var cp = require('child_process');
var fs = require('fs');

var nodeMajorVersion = +process.version.match(/^v(\d+)\.\d+/)[1];

if (fs.existsSync('src') && nodeMajorVersion >= 10) {
  cp.spawnSync('npm', ['run', 'build'], { stdio: 'inherit', shell: true });
} else {
  if (!fs.existsSync('lib')) {
    console.warn('BSON: No compiled javascript present, the library is not installed correctly.');
    if (nodeMajorVersion < 10) {
      console.warn(
        'This library can only be compiled in nodejs version 10 or later, currently running: ' +
          nodeMajorVersion
      );
    }
  }
}
```

src/symbol.ts
nbbeeken: The class's name is BSONSymbol and we export it as such, not sure how we can drop that. Also Symbols don't have a .value property, and they're not constructed i.e. `new Symbol` fails.

src/map.ts
nbbeeken: So I believed this was getting polyfilled by babel this was my way of plucking out the Map constructor and the generated code here would look like an import or injection of the Map type. Turns out that's not the case so I'm back to manually importing it. Which... breaks the bundle again, not sure why but commonjs isn't translating the core-js dependency. I'm ready to let this goal go.

In the latest commit I brought back the old polyfil, its not used by most but I can improve it in the next cycle of typing, if that makes sense?

rollup.config.js
nbbeeken: In the `dist/bson.esm.js` case it is a redundant setting since we have  `external: ['buffer']` it won't bundle it the locally installed one nor the node one. This bundle is the odd one out kinda, and whether or not we need nodePolyfills (replacement for nodeBuiltins)  depends on who is consuming it. If the target is just node then I can just have the commonjs() plugin only I believe.

Just some side info:
I learned on my bundling journey that these appear to be abandoned:
rollup-plugin-node-globals
rollup-plugin-node-builtins
This is a remake of the builtins lib and rollup wants to bring it into their monorepo for plugins but that has yet to happen
rollup-plugin-node-polyfills

src/timestamp.ts
```diff
@@ -1,20 +1,21 @@
...
+export class Timestamp extends Long {
+  constructor(low: Long);
+  constructor(low: number, high: number);
+  constructor(low: number | Long, high?: number) {
     if (Long.isLong(low)) {
       super(low.low, low.high, true);
     } else {
       super(low, high, true);
     }
+    this._bsontype = 'Timestamp';
```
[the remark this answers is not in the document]
nbbeeken: There's a crash I get where it says you can't write _bsontype to readonly property. But only on the browser tests, I think this is because the generated bundles make exports from other modules readonly? 🤔 so like Timestamp.prototype is Long, and that's readonly. Not sure, but I'm investigating, threw up a potential solution in the meantime.

.evergreen/install-dependencies.sh
nbbeeken: Windows nvm doesn't appear to support an lts alias, I can make a ticket to add some curl work to get the latest dubnium.

mbroadst: oh duh, right right sorry I forgot about nvm-windows

PR #389, opened 2020-09-03
NODE-2717: Improve TS Typings

nbbeeken:
Improved the typings, a few any cases had to be left in to favor leaving code more or less untouched rather than refactoring.

### DIFF
--- a/src/binary.ts
+++ b/src/binary.ts
@@ -1,74 +1,51 @@
 import { Buffer } from 'buffer';
+import type { EJSONOptions } from './extended_json';
+import { haveBuffer, isBuffer, isUint8Array } from './parser/utils';
 
-/**
- * A class representation of the BSON Binary type.
- */
+type BinarySequence = Uint8Array | Buffer | number[];
+
+export type BinaryEJSON =
+  | {
+      $type: string;
+      $binary: string;
+    }
+  | {
+      $binary: {
+        subType: string;
+        base64: string;
+      };
+    };
+
+/** A class representation of the BSON Binary type. */
 export class Binary {
-  static readonly BUFFER_SIZE = 256;
+  _bsontype!: 'Binary';
 
-  /**
-   * Default BSON type
-   *
-   * @classconstant SUBTYPE_DEFAULT
-   **/
+  /** Initial buffer default size */
+  static readonly BUFFER_SIZE = 256;
+  /** Default BSON type */
   static readonly SUBTYPE_DEFAULT = 0;
-  /**
-   * Function BSON type
-   *
-   * @classconstant SUBTYPE_DEFAULT
-   **/
+  /** Function BSON type */
   static readonly SUBTYPE_FUNCTION = 1;
-  /**
-   * Byte Array BSON type
-   *
-   * @classconstant SUBTYPE_DEFAULT
-   **/
+  /** Byte Array BSON type */
   static readonly SUBTYPE_BYTE_ARRAY = 2;
-  /**
-   * OLD UUID BSON type
-   *
-   * @classconstant SUBTYPE_DEFAULT
-   **/
+  /** Deprecated UUID BSON type @deprecated Please use SUBTYPE_UUID */
   static readonly SUBTYPE_UUID_OLD = 3;
-  /**
-   * UUID BSON type
-   *
-   * @classconstant SUBTYPE_DEFAULT
-   **/
+  /** UUID BSON type */
   static readonly SUBTYPE_UUID = 4;
-  /**
-   * MD5 BSON type
-   *
-   * @classconstant SUBTYPE_DEFAULT
-   **/
+  /** MD5 BSON type */
   static readonly SUBTYPE_MD5 = 5;
-  /**
-   * User BSON type
-   *
-   * @classconstant SUBTYPE_DEFAULT
-   **/
+  /** User BSON type */
   static readonly SUBTYPE_USER_DEFINED = 128;
 
-  buffer: any;
+  buffer: BinarySequence;
   sub_type: number;
   position: number;
 
   /**
-   * Create a Binary type
-   *
-   * Sub types
-   *  - **BSON.BSON_BINARY_SUBTYPE_DEFAULT**, default BSON type.
-   *  - **BSON.BSON_BINARY_SUBTYPE_FUNCTION**, BSON function type.
-   *  - **BSON.BSON_BINARY_SUBTYPE_BYTE_ARRAY**, BSON byte array type.
-   *  - **BSON.BSON_BINARY_SUBTYPE_UUID**, BSON uuid type.
-   *  - **BSON.BSON_BINARY_SUBTYPE_MD5**, BSON md5 type.
-   *  - **BSON.BSON_BINARY_SUBTYPE_USER_DEFINED**, BSON user defined type.
-   *
-   * @param {Buffer} buffer a buffer object containing the binary data.
-   * @param {Number} [subType] the option binary type.
-   * @return {Binary}
+   * @param buffer - a buffer object containing the binary data.
+   * @param subType - the option binary type.
    */
-  constructor(buffer, subType) {
+  constructor(buffer: string | BinarySequence, subType?: number) {
     if (
       buffer != null &&
       !(typeof buffer === 'string') &&
@@ -86,7 +63,7 @@ export class Binary {
       // Only accept Buffer, Uint8Array or Arrays
       if (typeof buffer === 'string') {
         // Different ways of writing the length of the string for the different types
-        if (typeof Buffer !== 'undefined') {
+        if (haveBuffer()) {
           this.buffer = Buffer.from(buffer);
         } else if (typeof Uint8Array !== 'undefined' || Array.isArray(buffer)) {
           this.buffer = writeStringToArray(buffer);
@@ -98,7 +75,7 @@ export class Binary {
       }
       this.position = buffer.length;
     } else {
-      if (typeof Buffer !== 'undefined') {
+      if (haveBuffer()) {
         this.buffer = Buffer.alloc(Binary.BUFFER_SIZE);
       } else if (typeof Uint8Array !== 'undefined') {
         this.buffer = new Uint8Array(new ArrayBuffer(Binary.BUFFER_SIZE));
@@ -111,30 +88,29 @@ export class Binary {
   /**
    * Updates this binary with byte_value.
    *
-   * @method
-   * @param {string} byte_value a single byte we wish to write.
+   * @param byteValue - a single byte we wish to write.
    */
-  put(byte_value) {
+  put(byteValue: string | number | Uint8Array | Buffer | number[]): void {
     // If it's a string and a has more than one character throw an error
-    if (byte_value['length'] != null && typeof byte_value !== 'number' && byte_value.length !== 1)
+    if (byteValue['length'] != null && typeof byteValue !== 'number' && byteValue.length !== 1)
       throw new TypeError('only accepts single character String, Uint8Array or Array');
-    if ((typeof byte_value !== 'number' && byte_value < 0) || byte_value > 255)
+    if (typeof byteValue === 'number' && (byteValue < 0 || byteValue > 255))
       throw new TypeError('only accepts number in a valid unsigned byte range 0-255');
 
     // Decode the byte value once
     let decoded_byte = null;
-    if (typeof byte_value === 'string') {
-      decoded_byte = byte_value.charCodeAt(0);
-    } else if (byte_value['length'] != null) {
-      decoded_byte = byte_value[0];
+    if (typeof byteValue === 'string') {
+      decoded_byte = byteValue.charCodeAt(0);
+    } else if (byteValue['length'] != null) {
+      decoded_byte = byteValue[0];
     } else {
-      decoded_byte = byte_value;
+      decoded_byte = byteValue;
     }
 
     if (this.buffer.length > this.position) {
       this.buffer[this.position++] = decoded_byte;
     } else {
-      if (typeof Buffer !== 'undefined' && Buffer.isBuffer(this.buffer)) {
+      if (isBuffer(this.buffer)) {
         // Create additional overflow buffer
         const buffer = Buffer.alloc(Binary.BUFFER_SIZE + this.buffer.length);
         // Combine the two buffers together
@@ -166,24 +142,22 @@ export class Binary {
   /**
    * Writes a buffer or string to the binary.
    *
-   * @method
-   * @param {(Buffer|string)} string a string or buffer to be written to the Binary BSON object.
-   * @param {number} offset specify the binary of where to write the content.
-   * @return {null}
+   * @param sequence - a string or buffer to be written to the Binary BSON object.
+   * @param offset - specify the binary of where to write the content.
    */
-  write(string, offset) {
+  write(sequence: string | BinarySequence, offset: number): void {
     offset = typeof offset === 'number' ? offset : this.position;
 
     // If the buffer is to small let's extend the buffer
-    if (this.buffer.length < offset + string.length) {
+    if (this.buffer.length < offset + sequence.length) {
       let buffer = null;
       // If we are in node.js
-      if (typeof Buffer !== 'undefined' && Buffer.isBuffer(this.buffer)) {
-        buffer = Buffer.alloc(this.buffer.length + string.length);
+      if (isBuffer(this.buffer)) {
+        buffer = Buffer.alloc(this.buffer.length + sequence.length);
         this.buffer.copy(buffer, 0, 0, this.buffer.length);
       } else if (isUint8Array(this.buffer)) {
         // Create a new buffer
-        buffer = new Uint8Array(new ArrayBuffer(this.buffer.length + string.length));
+        buffer = new Uint8Array(new ArrayBuffer(this.buffer.length + sequence.length));
         // Copy the content
         for (let i = 0; i < this.position; i++) {
           buffer[i] = this.buffer[i];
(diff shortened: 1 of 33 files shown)


src/constants.ts
```diff
@@ -6,194 +6,89 @@ export const BSON_INT64_MAX = Math.pow(2, 63) - 1;
...
- **/
+/** Binary Deprecated UUID Type @deprecated Please use BSON_BINARY_SUBTYPE_UUID_NEW */
 export const BSON_BINARY_SUBTYPE_UUID = 3;
 
-/**
- * Binary MD5 Type
- *
- * @classconstant BSON_BINARY_SUBTYPE_MD5
- **/
-export const BSON_BINARY_SUBTYPE_MD5 = 4;
```
nbbeeken, line 192: We don't use this anywhere internally, but we do export it.
I added tests to make sure we have the correct values, who knows typos happen best to guard against that.

src/extended_json.ts
```diff
@@ -1,20 +1,50 @@
...
+  | ObjectId
+  | BSONRegExp
+  | BSONSymbol
+  | Timestamp;
+
+export function isBSONType(value: unknown): value is BSONType {
+  return (
+    isObjectLike(value) && Reflect.has(value, '_bsontype') && typeof value._bsontype === 'string'
+  );
+}
```
nbbeeken, line 36: This makes for some very nice TS
```
if (isBSONType(x)) {
  x._bsontype // union of all the type names
  if(x._bsontype === 'Double') {
    x.toExtendedJSON() // function exists and return type can be narrowed to `{$numberDouble: string}` 
  }
}
```

src/parser/deserializer.ts
```diff
@@ -255,9 +298,13 @@ function deserializeObject(buffer, index, options, isArray) {
       // Update index
       index = index + 16;
       // Assign the new Decimal128 value
-      const decimal128 = new Decimal128(bytes);
+      const decimal128 = new Decimal128(bytes) as Decimal128 | { toObject(): unknown };
       // If we have an alternative mapper use that
-      object[name] = decimal128.toObject ? decimal128.toObject() : decimal128;
+      if ('toObject' in decimal128 && typeof decimal128.toObject === 'function') {
+        object[name] = decimal128.toObject();
```
nbbeeken, line 304: Here's the special case mapper I mentioned in triage today

src/parser/serializer.ts
```diff
@@ -583,13 +689,19 @@ function serializeBinary(buffer, key, value, index, isArray?: boolean) {
   }
 
   // Write the data to the object
-  data.copy(buffer, index, 0, value.position);
+  buffer.set(data, index);
```
nbbeeken, line 692: This was a bug in the browser bundles, data can be a Uint8Array, I tried going back to the old bundles to figure out how this worked but I couldn't get it to work.
Anyway `set` is the Uint8Array equivalent of buffer's copy and is available on both.

src/timestamp.ts
```diff
@@ -1,15 +1,24 @@
...
-export class Timestamp extends Long {
-  _bsontype: string;
+type TimestampOverrides = '_bsontype' | 'toExtendedJSON' | 'fromExtendedJSON';
+type LongWithoutOverrides = new (low: number | Long, high?: number, unsigned?: boolean) => {
+  [P in Exclude<keyof Long, TimestampOverrides>]: Long[P];
+};
+const LongWithoutOverridesClass: LongWithoutOverrides = (Long as unknown) as LongWithoutOverrides;
 
+export class Timestamp extends LongWithoutOverridesClass {
+  _bsontype!: 'Timestamp';
```
nbbeeken, line 10: I know what you're thinking.. this looks gross! 😅 
This is the only case where we extend from another BSON type but this means we can't just have different types for `_bsontype` and from/toExtendedJSON so this does some trickery to create a Long type that doesn't have those properties. The generated code is not effected it still extends Long as normal.

src/max_key.ts
```diff
@@ -2,17 +2,15 @@
  * A class representation of the BSON MaxKey type.
  */
 export class MaxKey {
-  /**
-   * @ignore
-   */
-  toExtendedJSON() {
+  _bsontype!: 'MaxKey';
```
reggi, line 5: Is this necessary? If so shouldn't it be `{ value: ... }`?

reggi (approved):

src/max_key.ts
reggi: I noticed this around a lot, would love to learn more about the difference between this and the `prototype` addition to the class. Could just be `_bsontype = { value: "MaxKey" }` curious why not.

nbbeeken: We still want it to be a string so we would want it to be `_bsontype = "MaxKey"` but defining it this way makes it an editable property on the MaxKey instance and not on its prototype. Also by defining it with a call to `Object.defineProperty` by default the property descriptor is 
```
{
  value: 'MaxKey',
  writable: false,
  enumerable: false,
  configurable: false
}
```
So it is neither writable, enumerable, nor configurable. In addition to this, and I could be mistaken but the prototype is not editable outside the module so you can't change this once its exported which is good for security.

src/objectid.ts
```diff
@@ -42,10 +33,17 @@ function makeObjectIdError(invalidString, index) {
   );
 }
 
+export interface ObjectIdLike {
```
mbroadst, line 36: shouldn't this be:
```
export type ObjectIdLike = ObjectId | {
  id: string | Buffer;
  toHexString(): string;
}
```

src/objectid.ts
```diff
@@ -185,39 +220,38 @@ export class ObjectId {
...
-   * @param {String} format The Buffer toString format parameter.
-   * @return {String} return the 24 byte hex string representation.
-   * @ignore
+   * @param format - The Buffer toString format parameter.
+   * @internal
    */
   toString(format?: BufferEncoding): string {
     // Is the id a buffer then use the buffer toString method to return the format
-    if (this.id && 'copy' in (this.id as any)) {
+    if (this.id && typeof this.id !== 'string' && 'copy' in (this.id as Buffer)) {
```
mbroadst, line 230: why the added check for `string`?

src/objectid.ts
```diff
@@ -56,18 +54,20 @@ export class ObjectId {
...
+      this.id = id.id;
+      this.__id = id.__id;
+      return;
+    }
 
     // The most common use case (blank id, new objectId instance)
     if (id == null || typeof id === 'number') {
       // Generate a new id
-      this.id = ObjectId.generate(id);
+      this.id = ObjectId.generate(typeof id === 'number' ? id : undefined);
```
mbroadst, line 70: now I see where all the checks for `string` are coming from. Let's solve this once and for all, the type for `id` should just be a `Buffer`

src/parser/serializer.ts
mbroadst: This might be fixed with the proposal to use `Buffer` everywhere

src/validate_utf8.ts
```diff
@@ -11,12 +11,11 @@ const CONTINUING_CHAR = 0x80;
...
- * @param {Buffer|Uint8Array} bytes An array of 8-bit bytes. Must be indexable and have length property
- * @param {Number} start The index to start validating
- * @param {Number} end The index to end validating
- * @returns {boolean} True if valid utf8
+ * @param bytes - An array of 8-bit bytes. Must be indexable and have length property
+ * @param start - The index to start validating
+ * @param end - The index to end validating
  */
-export function validateUtf8(bytes, start, end) {
+export function validateUtf8(bytes: Buffer | Uint8Array, start: number, end: number): boolean {
```
mbroadst, line 18: looks like it should be `BufferLike`?

src/timestamp.ts
mbroadst: Another way to accomplish this would be to make something like:
```
class LongBase { ... the vendored Long implementation ... }

export class Long extends LongBase {
  _bsontype!: 'Long';
  fromExtendedJSON(..) {}
  toExtendedJSON(..) {}
}

export class Timestamp extends LongBase {
  _bsontype!: 'Timestamp';
  fromExtendedJSON(..) {}
  toExtendedJSON(..) {}
}
```

This approach might make it easier to use `BigInt` in the future as well?

mbroadst (changes requested):

src/float_parser.ts
```diff
@@ -30,9 +30,17 @@
 //
 // Modifications to writeIEEE754 to support negative zeroes made by Brian White
 
-export function readIEEE754(buffer, offset, endian, mLen, nBytes) {
-  let e;
-  let m;
+type NumericalSequence = { [index: number]: number };
```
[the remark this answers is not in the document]
nbbeeken: It is true we only pass Buffer into writeIEE754, but this seems like a case where being generic is better than being specific. These functions don't use any Buffer specific apis, they don't even use the length property, that makes this type the most accepting that will still work with these functions. Narrowing this to buffer will add a type import that just an extra cycle in the type checker that doesn't get used.

src/binary.ts
```diff
@@ -1,74 +1,51 @@
 import { Buffer } from 'buffer';
+import type { EJSONOptions } from './extended_json';
+import { haveBuffer, isBuffer, isUint8Array } from './parser/utils';
 
-/**
- * A class representation of the BSON Binary type.
- */
+type BinarySequence = Uint8Array | Buffer | number[];
```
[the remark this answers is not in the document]
nbbeeken: Unlike NumericalSequence this is a union of types that we specifically depend on their APIs for. For example our buffer prop can be Buffer or Uint8Array, this union makes us check whether we have one type or the other before using the methods they provide.

src/binary.ts
```diff
@@ -86,7 +63,7 @@ export class Binary {
       // Only accept Buffer, Uint8Array or Arrays
       if (typeof buffer === 'string') {
         // Different ways of writing the length of the string for the different types
-        if (typeof Buffer !== 'undefined') {
+        if (haveBuffer()) {
```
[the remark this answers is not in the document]
nbbeeken: I do agree we should consolidate the type one thing to make sense of some confusing code. I think we can do this in a seperate unit of work in order to parallelize this.

src/binary.ts
```diff
@@ -300,38 +270,31 @@ export class Binary {
...
 
-  /**
-   * @ignore
-   */
-  toJSON() {
-    return this.buffer != null ? this.buffer.toString('base64') : '';
+  /** @internal */
+  toJSON(): string {
+    if (!this.buffer) return '';
+    const buffer = Buffer.from(this.buffer as Uint8Array);
```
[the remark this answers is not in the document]
nbbeeken: We are preserving a bug, or potentially a bug was uncovered when the bundle formats started working. Further discussion taken to slack, but this is best left as a seperate unit of work.

src/bson.ts
```diff
@@ -85,13 +88,12 @@ export const EJSON = {
   deserialize: EJSON_deserialize
 };
 
-export interface Document {
+export interface BSONDocument {
```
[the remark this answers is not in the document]
nbbeeken: Since there are DOM types here Document is the type of the global `document` object. We can rename this in the driver import but it seems best that this library doesn't clash.

src/code.ts
```diff
@@ -1,45 +1,39 @@
-import type { Document } from './bson';
+import type { BSONDocument } from './bson';
 
-/**
- * A class representation of the BSON Code type.
- */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type CodeFunction = (...args: any[]) => any;
```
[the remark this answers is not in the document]
nbbeeken: Is there something confusing about this type?
We can Function just throws warnings which I can silence, maybe even globally in the config but if not any funciton that works with this type will have to be annotated. This accepts any function.

src/code.ts
```diff
@@ -1,45 +1,39 @@
...
 
     return { $code: this.code };
   }
 
-  /**
-   * @ignore
-   */
-  static fromExtendedJSON(doc) {
+  /** @internal */
+  static fromExtendedJSON(doc: { $code: string | CodeFunction; $scope?: BSONDocument }): Code {
```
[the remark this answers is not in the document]
nbbeeken: I typed this param in all the other `fromExtendedJSON` functions it helps to ensure that the correct properties exist, otherwise you'll make a Code object with an undefined code property, seems like an unexpected behavoir, no?

src/extended_json.ts
```diff
@@ -1,20 +1,50 @@
...
+
+export type EJSONDocument = { [key: string]: ReturnType<BSONType['toExtendedJSON']> };
```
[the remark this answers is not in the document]
nbbeeken: Yes! unfortunately it also includes the relaxed return types as well, I wish there was a way to seperate them but this is a good start!

src/extended_json.ts
```diff
@@ -172,65 +192,67 @@ const BSON_INT32_MAX = 0x7fffffff,
...
+ * ```
  */
+export function stringify(value: BSONDocument): string;
+export function stringify(value: BSONDocument, options?: EJSONOptions): string;
 export function stringify(
-  value: any,
-  replacer?: (this: any, key: string, value: any) => any,
+  value: BSONDocument,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  replacer?: (number | string)[] | ((this: any, key: string, value: any) => any) | EJSONOptions,
```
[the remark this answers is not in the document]
nbbeeken: The only reason it can't is that it would be unpleasant for users, their callbacks which might be inlined declarations would have to do type assertions
```js
EJSON.stringify(doc, (k, v) => {
  return v.toFixed(2); // maybe I know my values are numbers but I can't just do this freely anymore if we use unknown
})
```
It would also add a type incompatibility with JSON.stringify, so you couldn't share replacers.

src/extended_json.ts
```diff
@@ -278,13 +303,30 @@ function serializeValue(value, options) {
...
 
 const BSON_TYPE_MAPPINGS = {
-  Binary: o => new Binary(o.value(), o.subtype),
-  Code: o => new Code(o.code, o.scope),
-  DBRef: o => new DBRef(o.collection || o.namespace, o.oid, o.db, o.fields), // "namespace" for 1.x library backwards compat
-  Decimal128: o => new Decimal128(o.bytes),
-  Double: o => new Double(o.value),
-  Int32: o => new Int32(o.value),
-  Long: o =>
+  Binary: (o: {
```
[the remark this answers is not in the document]
nbbeeken: Binary has `sub_type` and this accepts `subtype`, they're not compatible types.

src/timestamp.ts
nbbeeken: This is a good solution!
But there's a type issues that arise in a few places, namely Long.isLong asserts that something is `LongBase` and then we would need a further check to see if something is Long with _bsontype
My solution lives only in TS land and doesn't effect code output. I know its a hack kinda but it feels like the right way to go to keep types consistent.

src/validate_utf8.ts
nbbeeken: Actually this should be more generic and accept any sequence of numbers, if you think that works here

nbbeeken (commented):
My orignal comments got lost in a browser refresh, please bare with the terseness of these as I tried to quickly make up for lost time.

src/bson.ts
```diff
@@ -159,18 +156,18 @@ export function serialize(object, options) {
...
- * @param {Number} [options.index] the index in the buffer where we wish to start serializing into.
- * @return {Number} returns the index pointing to the last written byte in the buffer.
+ * @param object - the Javascript object to serialize.
+ * @param finalBuffer - the Buffer you pre-allocated to store the serialized BSON object.
+ * @returns the index pointing to the last written byte in the buffer.
  */
-export function serializeWithBufferAndIndex(object, finalBuffer, options) {
-  options = options || {};
+export function serializeWithBufferAndIndex(
+  object: BSONDocument,
```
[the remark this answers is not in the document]
nbbeeken: I think you meant this on the EJSON APIs, no? if you pass primitives to the BSON serializer you get unexpected results, like a string 'hello' turns into `{ '0': 'h', '1': 'e', '2': 'l', '3': 'l', '4': 'o' }`.
I've updated the EJSON apis

emadum (approved):

src/extended_json.ts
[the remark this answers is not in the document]
mbroadst: oof, that's unintentional I'm sure and probably a bug (it was never called `subtype`, even in [0.2.x](https://github.com/mongodb/js-bson/blob/0.2.x/lib/bson/binary.js#L31))! My comment generally applies to this whole section though

src/parser/deserializer.ts
```diff
@@ -452,7 +499,7 @@ function deserializeObject(buffer, index, options, isArray) {
       if (evalFunctions) {
         // If we have cache enabled let's look for the md5 of the function in the cache
         if (cacheFunctions) {
-          const hash = cacheFunctionsCrc32 ? crc32(functionString) : functionString;
+          const hash = cacheFunctionsCrc32 && crc32 ? crc32(functionString) : functionString;
```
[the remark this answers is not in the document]
mbroadst: oh I see, I missed the assignment to `null`. Let's do this in follow-on work

src/timestamp.ts
mbroadst: looks like the [definitely typed typings](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/bson/index.d.ts#L210) use a similar approach to what I suggested?

src/objectid.ts
mbroadst: looks like this comment was lost in the page refresh

src/objectid.ts
mbroadst: This was lost as well, we can do it in follow-on work but please make a ticket for that

mbroadst (changes requested):

src/objectid.ts
nbbeeken: Sorry about that again, I don't see the benefit to unioning the two if something accepts ObjectIdLike things it will accept ObjecId as well since it satisfies the minimal shape of the ObjectIdLike interface. Meanwhile if something accepts `ObjectIdLike | ObjectId` things you can have runtime errors if you don't check that the additional properties on ObjectId exist on the thing you're operating on. 
For example PromiseLike is only a .then method and doesn't include anything more. Also our DBRefLike isn't a union of the actual DBRef becuase it is supposed to be an object that lacks the toExtendedJSON method as an example.

src/db_ref.ts
```diff
@@ -1,37 +1,39 @@
...
- */
+/** A class representation of the BSON DBRef type. */
 export class DBRef {
+  _bsontype!: 'DBRef';
+
   collection: string;
   oid: ObjectId;
-  db: string;
-  fields: any;
+  db?: string | null;
```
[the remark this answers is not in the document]
nbbeeken: Whoops good catch

src/extended_json.ts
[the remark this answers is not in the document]
nbbeeken: Fixed!

src/timestamp.ts
nbbeeken: Oh yea I could paramertize the Long class, I'll look into that.

src/objectid.ts
nbbeeken: NODE-2821

src/code.ts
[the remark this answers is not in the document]
mbroadst: Hm, I don't know where my comment to this went.. I'm wondering because in spirit the `Code` type is meant to be a wrapper around a JavaScript function, so it seemed weird to me that we would introduce a new type which accepts any function rather than just using the primitive type.

Maybe we can disable the `Function` rule just for the files where we need to use it? When it comes to BSON there will have to be exceptions, we already disable linter rules for `any` in a number of key places because.. well you need to support `any` sometimes! I'd rather not introduce new types just to get around a linter warning, if it's going to make the code less readable

mbroadst (approved):
LGTM! I left one comment about `CodeFunction` that I feel like 4/10 about, feel free to just merge if you strongly disagree with my position.

PR #394, opened 2020-09-15
refactor: Consolidate code based on Buffer Presence

nbbeeken:
Update code to rely on Buffer class presence, removes redundant code to handle supporting typed arrays internally

NODE-2722

## Description

Something of note, TypedArrays are now serialized as BSON Binary. I have a code snippet below that breaks apart the details for the changes, This seems like a bug, but I'm looking for feedback, I think a primary consideration is to think about query-ability, although sometimes plain binary persistence is desired while some other meta data is best kept for querying.

- Should TypedArrays instead be serialized to BSON arrays with the closest matching type, examples:
  - Float32Arrays become a BSON array of BSON Double. 
  - Int32Arrays become BSON array of BSON Int32 meanwhile Uint32Arrays would become BSON Array of int64s
- Or should TypedArrays be considered binary blobs and kept as a sequence of bytes

I went with the direction of relying on Buffer being present (it's not bundled into BSON). I can however take this work in the opposite direction and eliminate the usage of Buffer in favor of JS TypedArrays and ArrayBuffer. The cost of removing Buffer is roping in either dependencies or vendor-ing code to support the features of Buffer we rely on, for example base64 translations or utf8 parsing. (In order to read an write utf8 text from TypedArrays we would need a TextEncoder / TextDecoder polyfill, for older versions of node)

```js
const BSON = require('PATH_TO_OLD_BSON');
const newBSON = require('PATH_TO_NEW_BSON');

// CURRENT

var b = BSON.serialize({ floats: new Float32Array([12.34, 34, 21.23, 3]) })
console.log(b.toString('hex'))
console.log(BSON.deserialize(b))
/*
36000000        // size
03              // Embedded Document (not array! would be 0x04)
666c6f61747300  // "floats\0"
   29000000           // size (embedded)
   01 3000            // type double and key '0\0'
   0000008014ae2840
   10 3100            // type int32 and key '1\0'
   22000000
   01 3200            // type double and key '2\0'
   00000040e13a3540
   10 3300            // type int32 and key '3\0'
   03000000
   00                 // doc null (embedded)
00              // doc null

{
  floats: { '0': 12.34000015258789, '1': 34, '2': 21.229999542236328, '3': 3 }
}
*/

// NEW

var b = newBSON.serialize({ floats: new Float32Array([12.34, 34, 21.23, 3]) })
console.log(b.toString('hex'))
console.log(newBSON.deserialize(b))
/*

22000000                              // size
05                                    // Binary type
666c6f61747300                        // "floats\0"
   10000000                           // 16 bytes
   00                                 // subtype 0
   a4704541000008420ad7a94100004040   // Bytes
00                                    // doc null

{
  floats: Binary {
    sub_type: 0,
    buffer: <Buffer a4 70 45 41 00 00 08 42 0a d7 a9 41 00 00 40 40>,
    position: 16
  }
}
*/
```

Another note: It's not easy to round trip the current output, so we should at least try and make this code produce an array I believe:
for example, `Float32Array.from({ '0': 12.34000015258789, '1': 34, '2': 21.229999542236328, '3': 3 })` doesn't work, you'd have to manually iterate the keys.


test/node/promote_values_test.js
```diff
@@ -11,290 +11,8 @@ describe('promote values', function () {
...
-      108,
-      108,
-      0,
-      0
-    ];
+    // prettier-ignore
+    var bytes = [26, 1, 0, 0, 7, 95, 105, 100, 0, 161, 190, 98, 75, 118, 169, 3, 0, 0, 3, 0, 0, 4, 97, 114, 114, 97, 121, 0, 26, 0, 0, 0, 16, 48, 0, 1, 0, 0, 0, 16, 49, 0, 2, 0, 0, 0, 16, 50, 0, 3, 0, 0, 0, 0, 2, 115, 116, 114, 105, 110, 103, 0, 6, 0, 0, 0, 104, 101, 108, 108, 111, 0, 3, 104, 97, 115, 104, 0, 19, 0, 0, 0, 16, 97, 0, 1, 0, 0, 0, 16, 98, 0, 2, 0, 0, 0, 0, 9, 100, 97, 116, 101, 0, 161, 190, 98, 75, 0, 0, 0, 0, 7, 111, 105, 100, 0, 161, 190, 98, 75, 90, 217, 18, 0, 0, 1, 0, 0, 5, 98, 105, 110, 97, 114, 121, 0, 7, 0, 0, 0, 2, 3, 0, 0, 0, 49, 50, 51, 16, 105, 110, 116, 0, 42, 0, 0, 0, 1, 102, 108, 111, 97, 116, 0, 223, 224, 11, 147, 169, 170, 64, 64, 11, 114, 101, 103, 101, 120, 112, 0, 102, 111, 111, 98, 97, 114, 0, 105, 0, 8, 98, 111, 111, 108, 101, 97, 110, 0, 1, 15, 119, 104, 101, 114, 101, 0, 25, 0, 0, 0, 12, 0, 0, 0, 116, 104, 105, 115, 46, 120, 32, 61, 61, 32, 51, 0, 5, 0, 0, 0, 0, 3, 100, 98, 114, 101, 102, 0, 37, 0, 0, 0, 2, 36, 114, 101, 102, 0, 5, 0, 0, 0, 116, 101, 115, 116, 0, 7, 36, 105, 100, 0, 161, 190, 98, 75, 2, 180, 1, 0, 0, 2, 0, 0, 0, 10, 110, 117, 108, 108, 0, 0];
```
nbbeeken, line 15: 320 line file -> 38 line file 😌 I felt this was just the right thing to do

tsconfig.json
```diff
@@ -20,10 +20,11 @@
     // Generate separate source maps files with sourceContent included
     "sourceMap": true,
     "inlineSourceMap": false,
-    "inlineSources": true,
+    "inlineSources": false,
     // API-extractor makes use of the declarations, npm script should be cleaning these up
     "declaration": true,
-    "declarationMap": true
+    "declarationMap": true,
+    "types": []
```
nbbeeken, line 27: This is to avoid any types being relied upon, namely the `@types/node` I want there to always have to be an `import {Buffer} from 'buffer'` where needed, and not rely on global namespace types.

tsconfig.json
```diff
@@ -20,10 +20,11 @@
     // Generate separate source maps files with sourceContent included
     "sourceMap": true,
     "inlineSourceMap": false,
-    "inlineSources": true,
+    "inlineSources": false,
```
nbbeeken, line 23: To sync with the driver, `src` will be shipped, no need to inline sources.

src/binary.ts
```diff
@@ -37,51 +44,45 @@ export class Binary {
...
    */
-  constructor(buffer: string | BinarySequence, subType?: number) {
+  constructor(buffer?: string | BinarySequence, subType?: number) {
     if (
-      buffer != null &&
+      !(buffer === undefined || buffer === null) &&
       !(typeof buffer === 'string') &&
-      !Buffer.isBuffer(buffer) &&
-      !(buffer instanceof Uint8Array) &&
+      !ArrayBuffer.isView(buffer) &&
```
nbbeeken, line 59: `ArrayBuffer.isView()` returns true for any typed array class, this includes Buffer since it is a subclass of Uint8Array, as well a DataView. ensureBuffer now handles bytes-ifying any typed array / view

reggi (approved):
💯++

src/binary.ts
```diff
@@ -277,24 +193,18 @@ export class Binary {
...
-    return buffer.toString('base64');
+    return this.buffer.toString('base64');
   }
 
   /** @internal */
   toString(format: BufferEncoding): string {
-    if (!this.buffer) return '';
-    const buffer = Buffer.from(this.buffer.slice(0, this.position) as Uint8Array);
-    return buffer.toString(format);
+    return this.buffer.slice(0, this.position).toString(format);
```
mbroadst, line 201: I know you didn't write this, but it's not clear why we want to slice the buffer. I think you can just have `return this.buffer.toString(format)`

src/binary.ts
```diff
@@ -312,14 +222,11 @@ export class Binary {
   }
 
   /** @internal */
-  static fromExtendedJSON(
-    doc: { $type: string; $binary: string | { subType: string; base64: string } },
-    options?: EJSONOptions
-  ): Binary {
+  static fromExtendedJSON(doc: BinaryEJSON, options?: EJSONOptions): Binary {
```
mbroadst, line 225: nit: should our "extended JSON" types just be suffixed/prefixed with `Extended`? Like `BinaryExtended`, `ExtendedUInt32`, etc? Just a thought.

src/binary.ts
```diff
@@ -312,14 +222,11 @@ export class Binary {
...
     options = options || {};
     let data: Buffer | undefined;
     let type;
-    if (options.legacy && typeof doc.$binary === 'string') {
+    if (options.legacy && '$type' in doc) {
```
mbroadst, line 229: :thinking: this doesn't seem equivalent

src/fnv1a.ts
```diff
@@ -1,5 +1,6 @@
 import { Buffer } from 'buffer';
 import { Long } from './long';
+import type { BufferEncoding } from './parser/utils';
```
mbroadst, line 3: This doesn't look used, was it included erroneously? If so, shouldn't we have a lint rule disallowing imports which are never used?

src/map.ts
```diff
@@ -4,6 +4,7 @@
 /* We do not want to have to include DOM types just for this check */
 declare const window: unknown;
 declare const self: unknown;
+declare const global: unknown;
```
mbroadst, line 7: this also looks like its added without a usage?

src/objectid.ts
```diff
@@ -1,6 +1,10 @@
 import { Buffer } from 'buffer';
-import { deprecate, inspect } from 'util';
-import { haveBuffer, randomBytes } from './parser/utils';
+import { ensureBuffer } from './ensure_buffer';
+import { BufferEncoding, haveBuffer, randomBytes } from './parser/utils';
+
+declare const require: Function;
```
mbroadst, line 5: is this needed if you disable `no-var-requires` below? It also occurs to me that `util` requires us to bring in node support for the web bundle, maybe we should consider replacing `deprecate` and only use `inspect` when used server-side

src/objectid.ts
```diff
@@ -47,9 +39,12 @@ export class ObjectId {
   /** @internal */
   static index = ~~(Math.random() * 0xffffff);
 
-  static cacheHexString?: boolean;
-  id: string | Buffer;
-  __id?: string;
+  static cacheHexString: boolean;
+
+  /** ObjectId Bytes @internal */
+  private _id: Buffer;
```
mbroadst, line 45: let's use a symbol here if we're going to take the approach of providing a getter/setter for the bytes?

src/objectid.ts
```diff
@@ -59,67 +54,77 @@ export class ObjectId {
...
+        if (bytes.byteLength === 12) {
+          this._id = bytes;
+        }
+      }
+      if (id.length === 24 && checkForHexRegExp.test(id)) {
+        this._id = Buffer.from(id, 'hex');
+      }
+    }
+
+    if (typeof this['_id'] === 'undefined') {
```
mbroadst, line 95: is this path reachable? `_id` is defined as non-optional

src/objectid.ts
```diff
@@ -227,11 +200,7 @@ export class ObjectId {
    */
   toString(format?: BufferEncoding): string {
     // Is the id a buffer then use the buffer toString method to return the format
-    if (this.id && typeof this.id !== 'string' && 'copy' in (this.id as Buffer)) {
-      return this.id.toString(typeof format === 'string' ? format : 'hex');
-    }
-
-    return this.toHexString();
+    return this.id.toString(typeof format === 'string' ? format : 'hex');
```
mbroadst, line 203: this change misses the hex string caching (not sure we care about that long term). I think for parity you want: 
```
if (format) return this.id.toString(format);
return this.toHexString();
```

src/objectid.ts
```diff
@@ -287,12 +256,7 @@ export class ObjectId {
   /** Returns the generation date (accurate up to the second) that this ID was generated. */
   getTimestamp(): Date {
     const timestamp = new Date();
-    let time: number;
-    if (typeof this.id !== 'string') {
-      time = this.id.readUInt32BE(0);
-    } else {
-      time = this.generationTime;
-    }
+    const time = this.id.readUInt32BE(0, false);
```
mbroadst, line 259: where are these `false` parameters coming from? I don't see them in the [node docs](https://nodejs.org/api/buffer.html#buffer_buf_readuint32be_offset)

src/parser/deserializer.ts
```diff
@@ -75,9 +75,7 @@ export function deserialize(
 
   if (size + index > buffer.length) {
     throw new Error(
-      `(bson size ${size} + options.index ${index} must be <= buffer length ${Buffer.byteLength(
-        buffer
-      )})`
+      `(bson size ${size} + options.index ${index} must be <= buffer length ${buffer.byteLength})`
```
mbroadst, line 78: I think we should continue to depend on [Buffer.byteLength](https://nodejs.org/api/buffer.html#buffer_static_method_buffer_bytelength_string_encoding), what if this is a `Buffer` instance? I'm sure there's a reason they provide the helper even though `Buffer` is implemented as a `UInt8Array`

src/parser/serializer.ts
```diff
@@ -368,7 +369,7 @@ function serializeBuffer(
   index = index + numberOfWrittenBytes;
   buffer[index++] = 0;
   // Get size of the buffer (current write point)
-  const size = value.length;
+  const size = value.byteLength;
```
mbroadst, line 372: was this just a bug? There is a difference between these two values

src/parser/utils.ts
```diff
@@ -1,3 +1,14 @@
+import { Buffer } from 'buffer';
+export type BufferEncoding =
```
mbroadst, line 2: is this really not provided by the node or `buffer` types?

src/parser/utils.ts
```diff
@@ -37,16 +49,14 @@ export function isUint8Array(value: unknown): value is Uint8Array {
   return Object.prototype.toString.call(value) === '[object Uint8Array]';
 }
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+declare const global: any;
+
 /** Call to check if your environment has `Buffer` */
 export function haveBuffer(): boolean {
```
mbroadst, line 56: do we actually need to check for `haveBuffer` anymore? Because we depend on `buffer` its always there, either polyfilled or not.

mbroadst (changes requested):

src/parser/utils.ts
nbbeeken: It's there but its not exported, maybe we should just accept string, it would mean not having to ever change this if there was some new format. (waiting on utf32 to take off 😆)

src/binary.ts
nbbeeken: I think it is because `this.position` represents the end of written data, without it the toString will return the whole buffer, which can be larger than the data we are concerned about.

src/fnv1a.ts
nbbeeken: It is used in both functions in this file..? 😅 are we seeing something different 
I removed this type anyway.

src/map.ts
nbbeeken: It's used in the `getGlobal` function, I'm worried GH is messing something up.

src/objectid.ts
nbbeeken: This is needed because I removed node types so this just a little patch to allow use to use require in the "is this node" guarded code. 

NODE-2826

src/objectid.ts
nbbeeken: Oops legacy node pre 1.0 arguments, removed.

src/parser/deserializer.ts
nbbeeken: > When string is a Buffer/DataView/TypedArray/ArrayBuffer/ SharedArrayBuffer, the byte length as reported by .byteLength is returned.

So this is just grabbing the same property. I changed this because the typings for this function only accept string even though it takes a variety of types 🤷  could cast to fix?

src/parser/serializer.ts
nbbeeken: Now that any TypedArrays land in this path of serialization it's important to use the byteLength. `Float32Array([1.2, 3.4])` has length 2 but byteLength 8. If it were only int8 types arriving here then length and  byteLength would be the same.

src/parser/utils.ts
nbbeeken: Whoosh 🌬️ and just like that its gone, good catch!

src/binary.ts
nbbeeken: We have "NODE-2723 Add type information for EJSON" so we can figure that out there.
I do think `BinaryExtended` or `ExtendedBinary` could be confusing _maybe? maybe not_ because if your in the domain of EJSON then its obvious, but it does seem to imply its an extension of the Binary class 🤔 def could come up with something better than tacking on `EJSON`

src/objectid.ts
nbbeeken: This was checking if the string length/format checks failed, I made the control flow clearer.

src/binary.ts
```diff
@@ -312,14 +222,11 @@ export class Binary {
...
-  static fromExtendedJSON(
-    doc: { $type: string; $binary: string | { subType: string; base64: string } },
-    options?: EJSONOptions
-  ): Binary {
+  static fromExtendedJSON(doc: BinaryEJSON, options?: EJSONOptions): Binary {
     options = options || {};
     let data: Buffer | undefined;
     let type;
-    if (options.legacy && typeof doc.$binary === 'string') {
+    if (options.legacy && '$type' in doc && typeof doc.$binary === 'string') {
```
mbroadst, line 229: I think this is still wrong, the original conditional is checking only for `doc.$binary`, the `doc.$type` is optional and defaults to 0 if its not present. Looks like we need a test for this if it's not triggering any breakage

src/fnv1a.ts
mbroadst: yeah I dunno, the GitHub diff was just adding this and no other changes :shrug:

src/binary.ts
nbbeeken: Aha! It was an issue with the interface, should be fixed now.

src/binary.ts
```diff
@@ -37,51 +44,45 @@ export class Binary {
...
-      this.position = buffer.length;
+    this.sub_type = subType ?? Binary.BSON_BINARY_SUBTYPE_DEFAULT;
+
+    if (buffer === undefined || buffer === null) {
+      // create an empty binary buffer
+      this.buffer = Buffer.alloc(Binary.BUFFER_SIZE);
+      this.position = 0;
+    } else if (typeof buffer === 'string') {
+      // string
+      this.buffer = writeStringToArray(buffer);
```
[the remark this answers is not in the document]
nbbeeken: I don't think there's anything that could be made more simple here, there's a slight difference to each case. 

Somewhat related the ensureBuffer function is used at the top level deserialize functions so I've updated those types:

- [bson.ts:202](https://github.com/mongodb/js-bson/blob/6f678cddad4fd01f1f8b75967b9d897785c8042b/src/bson.ts#L202)
- [bson.ts:245](https://github.com/mongodb/js-bson/blob/6f678cddad4fd01f1f8b75967b9d897785c8042b/src/bson.ts#L245)

src/parser/deserializer.ts
mbroadst: oh I see, I missed that the method only accepts a string with an encoding. Move along!

src/parser/serializer.ts
mbroadst: > Now that any TypedArrays land in this path

Just in terms of typing right? TypedArrays have always been able to take this path. Does this mean we've been serializing them incorrectly all along? Might be a case for making more of a breaking change if it never worked correctly in the first place

nbbeeken: There was a modification in the serialization that led to this (because the type annotations followed the change it was having me change this to byteLength) this has be fixed

mbroadst (approved):
:tada: :rocket:

PR #396, opened 2020-09-16
NODE-2770/rm-crc32

reggi:
[NODE-2770](https://jira.mongodb.org/browse/NODE-2770)

PR to remove the `cacheFunctionsCrc32` option from the deserialize function. 

I've checked in a couple different prior tags / branches to see if there was ever an implementation of `crc32`, I only ever saw it set to `null`.

```js
const hash = cacheFunctionsCrc32 && crc32 ? crc32(functionString) : functionString;
```

With it set to null the hashing function would never be called an the hash would only ever be set to the `functionString`.

This deprecates the `crc32` hashing option as, it was never used.

### DIFF
--- a/src/parser/deserializer.ts
+++ b/src/parser/deserializer.ts
@@ -489,7 +489,7 @@ function deserializeObject(
         // If we have cache enabled let's look for the md5 of the function in the cache
         if (cacheFunctions) {
           // Got to do this to avoid V8 deoptimizing the call due to finding eval
-          object[name] = isolateEvalWithHash(functionCache, functionString, object);
+          object[name] = isolateEval(functionString, functionCache, object);
         } else {
           object[name] = isolateEval(functionString);
         }
@@ -557,7 +557,7 @@ function deserializeObject(
         // If we have cache enabled let's look for the md5 of the function in the cache
         if (cacheFunctions) {
           // Got to do this to avoid V8 deoptimizing the call due to finding eval
-          object[name] = isolateEvalWithHash(functionCache, functionString, object);
+          object[name] = isolateEval(functionString, functionCache, object);
         } else {
           object[name] = isolateEval(functionString);
         }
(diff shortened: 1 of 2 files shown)


src/parser/deserializer.ts
```diff
@@ -21,8 +21,6 @@ export interface DeserializationOptions {
   evalFunctions?: boolean;
   /** cache evaluated functions for reuse. */
   cacheFunctions?: boolean;
-  /** use a crc32 code for caching, otherwise use the string of the function. */
-  cacheFunctionsCrc32?: boolean;
```
mbroadst, line 25: Unfortunately, this is a breaking change so we can't make it. I'm not against the idea of removing crc32, but at very least we need to properly deprecate it and let users know to stop using the option

mbroadst (changes requested):

src/parser/deserializer.ts
```diff
@@ -656,17 +643,16 @@ function deserializeObject(
  */
 function isolateEvalWithHash(
```
nbbeeken, line 644: I think its worth renaming this, maybe event merging it with the other one and adding a flag to indicate whether to cache or not.

src/parser/deserializer.ts
nbbeeken: I think it s a bug fix for what its worth, I'll let Thomas fill in the details but I believe he found there was never a crc32 function in the repo. So maybe we could deprecate this flag but leave that it does nothing, as opposed to adding a new dep. 
As an aside I was wondering what's the advantage of making a hash out of the function string as opposed to using the function string itself as the cache key?

nbbeeken (changes requested):

src/parser/deserializer.ts
reggi: Hey I consolidated the functions here.

src/parser/deserializer.ts
reggi: I updated the description :)

mbroadst: Yeah I totally believe it wasn't there, but we've documented it so we should probably warn users that the option isn't really doing anything. Again, totally fine removing any logic related to crc32.

The benefit of using that would be to speed up lookups in tight loops by providing a faster lookup for a smaller string, and reducing the allocation (in particular if the function is _huge_). Interestingly enough we already embed a FNV1a hash implementation (used to be used for objectid) that would probably be just as effective here. Either way, I think it's premature optimization.

reggi: @mbroadst I just put the property back in the interface with the deprecated tag, and I removed it from the docs.

mbroadst (approved):

nbbeeken (approved):
🚀

PR #397, opened 2020-09-16
fix: Throw on BigInt type values

nbbeeken:
Given JavaScript's new BigInt type the serializer will attempt to fit the value in the smallest possible type. In the following order: Int32, Int64, Decimal128. Any values larger than Decimal128's maximum representable range will result in a thrown TypeError

NODE-2529

There's no logic in here to deserialize to big int only to accept it and serialize it to one of our BSON types. In terms of platform support I only used the BigInt constructor not the `23n` syntax so the code path will just go untravelled on older JS engines.

### DIFF
--- a/src/long.ts
+++ b/src/long.ts
@@ -1,6 +1,6 @@
-import type { Timestamp } from './timestamp';
 import type { EJSONOptions } from './extended_json';
 import { isObjectLike } from './parser/utils';
+import type { Timestamp } from './timestamp';
 
 interface LongWASMHelpers {
   /** Gets the high bits of the last operation performed */
@@ -194,7 +194,7 @@ export class Long {
    * @returns The corresponding Long value
    */
   static fromBigInt(value: bigint, unsigned?: boolean): Long {
-    return Long.fromString(value.toString(10), unsigned);
+    return Long.fromString(value.toString(), unsigned);
   }
 
   /**
--- a/src/long.ts
+++ b/src/long.ts
@@ -188,6 +188,16 @@ export class Long {
   }
 
   /**
+   * Returns a Long representing the given value, provided that it is a finite number. Otherwise, zero is returned.
+   * @param value - The number in question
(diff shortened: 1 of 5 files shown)


reggi (approved):

src/long.ts
```diff
@@ -186,6 +186,16 @@ export class Long {
...
 
+  /**
+   * Returns a Long representing the given value, provided that it is a finite number. Otherwise, zero is returned.
+   * @param value - The number in question
+   * @param unsigned - Whether unsigned or not, defaults to signed
+   * @returns The corresponding Long value
+   */
+  static fromBigInt(value: bigint, unsigned?: boolean): Long {
+    return Long.fromString(value.toString(10), unsigned);
+  }
```
nbbeeken, line 197: While we leave the serializer in a state where a translation must be made manually I figure it would still be good to provide these `to` and `from` helpers, thoughts?

mbroadst (changes requested):

src/parser/serializer.ts
```diff
@@ -806,6 +806,8 @@ export function serializeInto(
         index = serializeString(buffer, key, value, index, true);
       } else if (typeof value === 'number') {
         index = serializeNumber(buffer, key, value, index, true);
+      } else if (typeof value === 'bigint') {
+        throw new TypeError('Do not know how to serialize a BigInt, please use Decimal128');
```
mbroadst, line 810:
```suggestion
        throw new TypeError('Unsupported type BigInt, please use Decimal128');
```

this is just a deep nitpick, I think we do know how to serialize them, we just don't support them being passed in raw for now

mbroadst (approved):
:onion: on wording, feel free to ignore and merge, but I felt compelled to leave a note

src/parser/serializer.ts
nbbeeken: It just matches JSON's error precisely
But I think I agree and I'll change this

emadum (approved):

PR #399, opened 2020-09-17
refactor: Removes dead code paths from number serialization

nbbeeken:
Numbers that exceed the representable size of int32 will be serialized as doubles to keep consistency with JS VM behavior

NODE-2769


reggi (approved):

mbroadst (approved):
LGTM

emadum (approved):

PR #401, opened 2020-09-17
NODE-2723/ejson-interfaces

reggi:
[NODE-2723](https://jira.mongodb.org/browse/NODE-2723)

This PR breaks out interface types for the [Canonical Extended JSON Format](https://github.com/mongodb/specifications/blob/master/source/extended-json.rst#conversion-table). It uses the convention `<Type>Extended` where `<Type>` matches the class reference, so `DoubleExtended` or `BSONRegExpExtended` where chosen respectively. This also keeps the `toExtendedJSON` return values as unions and does not aim to create a singular representation of the return type for that function. This splits out legacy interfaces using the convention `<Type>ExtendedLegacy` where applicable.


src/binary.ts
```diff
@@ -313,13 +313,13 @@ export class Binary {
 
   /** @internal */
   static fromExtendedJSON(
-    doc: { $type: string; $binary: string | { subType: string; base64: string } },
```
reggi, line 316: This seemed to be a mistake here, the `{ subType: string; base64: string }` object wasn't  wrapped in `$binary`. 🌯

nbbeeken (approved):
🚀

mbroadst (approved):

emadum (approved):

PR #402, opened 2020-09-18
refactor: Remove util dependency

nbbeeken:
Removes util dependency and restricts rollup bundler to only include Buffer

NODE-2826

### DIFF
--- a/src/objectid.ts
+++ b/src/objectid.ts
@@ -362,12 +362,14 @@ Object.defineProperty(ObjectId, 'get_inc', {
   value: deprecate(() => ObjectId.getInc(), 'Please use the static `ObjectId.getInc()` instead')
 });
 
+const inspect = Symbol.for('nodejs.util.inspect.custom');
 /**
  * Converts to a string representation of this Id.
  *
  * @returns return the 24 character hex string representation.
  * @internal
  */
+Object.defineProperty(ObjectId.prototype, inspect, ObjectId.prototype.toString);
 Object.defineProperty(ObjectId.prototype, 'inspect', ObjectId.prototype.toString);
 
 Object.defineProperty(ObjectId.prototype, '_bsontype', { value: 'ObjectID' });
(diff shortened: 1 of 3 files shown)


emadum (approved):
LGTM

src/parser/utils.ts
```diff
@@ -60,3 +61,19 @@ export function isDate(d: unknown): d is Date {
 export function isObjectLike(candidate: unknown): candidate is Record<string, unknown> {
   return typeof candidate === 'object' && candidate !== null;
 }
+
+export function deprecate<T extends Function>(fn: T, message: string): T {
+  if (typeof window === 'undefined' || typeof self === 'undefined') {
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    return require('util').deprecate(fn, message);
```
mbroadst, line 68: Is there still any benefit to using the Node built-in at this point? I like this runtime agnostic one

mbroadst (changes requested):

reggi (approved):

src/objectid.ts
```diff
@@ -422,6 +421,6 @@ Object.defineProperty(ObjectId, 'get_inc', {
  * @returns return the 24 character hex string representation.
  * @internal
  */
-Object.defineProperty(ObjectId.prototype, inspect.custom || 'inspect', ObjectId.prototype.toString);
+Object.defineProperty(ObjectId.prototype, 'inspect', ObjectId.prototype.toString);
```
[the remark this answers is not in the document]
nbbeeken: Good idea, the inspect symbol will always exist here, Symbol.for creates a new symbol if one doesn't exist. So I'll drop the if statement.

src/parser/utils.ts
nbbeeken: If we're in nodejs this will handle the `--throw-deprecation`, `--no-deprecation`, `--trace-deprecation` flags.

mbroadst (approved):
LGTM

PR #403, opened 2020-09-23
feat: add extended json parsing for $uuid

reggi:
[NODE-2805](https://jira.mongodb.org/browse/NODE-2805?filter=-1)

* Adds in extended JSON parsing for `$uuid`.
* Had a learning curve to understand how UUID's are converted to a binary string for `base64` encoding, until I realized it was a specialized for UUID's.
* Added a couple of UUID helper methods `UUIDPattern`, `parseUUID`, `formatUUID` to a new `uuid.ts` file, some pulled in from the [`uuid` module](https://www.npmjs.com/package/uuid). With the intention on adding the dependency at a later time if necessary.
* Seeking advice on where to put these methods new helper methods (and naming), as we may want to reserve the `uuid.ts` file for the `UUID` class in the future.
* Alternatively could create `UUID` class now and add static `parse` and `format` methods, swappable if we ever use the `uuid` module,  the `UUIDExtended` interface could also live in `uuid.ts` rather then`binary.ts`
* <s>Includes a `formatUUID` function not provided by the UUID package, in order to format the [UUID in accordance with python's `uuid.UUID` method](https://github.com/mongodb/specifications/blob/master/source/extended-json.rst#user-content-special-rules-for-parsing-uuid-fields:~:text=Parsers%20MUST%20interpret%20the%20%24uuid%20key,between%20hex%20character%20groups%20(e.g.%20c8edabc3f7384ca3b68dab92a91478a3).), where the `urn:uuid ` prefix is stripped and `-` are added in if they are not there prior to `base64` encoding.</s>

### DIFF
--- a/src/binary.ts
+++ b/src/binary.ts
@@ -1,7 +1,7 @@
 import { Buffer } from 'buffer';
 import { ensureBuffer } from './ensure_buffer';
 import type { EJSONOptions } from './extended_json';
-import { parseUUID } from './uuid';
+import { parseUUID, UUIDExtended } from './uuid';
 
 type BinarySequence = Uint8Array | Buffer | number[];
 
@@ -15,10 +15,6 @@ export interface BinaryExtended {
     subType: string;
     base64: string;
   };
-}
-
-export interface UUIDExtended {
-  $uuid: string;
 }
 
 /** A class representation of the BSON Binary type. */
--- a/src/binary.ts
+++ b/src/binary.ts
@@ -1,6 +1,7 @@
 import { Buffer } from 'buffer';
 import { ensureBuffer } from './ensure_buffer';
 import type { EJSONOptions } from './extended_json';
+import { parseUUID } from './uuid';
 
 type BinarySequence = Uint8Array | Buffer | number[];
 
@@ -14,6 +15,10 @@ export interface BinaryExtended {
     subType: string;
     base64: string;
   };
+}
+
+export interface UUIDExtended {
+  $uuid: string;
 }
 
 /** A class representation of the BSON Binary type. */
@@ -202,7 +207,7 @@ export class Binary {
   }
 
   /** @internal */
-  toExtendedJSON(options?: EJSONOptions): BinaryExtendedLegacy | BinaryExtended {
+  toExtendedJSON(options?: EJSONOptions): BinaryExtendedLegacy | BinaryExtended | UUIDExtended {
     options = options || {};
     const base64String = this.buffer.toString('base64');
 
@@ -223,20 +228,25 @@ export class Binary {
 
   /** @internal */
   static fromExtendedJSON(
-    doc: BinaryExtendedLegacy | BinaryExtended,
+    doc: BinaryExtendedLegacy | BinaryExtended | UUIDExtended,
     options?: EJSONOptions
   ): Binary {
     options = options || {};
     let data: Buffer | undefined;
     let type;
-    if (options.legacy && typeof doc.$binary === 'string' && '$type' in doc) {
-      type = doc.$type ? parseInt(doc.$type, 16) : 0;
-      data = Buffer.from(doc.$binary, 'base64');
-    } else {
-      if (typeof doc.$binary !== 'string') {
-        type = doc.$binary.subType ? parseInt(doc.$binary.subType, 16) : 0;
-        data = Buffer.from(doc.$binary.base64, 'base64');
+    if ('$binary' in doc) {
+      if (options.legacy && typeof doc.$binary === 'string' && '$type' in doc) {
+        type = doc.$type ? parseInt(doc.$type, 16) : 0;
+        data = Buffer.from(doc.$binary, 'base64');
+      } else {
+        if (typeof doc.$binary !== 'string') {
+          type = doc.$binary.subType ? parseInt(doc.$binary.subType, 16) : 0;
+          data = Buffer.from(doc.$binary.base64, 'base64');
+        }
       }
+    } else if ('$uuid' in doc) {
+      type = 4;
+      data = Buffer.from(parseUUID(doc.$uuid));
     }
     if (!data) {
       throw new TypeError(`Unexpected Binary Extended JSON format ${JSON.stringify(doc)}`);
--- a/src/binary.ts
+++ b/src/binary.ts
@@ -207,7 +207,7 @@ export class Binary {
   }
 
   /** @internal */
-  toExtendedJSON(options?: EJSONOptions): BinaryExtendedLegacy | BinaryExtended | UUIDExtended {
+  toExtendedJSON(options?: EJSONOptions): BinaryExtendedLegacy | BinaryExtended {
     options = options || {};
     const base64String = this.buffer.toString('base64');
 
--- a/src/uuid.ts
+++ b/src/uuid.ts
@@ -1,3 +1,7 @@
+/**
+ * UUID regular expression pattern copied from `uuid` npm module.
+ * @see https://github.com/uuidjs/uuid/blob/master/src/regex.js
+ */
 const UUID_RX = /^(?:[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}|00000000-0000-0000-0000-000000000000)$/i;
 
 export interface UUIDExtended {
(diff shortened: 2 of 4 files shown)


src/uuid.ts
```diff
@@ -0,0 +1,51 @@
+const UUIDPattern = /^(?:[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}|00000000-0000-0000-0000-000000000000)$/i;
+
+export function formatUUID(uuid: string): string {
+  uuid = uuid.replace(/^.+:/, '');
+  const result = uuid.match(/^(.{8})(.{4})(.{4})(.{4})(.{12})$/);
+  if (result) uuid = result.slice(1, 6).join('-');
+  return uuid;
+}
+
+export function parseUUID(uuid: string): Uint8Array {
```
mbroadst, line 10: was this copied from somewhere? please leave attribution if so

src/uuid.ts
```diff
@@ -0,0 +1,51 @@
...
+  uuid = formatUUID(uuid);
+
+  if (typeof uuid !== 'string') {
+    throw new Error('$uuid wrong type');
+  }
+  if (!UUIDPattern.test(uuid)) {
```
mbroadst, line 16: two comments:
- generally we tend to name file global constants IN_UPPER_CASE, this one might be called `UUID_RX` or something?
- it seems weird to validate using a regular expression _after_ you format the value in L11 above, is that intentional?

src/uuid.ts
```diff
@@ -0,0 +1,51 @@
+const UUIDPattern = /^(?:[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}|00000000-0000-0000-0000-000000000000)$/i;
+
+export function formatUUID(uuid: string): string {
```
mbroadst, line 3: should the `formatUUID` and `parseUUID` methods actually be exported? I think we would at least want to mark them `@internal` for now.

mbroadst (changes requested):

src/uuid.ts
reggi: Can fix uppercasing. 

Yes it is intentional to format -> validate. The validator accepts "proper" uuid's and will throw an error if it's invalid. The formatter is a buffer to convert SPEC-compliant uuids, to transform them before an error is thrown.

src/uuid.ts
reggi: just `parseUUID` does 👍

src/uuid.ts
mbroadst: it's still unclear to me from reading the source why the UUID needs to be formatted before its parsed, can you explain here? Or even better, explain in the source so that future readers can gain from what you learned.

src/uuid.ts
```diff
@@ -0,0 +1,57 @@
+const UUID_RX = /^(?:[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}|00000000-0000-0000-0000-000000000000)$/i;
+
+/** @internal */
+function formatUUID(uuid: string): string {
+  uuid = uuid.replace(/^.+:/, '');
```
nbbeeken, line 5: I think this is unnecessary
This is funky wording from the spec:
```
Parsers MUST accept textual representations of UUIDs that omit the URN prefix (usually urn:uuid:)
```
"MUST accept UUIDs that **_omit_** the prefix": It's like a positive/negative but I read it as the parser should not accept uuids with such a prefix.

src/binary.ts
```diff
@@ -16,6 +17,10 @@ export interface BinaryExtended {
   };
 }
 
+export interface UUIDExtended {
+  $uuid: string;
+}
```
nbbeeken, line 22: Might as well put this in the uuid.ts file, since we're encapsulating stuff in that module anyway. Super minor, up to you.

nbbeeken (changes requested):
nice! one discussion worthy topic on the formatter function

src/uuid.ts
reggi: In the description of this ticket:
> Includes a `formatUUID` function not provided by the UUID package, in order to format the [UUID in accordance with python's `uuid.UUID` method](https://github.com/mongodb/specifications/blob/master/source/extended-json.rst#user-content-special-rules-for-parsing-uuid-fields:~:text=Parsers%20MUST%20interpret%20the%20%24uuid%20key,between%20hex%20character%20groups%20(e.g.%20c8edabc3f7384ca3b68dab92a91478a3).), where the `urn:uuid ` prefix is stripped and `-` are added in if they are not there prior to `base64` encoding.

src/uuid.ts
reggi: Originally I was trying to match python.

```python
>>> import uuid, base64
>>> print(base64.b64encode(uuid.UUID('c8edabc3-f738-4ca3-b68d-ab92a91478a3').bytes))
yO2rw/c4TKO2jauSqRR4ow==
>>> print(base64.b64encode(uuid.UUID('urn:uuid:c8edabc3-f738-4ca3-b68d-ab92a91478a3').bytes))
yO2rw/c4TKO2jauSqRR4ow==
>>> print(base64.b64encode(uuid.UUID('c8edabc3f7384ca3b68dab92a91478a3').bytes))
yO2rw/c4TKO2jauSqRR4ow==
```

The requirements are:

```
urn:uuid:9b1deb4d-3b7d-4bad-9bdd-2b0d7b3dcb6d MAY return mx3rTTt9S62b3SsNez3LbQ==
9b1deb4d-3b7d-4bad-9bdd-2b0d7b3dcb6d MUST return mx3rTTt9S62b3SsNez3LbQ==
9b1deb4d3b7d4bad9bdd2b0d7b3dcb6d MAY return mx3rTTt9S62b3SsNez3LbQ==
```

It was my decision to play it safe by supporting this, I will remove the `format` function I didn't think it would raise that may eyebrows. Now it just seems like extra work, that is optional.

nbbeeken: Yea my initial reaction would've been the same. I guess when dealing with computer data formats (even though json is mostly readable) erring on the side of stricter feels more correct for the circumstance.

nbbeeken (approved):
🚀

src/uuid.ts
reggi: I removed the format function -> https://github.com/mongodb/js-bson/pull/403#discussion_r496909847

src/uuid.ts
```diff
@@ -0,0 +1,52 @@
...
+ * Parser function copied from `uuid` npm module.
+ * @see https://github.com/uuidjs/uuid/blob/master/src/parse.js
+ * @internal
+ */
+export function parseUUID(uuid: string): Uint8Array {
+  if (typeof uuid !== 'string') {
+    throw new Error('$uuid wrong type');
+  }
+
+  const UUID_RX = /^(?:[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}|00000000-0000-0000-0000-000000000000)$/i;
```
mbroadst, line 15: Let's pull this out so you're not declaring it every time a UUID is parsed. Likely v8 will optimize this for us, but I think it's best practice to try to minimize the allocation of expensive types.

src/uuid.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  if (!UUID_RX.test(uuid)) {
+    throw new Error('$uuid invalid value');
```
mbroadst, line 18:
```suggestion
    throw new TypeError('Invalid format for UUID: ' + uuid);
```

src/uuid.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+/**
+ * Parser function copied from `uuid` npm module.
+ * @see https://github.com/uuidjs/uuid/blob/master/src/parse.js
+ * @internal
+ */
+export function parseUUID(uuid: string): Uint8Array {
+  if (typeof uuid !== 'string') {
+    throw new Error('$uuid wrong type');
```
mbroadst, line 12:
```suggestion
    throw new TypeError('Invalid type for UUID, expected string but got ' + typeof uuid);
```

mbroadst (changes requested):
LGTM, but I think we can improve our error messages for users

mbroadst (approved):
LGTM

PR #404, opened 2020-09-28
fix: Rework rollup config to output named exports (NODE-2712)

nbbeeken:
Rollup's output was bundling the library to have one single default export this change makes the output preserve the named outputs as well as have the default export. In addition the rollup config now compiles from typescript source instead of the typescript output.

NODE-2712


src/bson.ts
```diff
@@ -277,3 +278,68 @@ export function deserializeStream(
...
+  serializeWithBufferAndIndex,
+  deserialize,
+  calculateObjectSize,
+  deserializeStream,
+  setInternalBufferSize,
+  // EJSON namespace
+  EJSON
+};
+
+export default bsonDefaultModuleExport;
```
mbroadst, line 345: Hmm, this looks pretty funky. Are we mixing named and default exports? It looks like you specified `named` in the rollup config, do you still need this here?

src/objectid.ts
```diff
@@ -60,7 +60,7 @@ export class ObjectId {
       this.__id = id.__id;
     }
 
-    if (typeof id === 'object' && 'id' in id) {
+    if (typeof id === 'object' && id && 'id' in id) {
```
mbroadst, line 63: why is this change needed? the `typeof` check should prove the existence of the object (it's not `typeof === undefined`)

rollup.config.js
```diff
@@ -1,45 +1,72 @@
...
+        name: 'BSON',
+        sourcemap: true
       },
-      /* Browser IIFE bundle */
-      {
+      /* Browser IIFE bundle */ {
         file: 'dist/bson.bundle.js',
-        format: 'iife',
-        name: 'BSON'
+        format: 'esm',
```
mbroadst, line 50:
```suggestion
        format: 'iife',
```

rollup.config.js
```diff
@@ -1,45 +1,72 @@
...
-      {
-        file: 'dist/bson.browser.esm.js',
-        format: 'es'
-      },
       /* Browser UMD bundle */
       {
         file: 'dist/bson.browser.umd.js',
-        format: 'umd',
-        name: 'BSON'
+        format: 'esm',
```
mbroadst, line 44:
```suggestion
        format: 'umd',
```

rollup.config.js
```diff
@@ -1,45 +1,72 @@
...
+import typescript from '@rollup/plugin-typescript';
 
 // Entry point of the library
-const input = './lib/bson.js';
+const input = './src/bson.ts';
 
 export default [
+  {
+    input,
+    /* Browser ESM bundle */
```
mbroadst, line 13: can you explain what's going on in this change? Also, it looks like the config is mostly duplicated below

rollup.config.js
```diff
@@ -49,13 +76,21 @@ export default [
...
+        sourcemap: true
+      }
+    ],
     // Notice the external buffer, and preferBuiltins false,
     // this bundle doesn't pull in 'buffer' or Map from 'core-js'
     external: ['buffer'],
-    plugins: [nodeResolve({ preferBuiltins: true }), commonjs()]
+    plugins: [
+      typescript({ tsconfig: './tsconfig.rollup.json' }),
+      nodeResolve({ preferBuiltins: true }),
```
mbroadst, line 92: I just noticed the comment above saying "note preferBuiltins false" which not actually the case here

tsconfig.rollup.json
```diff
@@ -0,0 +1,32 @@
+{
```
mbroadst, line 1: does it make sense to embed this config directly in the `rollup.config.js`? I feel like otherwise it's just another file added to the heap of tooling files laying around. Is there a benefit to having this in its own file that I'm missing?

package.json
```diff
@@ -32,6 +32,7 @@
     "@rollup/plugin-commonjs": "^15.0.0",
     "@rollup/plugin-json": "^4.1.0",
     "@rollup/plugin-node-resolve": "^9.0.0",
+    "@rollup/plugin-typescript": "^6.0.0",
```
mbroadst, line 35: was this change necessary for this fix?

mbroadst (changes requested):

package.json
nbbeeken: Yes, it fixed errors with the exports being incorrectly formatted since rollup was working off of the TS output it was translating from commonjs back to esm and this just made it easier to reason about the pipeline.

tsconfig.rollup.json
nbbeeken: Yep sounds good!

rollup.config.js
nbbeeken: Oops yea out of date comment, good catch.

rollup.config.js
nbbeeken: Yea the key difference is the babel settings. There is definitely a lot of duplication but after wrestling with the config for a better part of this project an explicit set of settings for each output is actually quite helpful rather than debugging a programatic config.

rollup.config.js
nbbeeken: Same as comment above. This is correct for what we're doing here.

rollup.config.js
nbbeeken: The format here is correct, babel is downgrading the code after rollup has done its transformations so babel is handling the bundle format. I use UMD in the babel preset because it is equivalent to an iife (babel doesn't have an iife output without a plugin, which seems unnecessary).

src/objectid.ts
nbbeeken: `typeof null` is 'object', so this is checking for that

src/bson.ts
nbbeeken: Yes, because our original bundles had both named and default exports so we need to keep the behavior.

reggi (approved):
LGTM, babel plugins to the rescue?

mbroadst (approved):
LGTM!

emadum (approved):

PR #405, opened 2020-09-30
docs: Add release tags and export symbols from entry point

nbbeeken:
Also re-enable api-extractor flags to lint our tsdoc blocks.

NODE-2837

### DIFF
--- a/src/extended_json.ts
+++ b/src/extended_json.ts
@@ -14,7 +14,11 @@ import { BSONRegExp } from './regexp';
 import { BSONSymbol } from './symbol';
 import { Timestamp } from './timestamp';
 
-export type BSONType =
+/** @public */
+export type EJSONOptions = EJSON.Options;
+
+/** @internal */
+type BSONType =
   | Binary
   | Code
   | DBRef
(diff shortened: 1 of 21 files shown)


src/extended_json.ts
```diff
@@ -380,3 +275,122 @@ function serializeDocument(doc: any, options: EJSONOptions) {
...
 }
+
+/**
+ * EJSON parse / stringify API
+ * @public
+ */
+// the namespace here is used to emulate `export * as EJSON from '...'`
+// which as of now (sept 2020) api-extractor does not support
+// eslint-disable-next-line @typescript-eslint/no-namespace
+export namespace EJSON {
```
nbbeeken, line 286: As the comment states, the ideal code would be to use `export * as ESJON from './...'` in bson.ts but unfortunately that syntax is not supported by api-extractor yet, so this is an alternative that could be changed later without breaking changes.
Also unlike our previous solution (a `const EJSON = {...}` in bson.ts) this will thread the documentation through to the user's import, before it was being lost.

reggi (approved):
LGTM, API EXTRACTOR 🙀✅

emadum (approved):
LGTM 👍

PR #406, opened 2020-10-01
refactor: Use Math.floor to round down

mvasilkov:
This commit replaces the `~~` operation with `Math.floor`, which is more consistent across the js-bson codebase, and arguably more readable too.

### DIFF
--- a/src/objectid.ts
+++ b/src/objectid.ts
@@ -39,7 +39,7 @@ export class ObjectId {
   _bsontype!: 'ObjectId';
 
   /** @internal */
-  static index = ~~(Math.random() * 0xffffff);
+  static index = Math.floor(Math.random() * 0xffffff);
 
   static cacheHexString: boolean;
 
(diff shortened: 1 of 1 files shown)


Uzlopak:
`~~` is faster than `Math.floor`. Also should be mentioned that the equivalent of `~~` is `Math.trunc` and not `Math.floor`, as `Math.floor` rounds down but `Math.trunc` just cuts off the decimal places. Also should be noticed that `~~` is only an issue if you get numbers bigger than 2^32. I dont see any issue here, as 1 * 0xffffff = 16777215, which is 2^24.

I really dont understand, why people argue about code readability, when performance should be the primary goal in such a heavy used package, and even more when it is the objectid- Its not even syntactical sugar, but actually a real operator - like bit-operators - in javascript, which has performance impacts.

But anyhow... I see that master already contains that change. So this PR should be closed.

nbbeeken:
Indeed this has already been changed in another commit, thanks for the submission! We really appreciate the help!

PR #411, opened 2020-10-29
fix: Fix backwards compatibility with older BSON package versions

nbbeeken:
The isBuffer check in the buffer module was too strict and led to a failure to deal with older BSON module Binary types. There is now an improved more relaxed method.

NODE-2848

I added testing for this, we can easily expand it to cover more types/cases with older versions of the library.

### DIFF
--- a/test/bson_older_versions_tests.js
+++ b/test/bson_older_versions_tests.js
@@ -17,7 +17,11 @@ function downloadZip(version, done) {
     })
     .then(r => {
       fs.writeFileSync(`bson-${version}.zip`, new Uint8Array(r));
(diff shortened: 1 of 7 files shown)


test/bson_older_versions_tests.js
```diff
@@ -0,0 +1,74 @@
+'use strict';
```
emadum, line 1: I think a comment block somewhere at the top of this file would be useful to future maintainers, explaining what it is doing and why it's needed.

emadum (approved):
LGTM, just one comment request 👍

PR #412, opened 2020-11-05
fix: make inspect for ObjectId work

addaleax:
The previous `Object.defineProperty()` calls did not work because
they did not specify a property descriptor, and even if they had
worked, the signatures of `.inspect()` and `.toString()` would
have mismatched, and even if they weren’t, the result would not
have been very useful as debugging output, because it only returned
the raw hex content of the `ObjectId` and not even e.g. the fact
that this object is an `ObjectId`.

I’m happy with any other solution, but this particular one would
make my life a tiny bit easier :)

## Description

Make the `inspect` and `[util.inspect.custom]` methods on `ObjectId` work.

---

Not sure if I should open a JIRA ticket for this, but happy to do so if it helps :)

PR #413, opened 2020-11-16
chore: expand inspect support of BSON types NODE-2875

rose-m:
## Description

Adds support for the NodeJS `inspect` function to more BSON data types. This is an upstream change proposed by the mongosh team.

A custom `inspect` function is added for the following data types:
* `Binary`
* `Code`
* `DBRef`
* `Decimal128`
* `Int32`
* `Long`
* `MinKey`
* `MaxKey`
* `BSONSymbol`
* `Timestamp`

### DIFF
--- a/src/binary.ts
+++ b/src/binary.ts
@@ -255,6 +255,16 @@ export class Binary {
     }
     return new Binary(data, type);
   }
+
+  /** @internal */
+  [Symbol.for('nodejs.util.inspect.custom')](): string {
+    return this.inspect();
+  }
+
+  inspect(): string {
+    const asBuffer = this.value(true);
+    return `Binary("${asBuffer.toString('hex')}", ${this.sub_type})`;
+  }
 }
 
 Object.defineProperty(Binary.prototype, '_bsontype', { value: 'Binary' });
--- a/src/code.ts
+++ b/src/code.ts
@@ -42,6 +42,16 @@ export class Code {
   static fromExtendedJSON(doc: CodeExtended): Code {
     return new Code(doc.$code, doc.$scope);
   }
+
+  /** @internal */
+  [Symbol.for('nodejs.util.inspect.custom')](): string {
+    return this.inspect();
+  }
+
+  inspect(): string {
+    const codeJson = this.toJSON();
+    return `Code("${codeJson.code}"${codeJson.scope ? `, ${JSON.stringify(codeJson.scope)}` : ''})`;
+  }
 }
 
 Object.defineProperty(Code.prototype, '_bsontype', { value: 'Code' });
--- a/test/node/bson_test.js
+++ b/test/node/bson_test.js
@@ -2274,112 +2274,92 @@ describe('BSON', function () {
     /**
      * @ignore
      */
-    it('Binary', function (done) {
+    it('Binary', function () {
       const binary = new Binary(Buffer.from('0123456789abcdef0123456789abcdef', 'hex'), 4);
       expect(inspect(binary)).to.equal('Binary("0123456789abcdef0123456789abcdef", 4)');
-      done();
     });
 
     /**
      * @ignore
      */
-    it('Code', function (done) {
+    it('BSONSymbol', function () {
+      const symbol = new BSONSymbol('sym');
+      expect(inspect(symbol)).to.equal('BSONSymbol("sym")');
+    });
+
+    /**
+     * @ignore
+     */
+    it('Code', function () {
       const code = new Code('this.a > i', { i: 1 });
       expect(inspect(code)).to.equal('Code("this.a > i", {"i":1})');
-      done();
     });
 
     /**
      * @ignore
      */
-    it('DBRef', function (done) {
+    it('DBRef', function () {
       const oid = new ObjectId('deadbeefdeadbeefdeadbeef');
       const dbref = new DBRef('namespace', oid, 'integration_tests_');
       expect(inspect(dbref)).to.equal(
         'DBRef("namespace", "deadbeefdeadbeefdeadbeef", "integration_tests_")'
       );
-      done();
     });
 
     /**
      * @ignore
      */
-    it('Decimal128', function (done) {
+    it('Decimal128', function () {
       const dec = Decimal128.fromString('1.42');
       expect(inspect(dec)).to.equal('Decimal128("1.42")');
-      done();
     });
 
     /**
      * @ignore
      */
-    it('Int32', function (done) {
+    it('Double', function () {
+      const double = new Double(-42.42);
+      expect(inspect(double)).to.equal('Double(-42.42)');
+    });
+
+    /**
+     * @ignore
+     */
+    it('Int32', function () {
       const int = new Int32(42);
       expect(inspect(int)).to.equal('Int32(42)');
-      done();
     });
 
     /**
      * @ignore
      */
-    it('Int32', function (done) {
-      const int = new Int32(42);
-      expect(inspect(int)).to.equal('Int32(42)');
-      done();
-    });
-
-    /**
-     * @ignore
-     */
-    it('Int32', function (done) {
-      const int = new Int32(42);
-      expect(inspect(int)).to.equal('Int32(42)');
-      done();
-    });
-
-    /**
-     * @ignore
-     */
-    it('Int32', function (done) {
+    it('Long', function () {
       const long = Long.fromString('42');
       expect(inspect(long)).to.equal('Long("42")');
-      done();
     });
 
     /**
      * @ignore
      */
-    it('MaxKey', function (done) {
+    it('MaxKey', function () {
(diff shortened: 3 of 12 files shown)


src/binary.ts
```diff
@@ -255,6 +255,35 @@ export class Binary {
...
+            .match(/^(.{8})(.{4})(.{4})(.{4})(.{12})$/)
+            ?.slice(1, 6)
+            .join('-');
+          return `UUID("${asUUID}")`;
+        }
+      // Fall through in case somebody did something weird and used an UUID
+      // with a non-standard length.
+      // eslint-disable-next-line no-fallthrough
+      default:
+        return `BinData(${this.sub_type}, "${asBuffer.toString('base64')}")`;
```
addaleax, line 284: @rose-m These are also shell types, so I think at least this one would probably be best changed as well (and, because the different names aren’t enough confusing, the arguments work pretty differently … `BinData()` in the shell takes subtype + base64 string in that order, `Binary` in BSON takes a “““binary””” string or buffer + subtype, i.e. in reverse order)… I think ideally this line could be something like 
```suggestion
        return `Binary(Buffer.from("${asBuffer.toString('base64')}", "base64"), ${this.sub_type})`;
```

but it sure isn’t pretty…

src/binary.ts
```diff
@@ -255,6 +255,35 @@ export class Binary {
     }
     return new Binary(data, type);
   }
+
+  /** @internal */
+  [Symbol.for('nodejs.util.inspect.custom')](): string {
```
mcasimir, line 260: `Symbol.for('nodejs.util.inspect.custom')` won't work with the version of `inspect` from browserify (which we use in the browser shell). Would also have some issue with older node.js versions, which we might not need to support though.

I think something like:

```
let inspectMethodName = Symbol.for('nodejs.util.inspect.custom');

if (!inspectMethodName) {
   try {
      // note that we can't detect much here cause webpack / parcel etc may mess with it
      // so we just try.
      inspectMethodName = require('util').inspect.custom;
   } catch (e) {
      inspectMethodName = 'inspect';
      // ...
   }
}
```

addaleax: @mcasimir This does provide a plain `inspect()` method, but I think it’s good practice to also define the symbol-based property if one provides the named variant

mcasimir: Oh i see what you mean: `require('util').inspect.custom` is always === 'inspect'?

addaleax: Uh … no, it’s always a symbol if it exists

What I mean is that in this PR, both the symbol variant and the plain `inspect() { ... }` variant are added, so it’s always going to work

mcasimir: With browserify too? I mean i think with browserify it very well be the string 'inspect'. I guess we can just read the code there. I'm sure few months ago was not`Symbol.for('nodejs.util.inspect.custom')` but if we are lucky and is always inspect we are good.

mcasimir: Ok: https://github.com/browserify/node-util/blob/0915f593c45eb131e3e7f4718e0b4861433acaff/test/node/inspect.js#L155

I now i understand, so browserify is using `inspect` directly, which is probably what was happening in some old node. Makes sense, so we don't need that weird hack.

src/code.ts
```diff
@@ -42,6 +42,18 @@ export class Code {
...
   }
+
+  /** @internal */
+  [Symbol.for('nodejs.util.inspect.custom')](): string {
+    return this.inspect();
+  }
+
+  inspect(): string {
+    const codeJson = this.toJSON();
+    return `{ "code" : "${codeJson.code}"${
```
mbroadst, line 53: This one, as well as `MaxKey`, `MinKey` seem to output some form of extended JSON. I'm sure this is a "shell-ism", but I think we should be consistent in our representation using `ClassName(...)`. Perhaps you can override this in mongosh to provide legacy shell compat?

mbroadst (changes requested):
Really just one small nit, but otherwise LGTM!

test/node/symbol_tests.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+const BSON = require('../register-bson');
+const util = require('util');
+const BSONSymbol = BSON.BSONSymbol;
+
+describe('MinKey tests', function () {
```
mbroadst, line 7:
```suggestion
describe('BSONSymbol tests', function () {
```

mbroadst: oops, missed this one. Actually, now that we're here does it make sense to just add all of these tests in [bson_tests](https://github.com/mongodb/js-bson/blob/master/test/node/bson_test.js) instead? I'm imagining a future maintainer forgetting to change one of these in the future because they have to change ~10 files

mbroadst (changes requested):

src/code.ts
rose-m: Sounds like a good idea - changed that for some types already but obviously missed some!

test/node/symbol_tests.js
rose-m: Yeah I wasn't sure how to handle that and then went with the individual files. I could implement a test case like https://github.com/mongodb/js-bson/blob/master/test/node/bson_test.js#L1310 which would `inspect` an object with all the different types and directly compare that instead of individual test cases?

mbroadst: I like the idea of checking the behavior in one place, it looks like maybe there's a build failure related to formatting of the string. Maybe it's easier to do something like:

rose-m: Ah alright now I see - sometimes it has line breaks, sometimes it doesn't. Will fix asap and split it up into individual cases to circumvent that.

rose-m: I've now added individual cases per type in one block in `bson_tests` - let's hope these now work in evergreen 😄

emadum (approved):
LGTM 👍

test/node/bson_test.js
```diff
@@ -2268,4 +2269,117 @@ describe('BSON', function () {
...
+    });
+
+    /**
+     * @ignore
+     */
+    it('Int32', function (done) {
+      const int = new Int32(42);
+      expect(inspect(int)).to.equal('Int32(42)');
+      done();
+    });
```
nbbeeken, line 2338: I don't see a difference between these tests. Accidental dupe? 
I think one of these should be `ObjectId` and another should be `Double` perhaps?

test/node/bson_test.js
```diff
@@ -2268,4 +2269,117 @@ describe('BSON', function () {
...
+
+    /**
+     * @ignore
+     */
+    it('Int32', function (done) {
```
nbbeeken, line 2343:
```suggestion
    it('Long', function (done) {
```

test/node/bson_test.js
```diff
@@ -2268,4 +2269,117 @@ describe('BSON', function () {
...
   });
+
+  describe('Should support util.inspect for', function () {
+    /**
+     * @ignore
+     */
+    it('Binary', function (done) {
+      const binary = new Binary(Buffer.from('0123456789abcdef0123456789abcdef', 'hex'), 4);
+      expect(inspect(binary)).to.equal('Binary("0123456789abcdef0123456789abcdef", 4)');
+      done();
```
nbbeeken, line 2280: nit: you only need `done();` when you have to run a callback so you can drop the done calls but that's optional.

test/node/bson_test.js
```diff
@@ -2268,4 +2269,117 @@ describe('BSON', function () {
...
+    it('MinKey', function (done) {
+      const minKey = new MinKey();
+      expect(inspect(minKey)).to.equal('MinKey()');
+      done();
+    });
+
+    /**
+     * @ignore
+     */
+    it('Symbol', function (done) {
```
nbbeeken, line 2370:
```suggestion
    it('BSONSymbol', function (done) {
```

nbbeeken (changes requested):
Thanks for contributing this!
A few patches to the tests and it looks good to me

test/node/bson_test.js
rose-m: Oh god 🙈 These were just dupes. I've added an implementation for `Double` nonetheless. The implementation of `ObjectId` currently is in another PR - #412 - and I skipped that here to avoid conflicts. If you'd like to I could move that in and we could close #412.

test/node/bson_test.js
rose-m: Right... was a little confused as I've found some other sync tests in there using the same pattern and wasn't that sure anymore 🤔 Sorry about that!

mbroadst (approved):
LGTM!

test/node/bson_test.js
nbbeeken: #412 is already merged so looks like we're good to go, thanks for adding Double!

nbbeeken (approved):

PR #415, opened 2020-12-01
fix: remove tslib usage and fix Long method alias

nbbeeken:
fix: remove tslib usage and fix Long method alias

### DIFF
--- a/src/bson.ts
+++ b/src/bson.ts
@@ -21,7 +21,40 @@ import { BSONSymbol } from './symbol';
 import { Timestamp } from './timestamp';
 export { BinaryExtended, BinaryExtendedLegacy, BinarySequence } from './binary';
 export { CodeExtended } from './code';
-export * from './constants';
+export {
+  BSON_BINARY_SUBTYPE_BYTE_ARRAY,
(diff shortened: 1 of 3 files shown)


mbroadst (approved):

src/bson.ts
```diff
@@ -21,7 +21,40 @@ import { BSONSymbol } from './symbol';
 import { Timestamp } from './timestamp';
 export { BinaryExtended, BinaryExtendedLegacy, BinarySequence } from './binary';
 export { CodeExtended } from './code';
-export * from './constants';
+export {
```
kmahar, line 24: why did this change?

kmahar (approved):
one question but LGTM

src/bson.ts
nbbeeken: This is what introduces the tslib import, exportStar needs some actions to replicate for commonjs.

https://github.com/microsoft/tslib/blob/fcba93d415a6d84914916769ba17871902dd4598/tslib.es6.js#L114

emadum (approved):

PR #416, opened 2020-12-04
fix: make inspection result of Binary evaluatable NODE-2947

addaleax:
Make sure that evaluating the inspection result of a `Binary` object
returns a result that can be evaluated to give an object equivalent
to the original.

### DIFF
--- a/src/binary.ts
+++ b/src/binary.ts
@@ -263,7 +263,7 @@ export class Binary {
 
   inspect(): string {
     const asBuffer = this.value(true);
-    return `Binary(Buffer.from("${asBuffer.toString('hex')}", "hex"), ${this.sub_type})`;
+    return `new Binary(Buffer.from("${asBuffer.toString('hex')}", "hex"), ${this.sub_type})`;
   }
 }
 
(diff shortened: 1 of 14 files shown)


emadum (approved):

nbbeeken (changes requested):
Thanks for this! it makes sense that the inspect prints should be "round-trippable" so to speak for debugging. 
Sorry to increase scope on you but would you mind adding Decimal128 to this fix? 

currently it outputs:
```js
> BSON.Decimal128.fromString('12345')
Decimal128('12345')
// But the constructor takes a sequence of bytes not a string:
> Decimal128('12345')
Decimal128("2.1542142465E-6166") // bad value
```
It should have the fromString() call on the end instead. 

Would it also make sense to add `new` to the beginning of the inspect strings?

addaleax:
@nbbeeken Sure, done!

> Would it also make sense to add `new` to the beginning of the inspect strings?

Well, yes, but I also didn't want to increase the scope here too much. Also done! :)

nbbeeken (approved):
LGTM, I'll merge when we get the ✅  :)

PR #418, opened 2020-12-10
isBuffer: Check object equality not class name

stefanor:
isBuffer: Check object equality not class name
    
If minified, the name of the buffer class may change, breaking the `isBuffer()` function, and thus the library.

Since 5167be2752369d5832057f7b69da0e240ed4c204 binary fields have been broken for users minifying their source. The `.buffer` attribute would contain the raw BSON blob, rather than just the field's value.

NODE-2963

### DIFF
--- a/src/parser/utils.ts
+++ b/src/parser/utils.ts
@@ -58,7 +58,7 @@ export function haveBuffer(): boolean {
 
 /** Callable in any environment to check if value is a Buffer */
 export function isBuffer(value: unknown): value is Buffer {
-  return isUint8Array(value) && typeof value?.toJSON === 'function';
+  return isUint8Array(value) && typeof (value as any)?.toJSON === 'function';
 }
 
 // To ensure that 0.4 of node works correctly
--- a/src/parser/utils.ts
+++ b/src/parser/utils.ts
@@ -58,7 +58,7 @@ export function haveBuffer(): boolean {
 
 /** Callable in any environment to check if value is a Buffer */
 export function isBuffer(value: unknown): value is Buffer {
-  return typeof value === 'object' && value?.constructor === Buffer;
+  return isUint8Array(value) && typeof value?.toJSON === 'function';
 }
 
 // To ensure that 0.4 of node works correctly
--- a/src/parser/utils.ts
+++ b/src/parser/utils.ts
@@ -58,7 +58,7 @@ export function haveBuffer(): boolean {
 
 /** Callable in any environment to check if value is a Buffer */
 export function isBuffer(value: unknown): value is Buffer {
-  return typeof value === 'object' && value?.constructor?.name === 'Buffer';
+  return typeof value === 'object' && value?.constructor === Buffer;
 }
 
 // To ensure that 0.4 of node works correctly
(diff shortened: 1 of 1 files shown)


nbbeeken:
Hello @stefanor thank you for your submission! Unfortunately this causes another issue on nodejs, essentially:

```js
global.Buffer != require('buffer').Buffer
```

So buffers constructed using the globally available constructor, which is common practice in Node, won't return true for this isBuffer helper. We will further investigate a solution that is cross platform.

stefanor:
Fair enough.

src/parser/utils.ts
```diff
@@ -58,7 +58,7 @@ export function haveBuffer(): boolean {
 
 /** Callable in any environment to check if value is a Buffer */
 export function isBuffer(value: unknown): value is Buffer {
-  return typeof value === 'object' && value?.constructor?.name === 'Buffer';
+  return typeof value === 'object' && value?.constructor === Buffer;
```
addaleax, line 61: How about e.g.

```suggestion
  return isUint8Array(value) && typeof value?.toJSON === 'function';
```

nbbeeken: I think this would work! Thanks for the suggestion, I'll merge it and see if I can safely minify it on my end

stefanor:
Thanks!

nbbeeken (approved):
Thanks again @stefanor for bringing this to our attention, after I get another pair of eyes from the team I'll see to merging this. 

I've made a ticket that we look into adding minification testing to our CI [NODE-3217](https://jira.mongodb.org/browse/NODE-3217)

src/parser/utils.ts
```diff
@@ -58,7 +58,8 @@ export function haveBuffer(): boolean {
 
 /** Callable in any environment to check if value is a Buffer */
 export function isBuffer(value: unknown): value is Buffer {
-  return typeof value === 'object' && value?.constructor?.name === 'Buffer';
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  return isUint8Array(value) && typeof (value as any)?.toJSON === 'function';
 }
```
dariakp, line 63: I think it would be better to check for the properties on the buffer that the driver actually needs to exist in order to successfully do its job instead of relying on an arbitrary condition to infer that the object is of the correct type; furthermore, we should add some unit tests around types that are expected to pass the check and the ones that are expected to fail

dariakp (changes requested):
See code comment regarding improving the check and adding tests

src/parser/utils.ts
addaleax: > I think it would be better to check for the properties on the buffer that the driver actually needs to exist in order to successfully do its job instead of relying on an arbitrary condition to infer that the object is of the correct type;

Soooo... if clean code is the goal here, the way to do that is to just never check whether something is a Buffer, because there's never a good reason to do that (e.g. all Node.js built-in APIs which accept a Buffer also accept all other kinds of Uint8Arrays). Luckily, that's actually pretty easy to do (bonus bugfix included 🙂): https://github.com/mongodb/js-bson/pull/432

dariakp: Yeah that was pretty much my thought, will take a look at the new code soon, thanks!

dariakp:
Closing this PR in favor of the more general approach in https://github.com/mongodb/js-bson/pull/432.

PR #423, opened 2021-02-04
check both window and self

fasidOnGit:
## Description

This PR makes this package usable in a web-worker.

Another approach would be to use `globalThis`. But, that would required node 12+, I see there is quite good support for it in browsers.


nbbeeken (approved):
Thanks for contributing this fix, LGTM we'll get this out in the next release.

PR #426, opened 2021-03-02
fix: Throw error on bigint usage and add helpers to Long

nbbeeken:
Adds helpers toBigInt and fromBigInt to the Long class to help
support 64bit values. Throws an error when `bigint` is attempted to
be serialized.

NODE-2378


emadum (approved):
LGTM 👍

durran (approved):
LGTM

PR #429, opened 2021-04-08
fix: improve ArrayBuffer brand check in ensureBuffer

addaleax:
## Description

`instanceof` is notoriously unreliable in JS, because it can
be fooled by prototype manipulation and breaks down entirely
for cross-realm brand checks.

Using `Object.prototype.toString` is generally a better choice
for this.

This also adds `SharedArrayBuffer` as an alternative, as both
should be treated the same under most circumstances.


durran (approved):

emadum (approved):

nbbeeken (approved):
LGTM!
(NODE-3195)

PR #430, opened 2021-04-15
fix: clean up instanceof usage

addaleax:
Similarly to 99722f66d, instanceof is not a great brand check in JS
in general, and using `Object.prototype.toString.call()` is better
at detecting built-in types.

This particular change is needed for mongosh, because the user input
it not evaluated in the same `vm.Context` as the bson package that
we’re using.

NODE-3208

## Description

**What changed?**


nbbeeken (approved):
Thanks for this fix, LGTM!

durran (approved):
LGTM

PR #431, opened 2021-04-22
fix(NODE-3153): correctly deserialize `__proto__` properties

addaleax:
NODE-3153

## Description

**What changed?**

### DIFF
--- a/src/parser/deserializer.ts
+++ b/src/parser/deserializer.ts
@@ -606,7 +606,7 @@ function deserializeObject(
       );
     }
     if (name === '__proto__') {
-      Object.defineProperty(object, name, { 
+      Object.defineProperty(object, name, {
         value,
         writable: true,
         enumerable: true,
--- a/src/parser/deserializer.ts
+++ b/src/parser/deserializer.ts
@@ -606,7 +606,12 @@ function deserializeObject(
       );
     }
     if (name === '__proto__') {
-      Object.defineProperty(object, name, { value, writable: true, enumerable: true, configurable: true });
+      Object.defineProperty(object, name, { 
+        value,
+        writable: true,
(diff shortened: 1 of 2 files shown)


emadum (approved):

src/parser/deserializer.ts
```diff
@@ -599,12 +599,22 @@ function deserializeObject(
...
       );
     }
+    if (name === '__proto__') {
+      Object.defineProperty(object, name, {
+        value,
+        writable: true,
+        enumerable: true,
+        configurable: true
+      });
+    } else {
```
dariakp, line 615: What's the reason for having an else for the non-proto cases? Are there special properties that the above block would not handle in the way we want?

addaleax: @dariakp This was a performance concern that @nbbeeken and I talked about in the ticket – I did benchmark this, `Object.defineProperty()` is (somewhat understandably) significantly slower than just setting the property directly.

dariakp: Oh that's interesting, I saw the note on the ticket saying that we weren't sure whether it was optimized by V8, but not the actual outcome of any benchmarking. Just out of curiosity, what *was* the performance degradation?

addaleax: It’s about 50 % slower: https://jsbench.me/w7ko1ntaf5/1

I guess it’s fair that V8 doesn’t optimize this as much, it’s not nearly as common for properties to be defined this way.

dariakp (approved):

src/parser/deserializer.ts
dariakp: Fair and yeah, that certainly qualifies as significant, thanks for providing the context!

PR #434, opened 2021-05-04
fix: make Long inspect result evaluable

addaleax:
Do basically the same thing we’re already doing for `Decimal128` to
ensure that the inspect result accurately represents the contents
in a way that can be used to evaluate it again.

NODE-3256

## Description

**What changed?**

### DIFF
--- a/src/long.ts
+++ b/src/long.ts
@@ -994,7 +994,7 @@ export class Long {
   }
 
   inspect(): string {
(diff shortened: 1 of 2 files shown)


dariakp (approved):
seems reasonable

durran (approved):
LGTM

PR #435, opened 2021-05-05
ObjectId's PROCESS_UNIQUE lazy loaded

steffenagger:
## Description

This PR changes `PROCESS_UNIQUE` in the `ObjectId` class to be lazy loaded. The main reason is allow for the `randomBytes` fallback-warning to only trigger if/when utilized.

I also took the liberty to remove som unused variables.


durran (approved):
LGTM

cristianoccazinsp:
Looking forward for this to be released as the warning is wrecking havoc in our error reporting tool :(

PR #436, opened 2021-05-12
test: bson corpus spec runner calling wrong fn

durran:
I ran across this one-liner when looking into `Decimal128` usages. (Of course they all were passing since `Decimal.128.fromString` would never be able to deserialize the EJSON, so we now use the EJSON parse helper function, which passes, but aren't false positives)

NODE-3277

### DIFF
--- a/test/node/bson_corpus_tests.js
+++ b/test/node/bson_corpus_tests.js
@@ -2,7 +2,6 @@
 
 const Buffer = require('buffer').Buffer;
 const BSON = require('../register-bson');
-const Decimal128 = BSON.Decimal128;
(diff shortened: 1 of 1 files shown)


dariakp:
@durran Looks good but I think per our convention, the commit message should be `test: ...` not `fix: ...` since we aren't fixing any functionality in the library itself

emadum (changes requested):
The require for `Decimal128` at the top also needs to be removed, currently it's causing a lint error in CI

nbbeeken (approved):

emadum (approved):

PR #437, opened 2021-05-17
feat: allow Decimal128(string), Long(string), Long(bigint)

addaleax:
## Description

This adds support for the following constructor signatures:

- `[new] Decimal128(string)`
- `[new] Long(string[, unsigned])`
- `[new] Long(bigint[, unsigned])`

While the first two of these would not have thrown exceptions before,
they would also not have returned any meaningful results, so this
is not a breaking change.

NODE-3264

### DIFF
--- a/src/decimal128.ts
+++ b/src/decimal128.ts
@@ -167,11 +167,18 @@ export class Decimal128 {
 
   readonly bytes!: Buffer;
 
-  /** @param bytes - a buffer containing the raw Decimal128 bytes in little endian order */
-  constructor(bytes: Buffer) {
+  /**
+   * @param bytes - a buffer containing the raw Decimal128 bytes in little endian order,
+   *                or a string representation as returned by .toString()
+   */
+  constructor(bytes: Buffer | string) {
     if (!(this instanceof Decimal128)) return new Decimal128(bytes);
 
-    this.bytes = bytes;
+    if (typeof bytes === 'string') {
+      this.bytes = Decimal128.fromString(bytes).bytes;
+    } else {
+      this.bytes = bytes;
+    }
   }
 
   /**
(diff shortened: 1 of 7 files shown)


src/decimal128.ts
```diff
@@ -796,7 +803,7 @@ export class Decimal128 {
   }
 
   inspect(): string {
-    return `Decimal128.fromString("${this.toString()}")`;
+    return `Decimal128("${this.toString()}")`;
```
nbbeeken, line 806: I think we should have `new` here to be consistent with the other types. Or we should make all the types not print out `new` in front of the type construction. @mongodb/dbx-node thoughts?

src/long.ts
```diff
@@ -994,7 +1006,7 @@ export class Long {
   }
 
   inspect(): string {
-    return `Long.fromString("${this.toString()}"${this.unsigned ? ', true' : ''})`;
+    return `Long("${this.toString()}"${this.unsigned ? ', true' : ''})`;
```
nbbeeken, line 1009: Leaving a comment here about the `new` decision we want to make.

nbbeeken (changes requested):

src/decimal128.ts
addaleax: @nbbeeken You’re right, thanks for pointing this out. I’ve added the `new` for consistency. If we can, let’s keep the `new`-or-not discussion separate from this ticket here. :)

emadum (approved):
LGTM

src/decimal128.ts
nbbeeken: Fair enough! Thanks :)

nbbeeken (approved):
🚀

PR #438, opened 2021-05-17
chore(test): add timeout to older versions test

addaleax:
## Description

This sometimes flakes out for me locally because it takes
slightly longer than the mocha default of 2 seconds to download
the legacy bson zip.


dariakp (approved):
Seems reasonable to me

nbbeeken (approved):
👍

PR #441, opened 2021-06-07
fix(NODE-3282): BSONRegExp options not alphabetized

andymina:
## Description
- Fixed alphabetizing the options.
- Corrected the degenerate BSON spec test to test deserializer


emadum (changes requested):
LGTM, nice work! One request - I don't think a new file is necessary for this one new test, could you put it in `test/node/bson_test.js` instead?

dariakp (approved):

emadum (approved):

PR #442, opened 2021-06-17
chore(NODE-3330): added mocha.opts

andymina:
## Description
- Added `mocha.opts` to `/test`


emadum (approved):

nbbeeken (changes requested):
let's delete .mocharc.json here too, next time one of comes around here and needs to change options they'll probably think the json is the place to go but it will have no effect

nbbeeken (approved):

dariakp (approved):

PR #443, opened 2021-06-24
fix(NODE-3247): bump spec tests and fix up DBRef handling

addaleax:
## Description

I wanted to bump the spec tests for NODE-3094, turns out there’s
nothing to do there, but NODE-3247 breaks the tests, so I’ve made
changes that align the handling with the spec here.


nbbeeken (approved):
LGTM

dariakp (approved):

emadum (approved):
LGTM

PR #444, opened 2021-06-25
fix(NODE-3376): use standard JS methods for copying Buffers

addaleax:
## Description

- Use standard JS methods for copying Buffers, because unlike the
  `buffer` module polyfill, these have working typechecking
- Add tests that ensure that for all BSON distributions and all
  legacy packages, their serialization is cross-compatible
  (modulo NODE-2848, which would require updates to either the
  `buffer` poyfill or the legacy 1.x bson versions).

### DIFF
--- a/test/bson_older_versions_tests.js
+++ b/test/bson_older_versions_tests.js
@@ -80,16 +80,25 @@ describe('Mutual version and distribution compatibility', function () {
 
   // Node.js requires an .mjs filename extension for loading ES modules.
   before(() => {
-    fs.writeFileSync(
-      './bson.browser.esm.mjs',
-      fs.readFileSync(__dirname + '/../dist/bson.browser.esm.js')
-    );
-    fs.writeFileSync('./bson.esm.mjs', fs.readFileSync(__dirname + '/../dist/bson.esm.js'));
+    try {
+      fs.writeFileSync(
+        './bson.browser.esm.mjs',
+        fs.readFileSync(__dirname + '/../dist/bson.browser.esm.js')
+      );
+      fs.writeFileSync('./bson.esm.mjs', fs.readFileSync(__dirname + '/../dist/bson.esm.js'));
+    } catch (e) {
+      // bundling fails in CI on Windows, no idea why, hence also the
+      // process.platform !== 'win32' check below
+    }
   });
 
   after(() => {
-    fs.unlinkSync('./bson.browser.esm.mjs');
-    fs.unlinkSync('./bson.esm.mjs');
+    try {
+      fs.unlinkSync('./bson.browser.esm.mjs');
+      fs.unlinkSync('./bson.esm.mjs');
+    } catch (e) {
+      // ignore
+    }
   });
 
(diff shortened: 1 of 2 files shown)


test/bson_older_versions_tests.js
```diff
@@ -73,18 +76,160 @@ describe('Current version', function () {
...
+    try {
+      fs.writeFileSync(
+        './bson.browser.esm.mjs',
+        fs.readFileSync(__dirname + '/../dist/bson.browser.esm.js')
       );
-    });
+      fs.writeFileSync('./bson.esm.mjs', fs.readFileSync(__dirname + '/../dist/bson.esm.js'));
+    } catch (e) {
+      // bundling fails in CI on Windows, no idea why, hence also the
+      // process.platform !== 'win32' check below
```
addaleax, line 91: Lmk if there is a ticket for this so I can mention it here, I couldn’t find any, but this issue doesn’t appear to be new.

test/bson_older_versions_tests.js
```diff
@@ -73,18 +76,160 @@ describe('Current version', function () {
...
+  })).concat([
+    {
+      name: 'Node.js lib/bson',
+      load: () => Promise.resolve(currentNodeBSON)
+    },
+    {
+      name: 'Browser ESM',
+      // eval because import is a syntax error in earlier Node.js versions
+      // that are still supported in CI
+      load: () => eval(`import("${__dirname}/../bson.browser.esm.mjs")`),
```
nbbeeken, line 122: 🤯 genius

test/bson_older_versions_tests.js
```diff
@@ -73,18 +76,160 @@ describe('Current version', function () {
...
+        });
+
+        it('serializes in a compatible way', function () {
+          for (const object of fromObjects) {
+            // If the object in question uses Buffers in its serialization, and
+            // its Buffer was created using the polyfill, and we're serializing
+            // using a legacy version that uses buf.copy(), then that fails
+            // because the Buffer polyfill's typechecking is buggy, so we
+            // skip these cases.
+            // This is tracked as https://jira.mongodb.org/browse/NODE-2848
```
nbbeeken, line 200: Just to cement my understanding we could change bson 1.x to use .set and .subarray to correct the incompatibility right?

nbbeeken (approved):

test/bson_older_versions_tests.js
nbbeeken: I do not see one already but I think this fine to leave out, since we test the functionality of the bson lib on windows this bundle compatibility seem adequately covered if it runs on at least linux. We can always file a ticket to track but I don't think we'll need to look into this

test/bson_older_versions_tests.js
addaleax: Yes, that would work :+1:

dariakp (approved):
Very nice work on the tests! @addaleax

emadum (approved):
LGTM! 👍

PR #445, opened 2021-06-28
fix(NODE-3390): serialize non-finite doubles correctly in EJSON

addaleax:
## Description

In relaxed mode, non-finite doubles are supposed to be serialized
the same way as in non-relaxed/canonical mode.


emadum (approved):
LGTM

PR #457, opened 2021-09-01
fix(NODE-3582): fix internal marked APIs, add toString methods to Int32 and Double

nbbeeken:
We have a number of the convenience helpers, toString and toJSON marked as internal on our Type classes. The should be made available in our type definitions. 

Additionally we were missing toString methods for Int32 and Double, I think its unexpected that these would return `[object Object]` for toString. 

I can revert if we want to do this seperately.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,7 +17,6 @@ bson.sublime-project
 bson.sublime-workspace
 
 lib
-types/
 .nyc_output/
 coverage/
 *.d.ts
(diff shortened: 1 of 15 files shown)


nbbeeken:
@addaleax I'd appreciate a review if you have the time (at-ing since I can't request directly)

addaleax (approved):

emadum (approved):
LGTM 👍

dariakp (approved):

PR #459, opened 2021-09-17
fix(NODE-3493): code and symbol tests are partially testing the wrong types

nbbeeken:
Fixes: NODE-3493


dariakp (commented):
Let's check the performance on this before merging

dariakp (approved):

PR #464, opened 2021-09-24
fix(NODE-3629): correct corpus runner and add null checks

nbbeeken:
The runner changes break out each parseError into its own function by type. Three new errors are added, key checking when parsing EJSON and at BSONRegex construction.


test/register-bson.js
```diff
@@ -8,10 +8,31 @@
...
+      // By default, lets check for BSONError or BSONTypeError
+      // Since we compile to es5 instanceof is broken???
+      const assertion = original.apply(this, args);
+      const object = util.flag(assertion, 'object');
+      return this.assert(
+        object && /BSONError|BSONTypeError/.test(object.stack),
+        'expected #{this} to be a BSONError or a BSONTypeError but got #{act}',
+        'expected #{this} to not be a BSONError nor a BSONTypeError but got #{act}',
+        (object && object.stack) || '<error undefined>'
+      );
```
nbbeeken, line 31: If someone makes use of a plain Error going forward this is what the test would fail with.
```
AssertionError: expected [Error: bson size must be >= 5, is 1] to throw a BSONError or a BSONTypeError but got [Error: bson size must be >= 5, is 1]
```

test/node/bson_corpus_tests.js
```diff
@@ -48,40 +49,97 @@ function normalize(cEJ) {
...
     'Current BSON implementation of timestamp/long cannot hold these values - 1 too large.',
   'Timestamp with high-order bit set on both seconds and increment (not UINT32_MAX)':
     'Current BSON implementation of timestamp/long cannot hold these values - 1 too large.'
 };
 
+const SKIP_TESTS = new Map([
+  ...Object.entries(skipBSON),
+  ...Object.entries(skipExtendedJSON),
+  [
+    'All BSON types',
```
dariakp, line 124: can we create/link a ticket to unskip this?

test/node/bson_corpus_tests.js
```diff
@@ -48,40 +49,97 @@ function normalize(cEJ) {
...
+        EJSON.parse(parseError.string);
+      } catch (error) {
+        caughtError = error;
+      }
+
+      if (/Null/.test(parseError.description)) {
+        expect(caughtError).to.be.instanceOf(Error);
+        expect(caughtError.message).to.match(/null bytes/);
+      } else if (/Bad/.test(parseError.description)) {
+        // There is a number of failing tests that start with 'Bad'
```
dariakp, line 94: can we create/link a ticket to fix this?

test/node/bson_corpus_tests.js
```diff
@@ -48,40 +49,97 @@ function normalize(cEJ) {
   return JSON.stringify(JSON.parse(cEJ));
 }
 
-// tests from the corpus that we need to skip, and explanations why
+const parseErrorForDecimal128 = scenario => {
+  const skipRegex = /dqbsr|Inexact/;
+  for (const parseError of scenario.parseErrors) {
+    it(parseError.description, function () {
+      if (skipRegex.test(parseError.description)) {
```
dariakp, line 56: can we add ticket to unskip?

test/register-bson.js
dariakp: Can you remind me why we need to do the method overrides or why we needed to create the new error classes to begin with? My memory of last week's discussions is a bit hazy...

dariakp (changes requested):

test/node/bson_corpus_tests.js
nbbeeken: These were all skipped before this work I just reworked the logic that checked for the test name to make the runner more consistent, look for the `'All BSON types'` string on the deletions side. So I'm pretty sure there's already tickets for these if they aren't inherent limitations to our driver.

test/node/bson_corpus_tests.js
nbbeeken: This would also be covered by: jira.mongodb.org/browse/NODE-3637

nbbeeken:
> Can you remind me why we need to do the method overrides or why we needed to create the new error classes to begin with? My memory of last week's discussions is a bit hazy...

(Can't reply directly for some reason) but we covered this offline, basically have to do some extra leg work because of compilation level and capturing intentional lib errors as opposed to misuse of API.

test/register-bson.js
nbbeeken: We covered this offline, basically have to do some extra leg work because of compilation level and capturing intentional lib errors as opposed to misuse of API.

test/node/bson_corpus_tests.js
dariakp: any objections to including it in a code comment here for easier reference? (same on the other place)

dariakp (approved):

test/node/bson_corpus_tests.js
dariakp: tyvm :)

Issue #234, opened 2018-02-26
Support for UUID binary type


niccottrell:
I'd like to be able to do:

```
var bson = require('bson');
return bson.UUID();
```

Is there a way to do this today? Can this helper function be added?

mbroadst:
@niccottrell Hi! This can be achieved by providing [BSON.BSON_BINARY_SUBTYPE_UUID](https://github.com/mongodb/js-bson/blob/1.0-branch/lib/bson/binary.js#L19) to the `Binary` [constructor](https://github.com/mongodb/js-bson/blob/1.0-branch/lib/bson/binary.js#L28).

typesafe:
I think it's really simple to extend the `serializer`/`deserializer` implementations to allows for effortless/seamless use of fields of type UUID. Actually I have already done so in a local PoC, where I can insert and query instances of values like this:

```
class Entity {
  ...
  someId: UUID; 
  ...
}
```

How do you feel about a PR for this?

nbbeeken:
Hi @typesafe we now have a UUID type in the BSON library that has convenience methods, here's an example of how it is used:

```js
import { UUID } from 'bson'
const id = new UUID()
await collection.insertOne({ field: id.toBinary() })
const result = await collection.findOne({ field: id.toBinary() })
const foundId = result.field.toUUID()
```

There's more documentation surrounding this in our specifications here: [Implicitly serializing UUID](https://github.com/mongodb/specifications/blob/master/source/uuid.rst#implicit-decoding-and-encoding)

PR #465, opened 2021-10-17
feat(NODE-1046): natively support UUID (de)serialization

typesafe:
This introduces a global `DefaultDeserializeOptions.convertUUIDs` and also extends the existing `DeserializeOptions` with a new `convertUUIDs` flag. Setting the flag will simply deserialize `Binary` values with subtype `SUBTYPE_UUID` as actual `UUID` instances.

It's great that we have the `UUID` type, but I really want to avoid having to tediously map these values/fields back-and-fort. This is bound to end up in subtle bugs sooner or later. As a bonus, `UUID` also features a `toJSON` eliminating mapping altogether when returning data as JSON.

In case you're using Typescript, you can define fields as `UUID` and it "just works":

```
export interface SomeDocument {
...
externalId: UUID;
...
}
```

I tried to get some feedback via https://github.com/mongodb/js-bson/issues/234, but thought I might as well open a PR 🙂.

### DIFF
--- a/src/bson.ts
+++ b/src/bson.ts
@@ -15,7 +15,11 @@ import { ObjectId } from './objectid';
 import { BSONError, BSONTypeError } from './error';
 import { calculateObjectSize as internalCalculateObjectSize } from './parser/calculate_size';
 // Parts of the parser
-import { deserialize as internalDeserialize, DeserializeOptions } from './parser/deserializer';
+import {
+  deserialize as internalDeserialize,
+  DeserializeOptions,
+  DefaultDeserializeOptions
+} from './parser/deserializer';
 import { serializeInto as internalSerialize, SerializeOptions } from './parser/serializer';
 import { BSONRegExp } from './regexp';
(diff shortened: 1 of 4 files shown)


typesafe:
Hey @dariakp & @nbbeeken, I was wondering if there an update on this 😇

dariakp:
@typesafe Unfortunately due to higher priority issues we have not yet had a chance to schedule this, and I don't know whether we'll be able to in the upcoming quarter. I did send the ticket into the triage queue to see if we can get something more concrete in terms of a date.

nbbeeken:
Hi @typesafe, thanks again for your assistance with this feature, we've fleshed out a design to make sure we hit all the marks here. EJSON and BSON will both support the UUID instance, along with an option to control the new changes. You can look forward to the feature landing soon, and you can view the project on our [JIRA here](https://jira.mongodb.org/browse/NODE-4405?jql=%22Epic%20Link%22%20%3D%20NODE-1046)
ref: #508

PR #467, opened 2021-10-26
fix(NODE-3662): error checking to make sure that ObjectId results in object with correct properties

gjchong25:
Tightens the logic within the ObjectId constructor to make sure that inputs to the constructor are of the proper types, otherwise throws a TypeError. Tests to make sure that invalid inputs have an error thrown (rather than creating a "broken" ObjectId object with no 'id' property, for instance) are added.

### DIFF
--- a/src/objectid.ts
+++ b/src/objectid.ts
@@ -62,7 +62,9 @@ export class ObjectId {
           try {
             this[kId] = ensureBuffer(id.id);
           } catch {
-            throw new BSONTypeError('Argument passed in must have an id that is of type string or Buffer');
+            throw new BSONTypeError(
+              'Argument passed in must have an id that is of type string or Buffer'
+            );
           }
         }
       } else if (id == null || typeof id === 'number') {
@@ -85,9 +87,7 @@ export class ObjectId {
           );
         }
       } else {
-        throw new TypeError(
-          'Argument passed in does not match the accepted types'
-        );
+        throw new BSONTypeError('Argument passed in does not match the accepted types');
       }
       // If we are caching the hex string
       if (ObjectId.cacheHexString) {
--- a/src/objectid.ts
+++ b/src/objectid.ts
@@ -31,7 +31,7 @@ export class ObjectId {
   _bsontype!: 'ObjectId';
 
   /** @internal */
-  static index = ~~(Math.random() * 0xffffff);
+  static index = Math.floor(Math.random() * 0xffffff);
 
   static cacheHexString: boolean;
 
@@ -48,6 +48,7 @@ export class ObjectId {
   constructor(inputId?: string | Buffer | number | ObjectIdLike | ObjectId) {
     if (!(this instanceof ObjectId)) return new ObjectId(inputId);
 
+    // workingId is set based on type of input and whether valid id exists for the input
     let workingId;
     if (typeof inputId === 'object' && inputId && 'id' in inputId) {
       if (typeof inputId.id !== 'string' && !ArrayBuffer.isView(inputId.id)) {
@@ -63,7 +64,8 @@ export class ObjectId {
     } else {
       workingId = inputId;
     }
-    // the following checks use workingId to construct an ObjectId
+
+    // the following cases use workingId to construct an ObjectId
     if (workingId == null || typeof workingId === 'number') {
       // The most common use case (blank id, new objectId instance)
       // Generate a new id
@@ -152,7 +154,7 @@ export class ObjectId {
    */
   static generate(time?: number): Buffer {
     if ('number' !== typeof time) {
-      time = ~~(Date.now() / 1000);
+      time = Math.floor(Date.now() / 1000);
     }
 
     const inc = ObjectId.getInc();
--- a/src/objectid.ts
+++ b/src/objectid.ts
@@ -43,44 +43,44 @@ export class ObjectId {
   /**
    * Create an ObjectId type
    *
-   * @param id - Can be a 24 character hex string, 12 byte binary Buffer, or a number.
+   * @param inputId - Can be a 24 character hex string, 12 byte binary Buffer, or a number.
    */
-  constructor(id?: string | Buffer | number | ObjectIdLike | ObjectId) {
-    if (!(this instanceof ObjectId)) return new ObjectId(id);
+  constructor(inputId?: string | Buffer | number | ObjectIdLike | ObjectId) {
+    if (!(this instanceof ObjectId)) return new ObjectId(inputId);
 
     // Duck-typing to support ObjectId from different npm packages
-    if (id instanceof ObjectId) {
-      this[kId] = id.id;
-      this.__id = id.__id;
+    if (inputId instanceof ObjectId) {
+      this[kId] = inputId.id;
+      this.__id = inputId.__id;
     } else {
-      if (typeof id === 'object' && id && 'id' in id) {
-        if ('toHexString' in id && typeof id.toHexString === 'function') {
-          this[kId] = Buffer.from(id.toHexString(), 'hex');
-        } else if (typeof id.id === 'string') {
-          this[kId] = Buffer.from(id.id);
+      if (typeof inputId === 'object' && inputId && 'id' in inputId) {
+        if ('toHexString' in inputId && typeof inputId.toHexString === 'function') {
+          this[kId] = Buffer.from(inputId.toHexString(), 'hex');
+        } else if (typeof inputId.id === 'string') {
+          this[kId] = Buffer.from(inputId.id);
         } else {
           try {
-            this[kId] = ensureBuffer(id.id);
+            this[kId] = ensureBuffer(inputId.id);
           } catch {
             throw new BSONTypeError(
               'Argument passed in must have an id that is of type string or Buffer'
             );
           }
         }
-      } else if (id == null || typeof id === 'number') {
+      } else if (inputId == null || typeof inputId === 'number') {
         // The most common use case (blank id, new objectId instance)
         // Generate a new id
-        this[kId] = ObjectId.generate(typeof id === 'number' ? id : undefined);
-      } else if (ArrayBuffer.isView(id) && id.byteLength === 12) {
-        this[kId] = ensureBuffer(id);
-      } else if (typeof id === 'string') {
-        if (id.length === 12) {
-          const bytes = Buffer.from(id);
+        this[kId] = ObjectId.generate(typeof inputId === 'number' ? inputId : undefined);
+      } else if (ArrayBuffer.isView(inputId) && inputId.byteLength === 12) {
+        this[kId] = ensureBuffer(inputId);
+      } else if (typeof inputId === 'string') {
+        if (inputId.length === 12) {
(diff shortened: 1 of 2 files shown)


src/objectid.ts
```diff
@@ -43,54 +43,49 @@ export class ObjectId {
...
-        this[kId] = typeof id.id === 'string' ? Buffer.from(id.id) : id.id;
+        workingId = inputId.id;
       }
+    } else {
+      workingId = inputId;
     }
-
-    // The most common use case (blank id, new objectId instance)
-    if (id == null || typeof id === 'number') {
+    if (workingId == null || typeof workingId === 'number') {
```
nbbeeken, line 66: nit: let's put a newline (and maybe a comment) here. At this stage the workingId should be setup, what follows is how the workingId gets used to make our ObjectId

test/node/object_id_tests.js
```diff
@@ -5,10 +5,11 @@ const BSON = require('../register-bson');
 const util = require('util');
 const ObjectId = BSON.ObjectId;
 
-describe('ObjectId', function () {
+describe.only('ObjectId', function () {
```
nbbeeken, line 8: Mistaken `.only` here 🙂

test/node/object_id_tests.js
```diff
@@ -24,6 +25,195 @@ describe('ObjectId', function () {
...
+      }
+    };
+
+    expect(() => new ObjectId(objectIdLike).toHexString()).to.not.throw(TypeError);
+  });
+
+  it('should correctly create ObjectId from number or null', function () {
+    expect(() => new ObjectId(42).toHexString()).to.not.throw();
+    expect(() => new ObjectId(0x2a).toHexString()).to.not.throw();
+    expect(() => new ObjectId(NaN).toHexString()).to.not.throw();
```
nbbeeken, line 193: Interesting.. NaN doesn't throw, it's probably silently doing something undesirable though, can you add a floating point number here as well.

In the generate() function this will get passed to buffer.writeUInt32LE(), which may do flooring? We should add a test to check.

test/node/object_id_tests.js
```diff
@@ -60,6 +250,12 @@ describe('ObjectId', function () {
     done();
   });
 
+  it('should throw an error if invalid Buffer passed in', function (done) {
+    var a = Buffer.from([1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13]);
+    expect(() => new ObjectId(a)).to.throw(TypeError);
+    done();
```
nbbeeken, line 256: nit: No need to call done, or have it passed in to the test here, done is needed when you are testing something async.

test/node/object_id_tests.js
```diff
@@ -24,6 +25,195 @@ describe('ObjectId', function () {
...
+    var objectBufferId = {
+      id: validBuffer1,
+      toHexString: newToHexString
+    };
+    var objectBufferFromArray = {
+      id: validBuffer2,
+      toHexString: newToHexString
+    };
+    expect(() => new ObjectId(objectBufferId).toHexString()).to.not.throw(TypeError);
+    expect(Buffer.from(new ObjectId(objectBufferId).id).equals(bufferToHex));
```
nbbeeken, line 121: Maybe this isn't the intention of the test, but I remember we wanted to have a test show that the toHexString takes precedent. I think here we have an objectId made up of 'AAAA' but the toHexString returns 'BBB' so we can add a check that the resulting ObjectId _**is not**_ made up of 'A's

nbbeeken (changes requested):
Looking good, some nits and notes :)

test/node/object_id_tests.js
gjchong25: Resolved offline after discussion

test/node/object_id_tests.js
```diff
@@ -1,5 +1,4 @@
 'use strict';
-
 const Buffer = require('buffer').Buffer;
```
dariakp, line 2: I think we should keep the new line between `'use strict'` and the `require` statements that follow, and delete the line between the `@ignore` and the following `it` block. Even better, the `@ignore` blocks can be removed entirely, since they serve no purpose here (Neal tells me it was for old jsdoc which was never even ran on the test files).

test/node/object_id_tests.js
```diff
@@ -24,6 +24,192 @@ describe('ObjectId', function () {
     done();
   });
 
+  it('should correctly create ObjectId from ObjectId', function () {
+    var tmp = new ObjectId();
```
dariakp, line 28: this is super nitpicky, but since we're already using `const` in this file anyway, could you update the tests to use `const` and `let` instead of `var`? (`var` is the legacy style javascript)

test/node/object_id_tests.js
```diff
@@ -24,6 +24,192 @@ describe('ObjectId', function () {
...
+      toHexString: function () {
+        return tmp.toHexString();
+      }
+    };
+
+    expect(() => new ObjectId(objectIdLike)).to.throw(TypeError);
+  });
+
+  it('should correctly create ObjectId from object with valid string id', function () {
+    var objectValidStringId1 = {
```
dariakp, line 56: could we name the variables after the valid cases they represent? (something containing `24` and `12` since the length is what matters); similarly for the other string/buffer cases in the file

test/node/object_id_tests.js
```diff
@@ -24,6 +24,192 @@ describe('ObjectId', function () {
...
+      id: 'aaaaaaaaaaaaaaaaaaaaaaaa'
+    };
+    var objectValidStringId2 = {
+      id: 'abcdefghijkl'
+    };
+    var buf1 = Buffer.from('aaaaaaaaaaaaaaaaaaaaaaaa', 'hex');
+    var buf2 = Buffer.from('abcdefghijkl', 'hex');
+    expect(Buffer.from(new ObjectId(objectValidStringId1).id).equals(buf1));
```
dariakp, line 64: why are we calling Buffer.from on the returned id? It should already be a buffer, so the extra wrapping seems unnecessary for the check (same goes for all the rest of the wrapped ObjectId's in the tests in this file). Also, this assertion is not actually running: if you look closely, you are calling `equals` on the generated Buffer object, not on the expect (you can tell by the extra closing parenthesis at the end of the line).

test/node/object_id_tests.js
```diff
@@ -24,6 +24,192 @@ describe('ObjectId', function () {
...
+
+  const numericIO = [
+    { input: 42, output: 42 },
+    { input: 0x2a, output: 0x2a },
+    { input: 4.2, output: 4 },
+    { input: NaN, output: 0 }
+  ];
+
+  for (const { input, output } of numericIO) {
+    it(`should correctly create ObjectId from ${input} and result in ${output}`, function () {
```
dariakp, line 174: you may want to add an extra description field to the test cases because `0x2a` is actually going to get rendered as `42` in the string test case description when it prints the test list, so the only way for it to show the right test case name is to pass in the string `'0x2a'`

test/node/object_id_tests.js
```diff
@@ -24,6 +24,192 @@ describe('ObjectId', function () {
...
+      expect(objId.id).to.be.instanceOf(Buffer);
+      expect(objId.id.readUInt32BE(0)).to.equal(output);
+    });
+  }
+
+  it('should correctly create ObjectId undefined or null', function () {
+    const objNull = new ObjectId(null);
+    const objNoArg = new ObjectId();
+    const objUndef = new ObjectId(undefined);
+    expect(objNull.id).to.be.instanceOf(Buffer);
```
dariakp, line 186: can we also put the expected buffer value check for these test cases?

test/node/object_id_tests.js
```diff
@@ -24,6 +24,192 @@ describe('ObjectId', function () {
...
+    expect(() => new ObjectId('FFFFFFFFFFFFFFFFFFFFFFFG')).to.throw(TypeError);
+    expect(() => new ObjectId('thisstringisdefinitelytoolong')).to.throw(TypeError);
+    expect(() => new ObjectId('tooshort')).to.throw(TypeError);
+    expect(() => new ObjectId('101010')).to.throw(TypeError);
+    expect(() => new ObjectId('')).to.throw(TypeError);
+  });
+
+  it('should correctly create ObjectId from 12 byte or 24 hex string', function () {
+    var str1 = 'AAAAAAAAAAAAAAAAAAAAAAAA';
+    var str2 = 'FFFFFFFFFFFFFFFFFFFFFFFF';
```
dariakp, line 201: why do we want 2 instances of the 24 string? also, the expect statements here are not executing, for the same reason as the other one above (if you use the chaining keywords, like `expect(...).to.equal(...)`, you're less likely to run into this issue)

test/node/object_id_tests.js
```diff
@@ -24,6 +24,192 @@ describe('ObjectId', function () {
...
+    var str3 = 'abcdefghijkl';
+    expect(new ObjectId(str1).id.equals(Buffer.from(str1, 'hex')));
+    expect(new ObjectId(str2).id.equals(Buffer.from(str2, 'hex')));
+    expect(new ObjectId(str3).id.equals(Buffer.from(str3, 'hex')));
+  });
+
+  it('should correctly create ObjectId from 12 byte sequence', function () {
```
dariakp, line 208: Is this case different from the other string 12 one? I can't tell what case it's actually trying to address

dariakp (changes requested):

test/node/object_id_tests.js
gjchong25: We can't check for an expected buffer value for these cases, since each time this runs it'll generate a random Buffer based on the timestamp.

test/node/object_id_tests.js
gjchong25: I think Neal wanted me to separate out this assertion to its own test (one of his original PR comments) because its a byte sequence, not just a 12-char string. So he wanted to make sure that the bytes in [kId] are the same as using Buffer.from(s, 'latin1') on the string.

nbbeeken (approved):

test/node/object_id_tests.js
```diff
@@ -6,16 +6,12 @@ const util = require('util');
...
-   */
   it('should correctly handle objectId timestamps', function (done) {
-    // var test_number = {id: ObjectI()};
-    var a = ObjectId.createFromTime(1);
+    let a = ObjectId.createFromTime(1);
     expect(Buffer.from([0, 0, 0, 1])).to.deep.equal(a.id.slice(0, 4));
     expect(1000).to.equal(a.getTimestamp().getTime());
 
-    var b = new ObjectId();
+    let b = new ObjectId();
```
dariakp, line 14: This and the declaration in L10 can be a `const`. One somewhat weird thing in JS that you might not be aware of is that `const` does not mean "unchanging", it only means "unchanging reference", so even though we set a property on `b` in L14, as far as JS is concerned, `b` is still a const, because its reference is unchanged.

dariakp (approved):
One tiny comment here, but looking great otherwise. Thank you for doing all this clean up!

dariakp (approved):

PR #468, opened 2021-11-01
fix(Node 3724): Fix BSONTypeError and BSONError to correctly handle instanceof checks

gjchong25:
Fixes the constructors for BSONTypeError and TypeError with Object.setPrototypeOf so that instanceof checks pass with both error types. The assertion override for `throw` in `register-bson.ts` is removed since it is no longer necessary with the fix. Tests making sure the errors respond properly to instanceof checks, show up with the correct names in the stack trace, and pass in input error messages correctly are added.

### DIFF
--- a/src/error.ts
+++ b/src/error.ts
@@ -1,7 +1,7 @@
 /** @public */
 export class BSONError extends Error {
-  constructor(m: string) {
-    super(m);
+  constructor(message: string) {
+    super(message);
     Object.setPrototypeOf(this, BSONError.prototype);
   }
 
@@ -12,8 +12,8 @@ export class BSONError extends Error {
 
 /** @public */
 export class BSONTypeError extends TypeError {
-  constructor(m: string) {
-    super(m);
+  constructor(message: string) {
+    super(message);
     Object.setPrototypeOf(this, BSONTypeError.prototype);
   }
 
--- a/src/error.ts
+++ b/src/error.ts
@@ -1,5 +1,10 @@
 /** @public */
 export class BSONError extends Error {
+  constructor(m: string) {
+    super(m);
+    Object.setPrototypeOf(this, BSONError.prototype);
+  }
+
   get name(): string {
     return 'BSONError';
   }
@@ -7,6 +12,11 @@ export class BSONError extends Error {
 
 /** @public */
 export class BSONTypeError extends TypeError {
+  constructor(m: string) {
+    super(m);
+    Object.setPrototypeOf(this, BSONTypeError.prototype);
+  }
+
   get name(): string {
     return 'BSONTypeError';
   }
--- a/test/node/error_test.js
+++ b/test/node/error_test.js
@@ -4,23 +4,31 @@ const BSON = require('../register-bson');
 const BSONTypeError = BSON.BSONTypeError;
 const BSONError = BSON.BSONError;
 
-describe('BSON error tests', function () {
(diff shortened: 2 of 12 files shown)


src/error.ts
```diff
@@ -1,12 +1,22 @@
 /** @public */
 export class BSONError extends Error {
+  constructor(m: string) {
+    super(m);
```
nbbeeken, line 4:
```suggestion
  constructor(message: string) {
    super(message);
```

IMO, Just preferable to be explicit. (Same for the other class)

test/node/error_test.js
```diff
@@ -0,0 +1,25 @@
...
+const BSONTypeError = BSON.BSONTypeError;
+const BSONError = BSON.BSONError;
+
+describe('BSON error tests', function () {
+  it('should correctly validate instanceof checks for BSONTypeError and BSONError', function () {
+    const bsonTypeErr = new BSONTypeError();
+    const bsonErr = new BSONError();
+
+    expect(bsonTypeErr instanceof BSONTypeError).to.be.true;
+    expect(bsonErr instanceof BSONError).to.be.true;
```
nbbeeken, line 13: We can use chai's assert here for some more readability and also just add a check that it still correctly walks the protoype chain to return true for the base class:

```js
expect(bsonTypeErr).to.be.instanceOf(BSONTypeError);
expect(bsonTypeErr).to.be.instanceOf(TypeError);
```

nbbeeken (changes requested):
just some small nits, ty!

nbbeeken (approved):

test/node/error_test.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+const BSON = require('../register-bson');
+const BSONTypeError = BSON.BSONTypeError;
+const BSONError = BSON.BSONError;
+
+describe('BSON error tests', function () {
```
dariakp, line 7: Can we restructure these just a bit for clarity? I think if we add a separate describe block for `BSONError` and `BSONTypeError` and then put the corresponding `it` statements instead each, it would set up a better precedent in case we decide to add more errors. The `it` statements would be something like `should evaluate true on instanceof BSONError and Error`/`should evaluate true on instanceof BSONTypeError and TypeError`, and then for the assertions, I would check both `instanceof ... to.be.true` and `to.be.instanceOf(...)` because that way we ensure that our tests work as expected and that the actual `instanceof` check that an external user might use would work as expected even if mocha decides to change its underlying implementation of `instanceOf` for some reason; and then for the second `it`, make it a `should correctly set the name and message properties`

dariakp (changes requested):
Just a few organizational suggestions here

test/node/error_test.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
+
+const BSON = require('../register-bson');
+const BSONTypeError = BSON.BSONTypeError;
+const BSONError = BSON.BSONError;
+
+describe('BSONTypeError', function () {
+  it('should evaluate true on instanceof BSONTypeError and TypeError', function () {
+    const bsonTypeErr = new BSONTypeError();
+    expect(bsonTypeErr instanceof BSONTypeError).to.be.true;
```
dariakp, line 10: seems to be missing `instanceof TypeError` and the case below is missing `instanceof Error` check

dariakp (changes requested):

dariakp (approved):

PR #472, opened 2021-11-12
feat(NODE-3740): Implement root and top level key utf-8 validation settings for BSON

gjchong25:
### Description
Adds a new validation option in DeserializeOptions that allows all or a specific subset of top-level keys in a document to be excluded from strict utf-8 validation and instead applies the replacement character behavior. Tests are added to make sure that both global and specific-key subsets can have utf8 validation enabled and disabled. 

Default behavior is to have strict utf-8 validation (an error will be thrown if invalid utf8 is detected). This change gives users the ability to opt into character replacement behavior instead of strict utf-8 validation. 

### DIFF
--- a/src/parser/deserializer.ts
+++ b/src/parser/deserializer.ts
@@ -45,7 +45,7 @@ export interface DeserializeOptions {
   index?: number;
 
   raw?: boolean;
-  /** allows for opt-out utf-8 validation for all keys or
+  /** Allows for opt-out utf-8 validation for all keys or
    * specified keys. Must be all true or all false.
    *
    * @example
@@ -53,11 +53,14 @@ export interface DeserializeOptions {
    * // disables validation on all keys
    *  validation: { utf8: false }
    *
-   * // enables validation on specified keys a, b, and c
+   * // enables validation only on specified keys a, b, and c
    *  validation: { utf8: { a: true, b: true, c: true } }
+   *
+   *  // disables validation only on specified keys a, b
+   *  validation: { utf8: { a: false, b: false } }
    * ```
    */
-  validation?: { utf8: boolean | Record<string, boolean> };
+  validation?: { utf8: boolean | Record<string, true> | Record<string, false> };
 }
 
 // Internal long versions
@@ -115,8 +118,7 @@ function deserializeObject(
   buffer: Buffer,
   index: number,
   options: DeserializeOptions,
-  isArray = false,
-  nestedKey?: boolean
+  isArray = false
 ) {
   const evalFunctions = options['evalFunctions'] == null ? false : options['evalFunctions'];
   const cacheFunctions = options['cacheFunctions'] == null ? false : options['cacheFunctions'];
@@ -146,32 +148,31 @@ function deserializeObject(
 
   // Check for boolean uniformity and empty validation option
   const utf8ValidatedKeys = validation.utf8;
-  if (typeof utf8ValidatedKeys !== 'boolean') {
+  if (typeof utf8ValidatedKeys === 'boolean') {
+    validationSetting = utf8ValidatedKeys;
+  } else {
     globalUTFValidation = false;
     const utf8ValidationValues = Object.keys(utf8ValidatedKeys).map(function (key) {
       return utf8ValidatedKeys[key];
     });
-    if (utf8ValidationValues.length !== 0) {
-      // Ensures boolean uniformity in utf-8 validation (all true or all false)
-      if (typeof utf8ValidationValues[0] === 'boolean') {
-        validationSetting = utf8ValidationValues[0];
-        if (!utf8ValidationValues.every(item => item === validationSetting)) {
-          throw new BSONError(
-            'Invalid UTF-8 validation option - keys must be all true or all false'
-          );
-        }
-      } else {
-        throw new BSONError('Invalid UTF-8 validation option, must specify boolean values');
-      }
-    } else {
-      throw new BSONError('validation option is empty');
+    if (utf8ValidationValues.length === 0) {
+      throw new BSONError('UTF-8 validation setting cannot be empty');
     }
-  } else {
-    validationSetting = utf8ValidatedKeys;
+    // if (typeof utf8ValidationValues[0] !== 'boolean') {
+    //   throw new BSONError('Invalid UTF-8 validation option, must specify boolean values');
+    // }
+    if (!utf8ValidationValues.every(item => typeof item === 'boolean')) {
+      throw new BSONError('Invalid UTF-8 validation option, must specify boolean values');
+    }
+    validationSetting = utf8ValidationValues[0];
+    // Ensures boolean uniformity in utf-8 validation (all true or all false)
+    if (!utf8ValidationValues.every(item => item === validationSetting)) {
+      throw new BSONError('Invalid UTF-8 validation option - keys must be all true or all false');
+    }
   }
 
   // Add keys to set that will either be validated or not based on validationSetting
-  if ((!validationSetting && !globalUTFValidation) || (validationSetting && !globalUTFValidation)) {
+  if (!globalUTFValidation) {
     for (const key of Object.keys(utf8ValidatedKeys)) {
       utf8KeysSet.add(key);
     }
@@ -219,20 +220,16 @@ function deserializeObject(
     // Represents the key
     const name = isArray ? arrayIndex++ : buffer.toString('utf8', index, i);
 
-    // keyValidate is true if the key should be validated, false otherwise
-    let keyValidate = true;
+    // shouldValidateKey is true if the key should be validated, false otherwise
+    let shouldValidateKey = true;
     if (globalUTFValidation) {
-      keyValidate = validationSetting;
+      shouldValidateKey = validationSetting;
     } else {
       if (utf8KeysSet.has(name)) {
-        keyValidate = validationSetting;
+        shouldValidateKey = validationSetting;
       } else {
-        keyValidate = !validationSetting;
+        shouldValidateKey = !validationSetting;
       }
-    }
-    // if nested key, validate based on top level key
-    if (nestedKey != null) {
-      keyValidate = nestedKey;
     }
 
     if (isPossibleDBRef !== false && (name as string)[0] === '$') {
@@ -255,7 +252,7 @@ function deserializeObject(
       ) {
         throw new BSONError('bad string length in bson');
       }
-      value = getValidatedString(buffer, index, index + stringSize - 1, validation, keyValidate);
+      value = getValidatedString(buffer, index, index + stringSize - 1, shouldValidateKey);
       index = index + stringSize;
     } else if (elementType === constants.BSON_DATA_OID) {
       const oid = Buffer.alloc(12);
@@ -308,7 +305,8 @@ function deserializeObject(
       if (raw) {
         value = buffer.slice(index, index + objectSize);
       } else {
-        value = deserializeObject(buffer, _index, options, false, keyValidate);
+        options.validation = { utf8: shouldValidateKey };
+        value = deserializeObject(buffer, _index, options, false);
       }
 
       index = index + objectSize;
@@ -336,8 +334,8 @@ function deserializeObject(
         }
         arrayOptions['raw'] = true;
       }
-
-      value = deserializeObject(buffer, _index, arrayOptions, true, keyValidate);
+      arrayOptions.validation = { utf8: shouldValidateKey };
+      value = deserializeObject(buffer, _index, arrayOptions, true);
       index = index + objectSize;
 
       if (buffer[index - 1] !== 0) throw new BSONError('invalid array terminator byte');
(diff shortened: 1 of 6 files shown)


src/parser/deserializer.ts
```diff
@@ -45,6 +45,8 @@ export interface DeserializeOptions {
   index?: number;
 
   raw?: boolean;
+  /** allows for opt-in utf-8 validation */
+  validation?: Document;
```
nbbeeken, line 49:
```suggestion
  validation?: { utf8: Record<string, boolean> };
```
Isn't it a nested utf8 key that points to the key->boolean? 
Just to add a bit stricter typing.

src/parser/deserializer.ts
```diff
@@ -120,6 +123,47 @@ function deserializeObject(
   const promoteLongs = options['promoteLongs'] == null ? true : options['promoteLongs'];
   const promoteValues = options['promoteValues'] == null ? true : options['promoteValues'];
 
+  // Ensures default validation option if none given
+  const validation = options['validation'] == null ? { utf8: true } : options['validation'];
```
nbbeeken, line 127:
```suggestion
  const validation = options.validation == null ? { utf8: true } : options.validation;
```

src/parser/deserializer.ts
```diff
@@ -120,6 +123,47 @@ function deserializeObject(
...
+  // Reflects utf8 validation boolean regardless of global or specific key validation
+  let uniformBool: boolean;
+  // Set of keys either to enable or disable validation on
+  const utf8KeysSet = new Set();
+
+  // Check for boolean uniformity and empty validation option
+  const keys = validation.utf8;
+  if (typeof keys !== 'boolean') {
+    globalUTFValidation = false;
+    const vals = Object.keys(keys).map(function (key) {
```
nbbeeken, line 140: nit: `vals` -> `values` or even more desciptive `utf8ValidationValues`

src/parser/deserializer.ts
```diff
@@ -120,6 +123,47 @@ function deserializeObject(
...
+
+  // Shows if global utf8 validation is enabled or disabled
+  let globalUTFValidation = true;
+  // Reflects utf8 validation boolean regardless of global or specific key validation
+  let uniformBool: boolean;
+  // Set of keys either to enable or disable validation on
+  const utf8KeysSet = new Set();
+
+  // Check for boolean uniformity and empty validation option
+  const keys = validation.utf8;
```
nbbeeken, line 137: Lower in this function this variable could be confused with the actual document keys so lets try: `utf8ValidatedKeys`

src/parser/deserializer.ts
```diff
@@ -120,6 +123,47 @@ function deserializeObject(
...
+    } else {
+      throw new BSONError('validation option is empty');
+    }
+  } else {
+    uniformBool = keys;
+  }
+
+  // Add keys to set that will either be validated or not based on uniformBool
+  if ((!uniformBool && !globalUTFValidation) || (uniformBool && !globalUTFValidation)) {
+    Object.keys(keys).forEach(key => utf8KeysSet.add(key));
```
nbbeeken, line 164: Super nit, in general we have preferred using js `for` loops as opposed to `.forEach`.

test/node/utf8_tests.js
```diff
@@ -0,0 +1,343 @@
+'use strict';
+
+const { Buffer } = require('buffer');
+const BSON = require('../register-bson');
+const { isNode6, isBrowser } = require('./tools/utils');
+const BSONError = BSON.BSONError;
+
+describe('UTF8 validation', function () {
+  // Test both browser shims and node which have different replacement mechanisms
+  const replacementChar = isNode6() || isBrowser() ? '���' : '�';
```
nbbeeken, line 10:
```suggestion
  const replacementChar = isNode6() || isBrowser() ? '\u{FFFD}\u{FFFD}\u{FFFD}' : '\u{FFFD}';
```

I think this just calls out the code point better for the next reader, just a suggestion.

test/node/tools/utils.js
```diff
@@ -125,3 +125,33 @@ const bufferFromHexArray = array => {
...
+ * @param string - representing what you want to encode into BSON
+ * @returns BSON string with byte size encoded
+ */
+const stringToUTF8HexBytes = str => {
+  var b = Buffer.from(str, 'utf8');
+  var len = b.byteLength;
+  var out = Buffer.alloc(len + 4 + 1);
+  out.writeInt32LE(len + 1, 0);
+  out.set(b, 4);
+  return out.toString('hex');
```
nbbeeken, line 144:
```suggestion
  out[len+1] = 0x00;
  return out.toString('hex');
```

The buffer is zero-ed out so this isn't exactly necessary but just to be sure the null byte is there at the end seems like a good idea.

nbbeeken (changes requested):
Some initial feedback!

Side note: can you run `npm install` and push the package-lock changes, looks like its not sync-ed up with main from the last release.

src/parser/deserializer.ts
gjchong25: Should I change this for the other options as well?

src/parser/deserializer.ts
```diff
@@ -45,6 +45,8 @@ export interface DeserializeOptions {
   index?: number;
 
   raw?: boolean;
+  /** allows for opt-in utf-8 validation */
```
nbbeeken, line 48: Let's expand this documentation a bit more. First off, IIRC this allows for opt-out, since the default is to always validate. Also, I think the type should be `utf8: boolean | Record<string, boolean>` right? cus it can just take an overall setting? Additionally, we could use some examples of how to use this option correctly.

(You can use the `@example` tsdoc block for this!)

src/parser/deserializer.ts
```diff
@@ -158,7 +204,26 @@ function deserializeObject(
...
     const name = isArray ? arrayIndex++ : buffer.toString('utf8', index, i);
+
+    // keyValidate is true if the key should be validated, false otherwise
+    let keyValidate = true;
+    if (globalUTFValidation) {
+      keyValidate = uniformBool;
+    } else {
+      if (utf8KeysSet.has(name)) {
+        keyValidate = uniformBool;
+      } else if (!utf8KeysSet.has(name)) {
```
nbbeeken, line 218: We can just use `else` here

src/parser/deserializer.ts
```diff
@@ -120,6 +123,49 @@ function deserializeObject(
   const promoteLongs = options['promoteLongs'] == null ? true : options['promoteLongs'];
   const promoteValues = options['promoteValues'] == null ? true : options['promoteValues'];
 
+  // Ensures default validation option if none given
+  const validation = options.validation == null ? { utf8: true } : options.validation;
+
+  // Shows if global utf8 validation is enabled or disabled
+  let globalUTFValidation = true;
+  // Reflects utf8 validation boolean regardless of global or specific key validation
+  let uniformBool: boolean;
```
nbbeeken, line 132: I think we need a new name for this, it is indeed the boolean calculated from looking at all the ones specified but when reading the code lower down I'm losing the meaning, this is whether or not we are by default : validating or not validating. Maybe `validationSetting`? its like more generic but kinda descriptive of the purpose?

nbbeeken (changes requested):

nbbeeken (changes requested):
whoops accidentally cleared my review state

nbbeeken:
I'm going to try and investigate the strange rollup issue in CI tomorrow, but we can manually verify this works as expected in the browser if it comes to it.

src/parser/deserializer.ts
```diff
@@ -45,6 +45,19 @@ export interface DeserializeOptions {
...
+   * @example
+   * ```js
+   * // disables validation on all keys
+   *  validation: { utf8: false }
+   *
+   * // enables validation on specified keys a, b, and c
+   *  validation: { utf8: { a: true, b: true, c: true } }
+   * ```
+   */
+  validation?: { utf8: boolean | Record<string, boolean> };
```
dariakp, line 60: We could codify the all true or all false requirement in typescript:
```suggestion
  validation?: { utf8: boolean | Record<string, true> |  Record<string, false> };
```

src/parser/deserializer.ts
```diff
@@ -45,6 +45,19 @@ export interface DeserializeOptions {
   index?: number;
 
   raw?: boolean;
+  /** allows for opt-out utf-8 validation for all keys or
+   * specified keys. Must be all true or all false.
```
dariakp, line 49: Let's also add type tests for this property (we probably want to add a new file, since the existing one just tests the actual BSON types; this one would test input to deserialize)

dariakp (changes requested):
Still working on reviewing the rest of the code, but wanted to capture this small suggestion

src/parser/deserializer.ts
nbbeeken: Sure! that would be a nice improvement

src/parser/deserializer.ts
```diff
@@ -670,14 +764,25 @@ function isolateEval(
   return functionCache[functionString].bind(object);
 }
 
-function getValidatedString(buffer: Buffer, start: number, end: number) {
+function getValidatedString(
+  buffer: Buffer,
+  start: number,
+  end: number,
+  validation: Document,
```
dariakp, line 771: I think this method should only care about whether or not to validateUtf8, so both of these can be condensed to a single `validateUtf8: boolean`; then lines 776 and 777 get condensed to `if (validateUtf8)`

src/parser/deserializer.ts
```diff
@@ -670,14 +764,25 @@ function isolateEval(
...
+  check: boolean
+) {
   const value = buffer.toString('utf8', start, end);
-  for (let i = 0; i < value.length; i++) {
-    if (value.charCodeAt(i) === 0xfffd) {
-      if (!validateUtf8(buffer, start, end)) {
-        throw new BSONError('Invalid UTF-8 string in BSON document');
+  // if utf8 validation is on, do the check
+  if (check) {
+    if (validation.utf8 != null && validation.utf8) {
```
dariakp, line 777: `validation.utf8` will only return truthy if it is `!= null`, so the first check is redundant

test/node/utf8_tests.js
```diff
@@ -0,0 +1,343 @@
...
+  });
+
+  it('should correctly handle validation if validation option contains all T or all F with valid utf8 example', function () {
+    let allTrue = { validation: { utf8: { a: true, b: true, c: true } } };
+    let allFalse = { validation: { utf8: { a: false, b: false, c: false, d: false } } };
+    expect(() => BSON.deserialize(sampleValidUTF8, allTrue)).to.not.throw();
+    expect(() => BSON.deserialize(sampleValidUTF8, allFalse)).to.not.throw();
+  });
+
+  it('should throw error if empty utf8 validation option passed in', function () {
```
dariakp, line 39: I think we're missing a test case for when a non-boolean validation option is specified for a key

dariakp (changes requested):
Finished reviewing the code, still need to double check the tests

src/parser/deserializer.ts
```diff
@@ -234,6 +305,7 @@ function deserializeObject(
       if (raw) {
         value = buffer.slice(index, index + objectSize);
       } else {
+        options.validation = { utf8: shouldValidateKey };
```
dariakp, line 308: in order to avoid mutating the options object, let's do something like:
```
let objectOptions = options;
if (!globalUTFValidation) { 
   objectOptions = { ...options, validation: { utf8: shouldValidateKey }  }
}
```
and then let's make a similar change to wrap the arrayOptions (no need to declare a new var though, just:
```
if (!globalUTFValidation) { 
   arrayOptions = { ...options, validation: { utf8: shouldValidateKey }  }
}
```

dariakp (changes requested):
Just a couple of things left to clean up

test/types/deserialize.test-d.ts
```diff
@@ -0,0 +1,17 @@
...
+});
+
+expectError(deserialize(sampleValidUTF8, { validation: { utf8: { a: false, b: true } } }));
+expectError(deserialize(sampleValidUTF8, { validation: { utf8: { a: true, b: true, c: false } } }));
+
+// all true and all false validation utf8 options are valid
+deserialize(sampleValidUTF8, { validation: { utf8: { a: true, b: true, c: true } } });
+deserialize(sampleValidUTF8, { validation: { utf8: { a: false, b: false, c: false} } });
+deserialize(sampleValidUTF8, { validation: { utf8: true } });
+deserialize(sampleValidUTF8, { validation: { utf8: true } });
```
dariakp, line 17: Missed one
```suggestion
deserialize(sampleValidUTF8, { validation: { utf8: false } });
```

dariakp (changes requested):

nbbeeken (dismissed):
LGTM!!!

src/parser/deserializer.ts
```diff
@@ -262,7 +330,9 @@ function deserializeObject(
         }
         arrayOptions['raw'] = true;
       }
-
+      if (!globalUTFValidation) {
+        arrayOptions = { ...options, validation: { utf8: shouldValidateKey } };
```
dariakp, line 334:
```suggestion
        arrayOptions = { ...arrayOptions, validation: { utf8: shouldValidateKey } };
```
Just noticed there is some logic that modifies these, not sure if it really does anything, but better preserve it

dariakp (changes requested):

dariakp (approved):

PR #473, opened 2021-11-16
fix(NODE-3760): Add regex check to 12 length string in isValid

MrEmanuel:
Issue with description filed here.
https://jira.mongodb.org/browse/DRIVERS-1989

### Description
Check the validity of strings with length 12 in the same way as is done for 24 length strings.

#### What is changing?
Added regex check to 12 length strings in ObjectId.isValid()

No.

Object.isValid() returns true for strings that it shoudn't. For example the string: ´Value with é`

### DIFF
--- a/src/objectid.ts
+++ b/src/objectid.ts
@@ -298,7 +298,7 @@ export class ObjectId {
     }
 
     if (typeof id === 'string') {
-      return id.length === 12 || (id.length === 24 && checkForHexRegExp.test(id));
+      return (id.length === 12 || id.length === 24) && checkForHexRegExp.test(id));
     }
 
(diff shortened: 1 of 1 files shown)


dariakp:
@MrEmanuel Hi there, thank you for reaching out about this issue. We moved the ticket you created to the NODE driver project, you can follow it here: https://jira.mongodb.org/browse/NODE-3760. It looks like we missed a validation case in our most recent `ObjectId` constructor validation improvement, so we'll look into adding it there as well as adding corresponding logic to `isValid()`.

nbbeeken:
Resolved in https://github.com/mongodb/js-bson/pull/475

PR #475, opened 2021-12-01
fix(NODE-3760): ObjectId.isValid string and byte length match

syz99:
#### What is changing?
adds validation for string and byte length matching for new ObjectID's.

bugs found with strings of char-length 12, but byte-length =/= 12.

### DIFF
--- a/src/objectid.ts
+++ b/src/objectid.ts
@@ -45,7 +45,7 @@ export class ObjectId {
    *
    * @param inputId - Can be a 24 character hex string, 12 byte binary Buffer, or a number.
    */
-  constructor(inputId?: string | Buffer | number | ObjectIdLike | ObjectId) {
+  constructor(inputId?: string | number | ObjectId | ObjectIdLike | Buffer | Uint8Array) {
     if (!(this instanceof ObjectId)) return new ObjectId(inputId);
 
     // workingId is set based on type of input and whether valid id exists for the input
@@ -290,38 +290,15 @@ export class ObjectId {
    *
    * @param id - ObjectId instance to validate.
    */
-  static isValid(id: number | string | ObjectId | Uint8Array | ObjectIdLike): boolean {
+  static isValid(id: string | number | ObjectId | ObjectIdLike | Buffer | Uint8Array): boolean {
     if (id == null) return false;
 
-    if (typeof id === 'number') {
+    try {
+      new ObjectId(id);
       return true;
+    } catch {
+      return false;
     }
-
-    if (typeof id === 'string') {
-      if (id.length === 12) {
-        const bytes = Buffer.from(id);
-        return bytes.byteLength === 12;
-      }
-      return id.length === 24 && checkForHexRegExp.test(id);
-    }
-
-    if (id instanceof ObjectId) {
-      return true;
-    }
-
-    if (isUint8Array(id) && id.length === 12) {
-      return true;
-    }
-
-    // Duck-Typing detection of ObjectId like objects
-    if (typeof id === 'object' && 'toHexString' in id && typeof id.toHexString === 'function') {
-      if (typeof id.id === 'string') {
-        return id.id.length === 12;
-      }
-      return id.toHexString().length === 24 && checkForHexRegExp.test(id.id.toString('hex'));
-    }
-
-    return false;
   }
 
   /** @internal */
--- a/src/objectid.ts
+++ b/src/objectid.ts
@@ -298,7 +298,12 @@ export class ObjectId {
     }
 
     if (typeof id === 'string') {
-      return id.length === 12 || (id.length === 24 && checkForHexRegExp.test(id));
+      if (id.length === 12) {
+        const bytes = Buffer.from(id);
+        return bytes.byteLength === 12;
+      } else if (id.length === 24) {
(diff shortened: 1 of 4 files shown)


test/mocha.opts
```diff
@@ -2,4 +2,3 @@
 --require chai/register-expect
 --require source-map-support/register
 --timeout 10000
---throw-deprecation
```
nbbeeken, line 5: This prevents a crash on node 16, chai makes use of a deprecated "exports" field in its package.json

src/objectid.ts
```diff
@@ -298,7 +298,12 @@ export class ObjectId {
     }
 
     if (typeof id === 'string') {
-      return id.length === 12 || (id.length === 24 && checkForHexRegExp.test(id));
+      if (id.length === 12) {
+        const bytes = Buffer.from(id);
+        return bytes.byteLength === 12;
+      } else if (id.length === 24) {
+        return checkForHexRegExp.test(id);
+      }
```
nbbeeken, line 306: There's a chance we fall through here where we didn't before, it will still end up making it to the false at the end of the function but I think we can keep this concise:

```suggestion
      if (id.length === 12) {
        const bytes = Buffer.from(id);
        return bytes.byteLength === 12;
      }
      return id.length === 24 && checkForHexRegExp.test(id);
```

This way we fast track to false if we don't have either 12 nor a 24 length string.

nbbeeken (changes requested):

src/objectid.ts
syz99: I did this earlier and it failed a BSON test case, when I explicitly checked the length first it passed

syz99: to be specific, the "Should fail to create ObjectId due to illegal hex code" test case in `bson_test`

nbbeeken (approved):
LGTM

dariakp (changes requested):
I see that this check is only in the isValid method, what we discussed regarding the implementation when we triaged this is in the comment on the ticket:

> what we need to do is add logic to catch this case in the constructor and then also add corresponding logic in the isValid check so that isValid uses the same criteria as the constructor to determine validity

@nbbeeken do you guys by any chance have some sort of kickoff notes about what is and isn't in scope for the implementation? because without unifying constructor logic with the validation logic we are just making a bandaid fix without actually improving the validation/constructor duality

syz99:
@dariakp 

Added retroactive kickoff notes; Grace's fix already addresses the user's specific issue in constructor, we're now syncing `isValid` to match: 

> Issue with constructor was fixed in NODE-3662; we can instead fix isValid to match the constructor's behavior. User's given example doesn't have the behavior mentioned where .id is undefined. This work should correct isValid to return true/false based on constructor's corrected behavior.

dariakp:
@syz99 thanks for double checking! I am still wondering though if we can consolidate the validation logic so that we don't have to rely on duplicating the correct validation blocks in the isValid function; have we consider if there can be a general validation method that can be used in both places or even the approach of calling the constructor to create an ObjectId?

baileympearson:
> @syz99 thanks for double checking! I am still wondering though if we can consolidate the validation logic so that we don't have to rely on duplicating the correct validation blocks in the isValid function; have we considered if there can be a general validation method that can be used in both places or even the approach of calling the constructor to create an ObjectId in the isValid function?

👍 to the idea of removing duplicate logic when possible!  @dariakp Are you wondering about replacing the entirety of the `isValid` method with a call to the `ObjectID` constructor or just the branches where `id` is a string?

dariakp:
> > @syz99 thanks for double checking! I am still wondering though if we can consolidate the validation logic so that we don't have to rely on duplicating the correct validation blocks in the isValid function; have we considered if there can be a general validation method that can be used in both places or even the approach of calling the constructor to create an ObjectId in the isValid function?
> 
> 👍 to the idea of removing duplicate logic when possible! @dariakp Are you wondering about replacing the entirety of the `isValid` method with a call to the `ObjectID` constructor or just the branches where `id` is a string?

The former, try/catch around the call and return true/false accordingly; it's the most straightforward way to maintain parity given how the validation is split up; it'll be less ugly if we break out the overloads into separate functions down the road for the next major versions, but for now there aren't many options to make it pretty aside from the aforementioned one

dariakp (approved):
Yay for less code and more consistency 🎉

src/objectid.ts
```diff
@@ -290,34 +290,15 @@ export class ObjectId {
    *
    * @param id - ObjectId instance to validate.
    */
-  static isValid(id: number | string | ObjectId | Uint8Array | ObjectIdLike): boolean {
+  static isValid(id: string | number | ObjectId | ObjectIdLike | Buffer | Uint8Array): boolean {
```
baileympearson, line 293: Was adding support for `Uint8Array`s intentional?  I notice it's added here and in the constructor

syz99: I think so -- Since we have parity between `isValid` and the `constructor`, the constructor now needs to take in an `Uint8Array` in order for it to be called from `isValid`. Buffer is a subclass of `Uint8Array`, so nothing will be broken by allowing the more general `Uint8Array` as a param

baileympearson (approved):
LGTM - nice work!

PR #477, opened 2021-12-16
fix(NODE-3821): nullish check before using toBSON override function

CPLepage:
### Description
In the serializer, when it comes to check for an override value (`toBSON`), changing  the `value` check to compare with `null` and `undefined` instead of relying on the truthiness of `value`.

Cannot extend BigInt prototype with`.toBSON()` since `0n === false`

trying to make this work
```ts
BigInt.prototype.toBSON = function() {
    return Long.fromBits(Number(this.valueOf() & BigInt(0xffffffff)),
        Number(this.valueOf() >> BigInt(32)))
}

// src/parser/serializer.ts:769
// Is there an override value
if(value && value.toBSON){
    // this won't go through if value === 0n
    ...
}

...

// then trows error here since value hasn't been overidden
// src/parser/serializer.ts:779
else if (typeof value === 'bigint') {
    throw new BSONTypeError('Unsupported type BigInt, please use Decimal128');
}
```

### DIFF
--- a/src/parser/serializer.ts
+++ b/src/parser/serializer.ts
@@ -958,8 +958,7 @@ export function serializeInto(
     for (const key in object) {
       let value = object[key];
       // Is there an override value
-      if (value !== null && value !== undefined && value.toBSON) {
-        if (typeof value.toBSON !== 'function') throw new BSONTypeError('toBSON is not a function');
+      if (typeof value?.toBSON === 'function') {
         value = value.toBSON();
       }
 
--- a/src/parser/serializer.ts
+++ b/src/parser/serializer.ts
@@ -767,8 +767,7 @@ export function serializeInto(
       let value = object[i];
 
       // Is there an override value
-      if (value !== null && value !== undefined && value.toBSON) {
-        if (typeof value.toBSON !== 'function') throw new BSONTypeError('toBSON is not a function');
+      if (typeof value?.toBSON === 'function') {
         value = value.toBSON();
       }
 
--- a/src/parser/serializer.ts
+++ b/src/parser/serializer.ts
@@ -946,12 +946,12 @@ export function serializeInto(
       }
     }
   } else {
-    // Did we provide a custom serialization method
-    if (object.toBSON) {
-      if (typeof object.toBSON !== 'function') throw new BSONTypeError('toBSON is not a function');
+    if (typeof object?.toBSON === 'function') {
+      // Provided a custom serialization method
       object = object.toBSON();
-      if (object != null && typeof object !== 'object')
+      if (object != null && typeof object !== 'object') {
         throw new BSONTypeError('toBSON function did not return an object');
+      }
     }
 
     // Iterate over all the keys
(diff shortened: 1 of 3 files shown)


src/parser/serializer.ts
```diff
@@ -767,7 +767,7 @@ export function serializeInto(
       let value = object[i];
 
       // Is there an override value
-      if (value && value.toBSON) {
+      if (value !== null && value !== undefined && value.toBSON) {
         if (typeof value.toBSON !== 'function') throw new BSONTypeError('toBSON is not a function');
         value = value.toBSON();
       }
```
nbbeeken, line 773:
```suggestion
      if (typeof value?.toBSON === 'function') {
        value = value.toBSON();
      }
```

We can use the optional chaining syntax here to be a little more succinct, null and undefined will not be accessed with a dot so `typeof` will return `'undefined'` so the if statement won't be entered.

This is what it compiles to, just so we can all double check this is what's expected.
```js
typeof (value === null || value === undefined ? undefined : value.toBSON) === 'function'
```

src/parser/serializer.ts
```diff
@@ -959,7 +959,7 @@ export function serializeInto(
     for (const key in object) {
       let value = object[key];
       // Is there an override value
-      if (value && value.toBSON) {
+      if (value !== null && value !== undefined && value.toBSON) {
         if (typeof value.toBSON !== 'function') throw new BSONTypeError('toBSON is not a function');
         value = value.toBSON();
       }
```
nbbeeken, line 965:
```suggestion
      if (typeof value?.toBSON === 'function') {
        value = value.toBSON();
      }
```

nbbeeken (changes requested):
Hi @CPLepage, thanks so much for contributing here! I think this is something we would certainly want to fix. I've gone ahead and created a JIRA ticket to track the work and I have some suggestions of improvements for ya. If you get the chance adding a test of your usecase to: `test/node/to_bson_test.js` would be a great help. 

@dariakp @durran @baileympearson, Just wanna get y'all's attention here I'm suggesting we drop the throw case with toBSON not being of type 'function'. While very unlikely it seems wrong that our BSON library would not allow the document: `{ toBSON: 1 }` to be serialized. That being said we could easily keep it, and revisit that at a later time.

CPLepage:
@nbbeeken good idea on the optional chaining, looks clean!<br />I added 2 tests on the current particular case. Let me know your thoughts

nbbeeken:
Hey @CPLepage do you mind if I push up some changes of my own here? due to our support for older node versions the testing arrangement can be kinda strange, I can make sure we still get bigint coverage on later versions of node 🙂

CPLepage:
@nbbeeken please go ahead! Tried a few things for the tests, but CI never passed

nbbeeken (approved):
@CPLepage  Let me know if you see anything I missed, I'll ask the team for a final look now 🙂  thanks again!

CPLepage:
@nbbeeken Looks good! Thanks to you for the quick replies!!

baileympearson (approved):

dariakp (approved):

PR #486, opened 2022-02-09
test(NODE-3719): reorganize spec tests and remove unneeded skips

dariakp:
### Description
NODE-3719

#### What is changing?
Spec test organization

N/A

Spec compliance review


test/node/bson_corpus.prose.test.js
```diff
@@ -0,0 +1,41 @@
...
+
+    it('Pattern for a regular expression', () => {
+      // eslint-disable-next-line no-control-regex
+      expect(() => BSON.serialize({ a: new RegExp('a\x00b') })).to.throw(/null bytes/);
+      expect(() => BSON.serialize({ a: new BSONRegExp('a\x00b') })).to.throw(/null bytes/);
+    });
+
+    it('Flags/options for a regular expression', () => {
+      expect(() => BSON.serialize({ a: new BSONRegExp('a', 'i\x00m') })).to.throw(/null bytes/);
+      // TODO: should we test RegExp, too?
```
dariakp, line 38: I don't remember if there is a reason why we test RegExp in the block above but not in this one

dariakp (commented):
Along with the comment below:
1. I noticed that specs/object-id might be dead code, I haven't been able to find any reference to vectors.json in the actual specs repo and nothing seems to be loading these ones
2. decodeErrors only checks `expect(() => BSON.deserialize(B, deserializeOptions)).to.throw();` - is it intentionally general or should we be checking for a better error? If so, NODE-3637 may be a good ticket to tack that work onto
3. There are a number of skipped tests with comments but no TODOs, it would be good to either definitively say "we'll never implement this", or to file actual backlog tickets to make sure that if we do implement relevant functionality, those tests get unskipped
```js
const skipBSON = {
  'NaN with payload':
    'passing this would require building a custom type to store the NaN payload data.'
};
// tests from the corpus that we need to skip, and explanations why
const skipExtendedJSON = {
  'Timestamp with high-order bit set on both seconds and increment':
    'Current BSON implementation of timestamp/long cannot hold these values - 1 too large.',
  'Timestamp with high-order bit set on both seconds and increment (not UINT32_MAX)':
    'Current BSON implementation of timestamp/long cannot hold these values - 1 too large.'
};

const SKIP_TESTS = new Map([
  ...Object.entries(skipBSON),
  ...Object.entries(skipExtendedJSON),
  [
    'All BSON types',
    'there is just too much variation in the specified expectation to make this work'
  ]
]);
```

test/node/bson_corpus.prose.test.js
nbbeeken: You can't put null bytes into the native regex constructor, it's a SyntaxError so we're covered 😎 

```js
> new RegExp('a', '\x00i')
Uncaught SyntaxError: Invalid flags supplied to RegExp constructor 'i'
> new RegExp('a', 'i\x00')
Uncaught SyntaxError: Invalid flags supplied to RegExp constructor 'i'
```

nbbeeken (approved):

nbbeeken:
CI failures aren't a blocker, I've opened https://github.com/mongodb/js-bson/pull/487 to address some issues.

durran (approved):

baileympearson (approved):

PR #494, opened 2022-03-23
chore: add typesversion to package.json && bump downlevel-dts version

baileympearson:
#### What is changing?

- downlevel-dts is bumped to the newest version
- a `typesVersion` is added to the package.json, so users on newer versions of TS are not using the output of downlevel-dts

No.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,13 @@
     "bower.json"
   ],
   "types": "bson.d.ts",
+  "typesVersions": {
+    "<=4.0.2": {
+      "bson.d.ts": [
+        "bson.ts34.d.ts"
+      ]
+    }
+  },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -15,6 +15,13 @@
     "bower.json"
```
nbbeeken, line 15: We need to add the new type definition file to this list

nbbeeken (changes requested):

package.json
baileympearson:
```suggestion
    "bson.ts34.d.ts"
```

nbbeeken (approved):

dariakp (approved):

PR #432, opened 2021-04-22
fix: accept Uint8Array where Buffer is accepted

addaleax:
Also fixes NODE-3223 (ensureBuffer ignores byteLength/byteOffset).
This is the "proper" alternative to https://github.com/mongodb/js-bson/pull/418 and matches what e.g. Node.js APIs do.

## Description

**What changed?**

PR #498, opened 2022-04-27
perf(NODE-4194): improve objectId constructor performance

nbbeeken:
#### What is changing?

Gating ensureBuffer with an instanceof check should avoid the reletively expensive `Buffer.from` call. 

Improve the performance of creating an ObjectId from a buffer.


nbbeeken:
@addaleax This change would revert a bits of #429 and #432. The thinking is that the instanceof check can succeed for nodejs environments dodging the Buffer.from call, and in environments where the prototype chain doesn't match we still fall back to calling `.from`. Any thoughts on this?

src/objectid.ts
```diff
@@ -72,7 +72,7 @@ export class ObjectId {
       // Generate a new id
       this[kId] = ObjectId.generate(typeof workingId === 'number' ? workingId : undefined);
     } else if (ArrayBuffer.isView(workingId) && workingId.byteLength === 12) {
-      this[kId] = ensureBuffer(workingId);
+      this[kId] = workingId instanceof Buffer ? workingId : ensureBuffer(workingId);
```
addaleax, line 75: Is it intentional that this check is added both here and inside of `ensureBuffer`?

src/ensure_buffer.ts
```diff
@@ -11,6 +11,10 @@ import { isAnyArrayBuffer } from './parser/utils';
  * @throws BSONTypeError If anything other than a Buffer or Uint8Array is passed in
  */
 export function ensureBuffer(potentialBuffer: Buffer | ArrayBufferView | ArrayBuffer): Buffer {
+  if (potentialBuffer instanceof Buffer) {
```
addaleax, line 14: It might be worth leaving a comment here about why this extra check is added :)

addaleax (commented):
@nbbeeken I don’t see any downsides in terms of correctness. I remember that `instanceof` could be quite slow on some older-but-not-that-old V8 versions, so it might be worth checking the performance impact across Node.js versions.

In the long run, I think the best solution would be to just use Uint8Array everywhere anyway :)

src/objectid.ts
nbbeeken: Yea with my intention being to avoid the overhead of even calling `ensureBuffer`

src/ensure_buffer.ts
nbbeeken: done!

src/ensure_buffer.ts
```diff
@@ -11,6 +11,11 @@ import { isAnyArrayBuffer } from './parser/utils';
...
```
Uzlopak, line 14: Why is it an instanceof check?

Imho should be 

```suggestion
  if (Buffer.isBuffer(potentialBuffer)) {
```

Uzlopak: Also js-bson uses npm package buffer, which also provides Buffer.isBuffer.

nbbeeken: Hey @Uzlopak we chose to use `instanceof` here because of it's sensitivity to buffers passed from other environments. In nodejs this will always return true so we can optimize for that case. In browsers if a different instance is passed to use we can be sure we align it with the type and version of buffer that our library expects.

Uzlopak: Can you please elaborate the Browser case? I mean instanceof would return true only if it is an instance of the buffer class which is used by the js-bson package. So if you would pass a buffer of a different class, it would also fail. Or am I wrong?

nbbeeken: Sure yea, it would return false for other implementations of the class which is the desired behavior because it will wrap the passed buffer in the exact version we depend on. It is possible for other versions of the `Buffer` polyfill to be included in the same bundle or `Buffer`s from across Realm boundaries to be passed to our APIs.
Here we want to make a minimal change the alleviates a performance cost on node that isn't necessary since it's the expected type already and there's no need to wrap the type.

durran (changes requested):
My only comment would be to up the iterations in the benchmark to get a more accurate representation of performance once it's warmed up. Here are the differences on my machine on 10.000 vs 1.000.000 iterations - it's far less of a discrepancy.

```
Iterations: 10000

BSON@current local
deserialize({ oid, string }, { validation: { utf8: false } }) takes 0.00283865ms on average
new Oid(buf) take 0.00018213ms on average

BSON@4.6.3
deserialize({ oid, string }, { validation: { utf8: false } }) takes 0.00303077ms on average
new Oid(buf) take 0.00059272ms on average

BSON@1.1.6
deserialize({ oid, string }, { validation: { utf8: false } }) takes 0.00203647ms on average
new Oid(buf) take 0.00031121ms on average
```

```
Iterations: 1000000

BSON@current local
deserialize({ oid, string }, { validation: { utf8: false } }) takes 0.00206223ms on average
new Oid(buf) take 0.00013110ms on average

BSON@4.6.3
deserialize({ oid, string }, { validation: { utf8: false } }) takes 0.00150312ms on average
new Oid(buf) take 0.00020934ms on average

BSON@1.1.6
deserialize({ oid, string }, { validation: { utf8: false } }) takes 0.00098606ms on average
new Oid(buf) take 0.00014574ms on average
```

Uzlopak:
What did you use to bench?

nbbeeken:
@durran I bumped the iterations and I added system info print out

```
- cpu: Intel(R) Core(TM) i5-8279U CPU @ 2.40GHz
- cores: 8
- os: darwin
- ram: 16GB
- iterations: 1,000,000

BSON@current local
deserialize({ oid, string }, { validation: { utf8: false } }) takes 0.00236919ms on average
new Oid(buf) take 0.00009414ms on average

BSON@4.6.3
deserialize({ oid, string }, { validation: { utf8: false } }) takes 0.00156644ms on average
new Oid(buf) take 0.00018286ms on average

BSON@1.1.6
deserialize({ oid, string }, { validation: { utf8: false } }) takes 0.00113818ms on average
new Oid(buf) take 0.00011388ms on average
```

durran (approved):

harrisonhunter:
Awesome! Thank you very much

PR #499, opened 2022-06-09
fix(NODE-4302): typo in typesVersions filename

jablko:
### Description

Older versions of TypeScript aren't working since https://github.com/mongodb/js-bson/pull/494#issue-1178148345 (BSON parser >= 4.6.3):
```
error TS7016: Could not find a declaration file for module 'bson'. 'node_modules/bson/lib/bson.js' implicitly has an 'any' type.
```

Looks like one instance was missed in https://github.com/mongodb/js-bson/pull/494/commits/740ce301ad77bfd6924b8e447907b92a40953783.

#### What is changing?

The `typesVersions` filename, so older versions of TypeScript can find the declaration file.

No.

I ran into this editing unrelated types on DefinitelyTyped that have a transitive dependency on BSON parser and are automatically tested on older versions of TypeScript. This typo causes that automation to complain, blocking edits to those transitively-dependent types.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
   "typesVersions": {
     "<=4.0.2": {
       "bson.d.ts": [
-        "bson.ts34.d.ts"
+        "bson-ts34.d.ts"
       ]
     }
   },


durran (approved):

nbbeeken:
Hi @jablko surprisingly I believe that downleveling the types isn't actually doing anything necessary here (our library should work on TS v4.0.0+) I've started another PR (#501) where I'm testing against a range of TS versions and the nightly version. 

If you still have your environment setup, can you see if locally removing the typesVersions alleviates the issue? if so we should be able to go with that approach in the next release.

PR #502, opened 2022-06-22
fix(NODE-3630): remove float parser and test edge cases for Double

aditi-khare-mongoDB:
No

The old custom float parsing (in float_parser.ts) did not support NaN with payload preservation, while native JS does.


nbbeeken (changes requested):

src/parser/serializer.ts
```diff
@@ -487,7 +486,7 @@ function serializeDouble(
   buffer[index++] = 0;
 
   // Write float
-  writeIEEE754(buffer, value.value, index, 'little', 52, 8);
+  buffer.writeDoubleLE(value.value, index); // little endian by default
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Ok sounds good!

test/node/double_tests.js
```diff
@@ -35,4 +35,63 @@ describe('Double', function () {
...
+    function twiceSerialized(value) {
+      let serializedDouble = BSON.serialize({ d: new Double(value) });
+      let deserializedDouble = BSON.deserialize(serializedDouble, { promoteValues: true });
+      let newVal = deserializedDouble.d;
+      return newVal;
+    }
+
+    it('inf', () => {
+      let value = Infinity;
+      let orig = new Double(value).valueOf();
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: That makes sense, I converted to just using value for inf, -inf, and epsilon

test/node/double_tests.js
```diff
@@ -35,4 +35,63 @@ describe('Double', function () {
...
+      let value = buffer.readDoubleLE(0);
+      let serializedDouble = BSON.serialize({ d: new Double(value) });
+      expect(serializedDouble.subarray(7, 15)).to.deep.equal(buffer);
+      let { d: newVal } = BSON.deserialize(serializedDouble, { promoteValues: true });
+      expect(Number.isNaN(newVal)).to.equal(true);
+    });
+
+    it('0', () => {
+      let value = 0;
+      let orig = new Double(value).valueOf();
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: What's interesting is that `expect(orig).to.equal(newVal);` and `expect(value).to.equal(newVal);`seem to be running with no errors for `value = -0` and `value = 0`.

test/node/double_tests.js
```diff
@@ -35,4 +35,63 @@ describe('Double', function () {
       });
     }
   });
+
+  describe('specialValues', () => {
+    function twiceSerialized(value) {
+      let serializedDouble = BSON.serialize({ d: new Double(value) });
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: This sounds good, thanks for reviewing my code!

test/node/double_tests.js
```diff
@@ -35,4 +35,63 @@ describe('Double', function () {
...
+    it('0', () => {
+      let value = 0;
+      let orig = new Double(value).valueOf();
+      let newVal = twiceSerialized(value);
+      expect(orig).to.equal(newVal);
+    });
+
+    it('-0', () => {
+      let value = -0;
+      let orig = new Double(value).valueOf();
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: It seems when we do the `let orig = new Double(value).valueOf();`, `typeof orig` is still a Number not an Object. So the class wrapping only converts to a Double when we just do Double(value).

test/node/double_tests.js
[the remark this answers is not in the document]
aditi-khare-mongoDB: With `let value = inf;`, when I do `typeof value` when it is still a Number, is this the desired behavior?

aditi-khare-mongoDB:
Deserialization - Perf Decline:

```
(specs are same as previous comment)

testing: Double Deserialization
local             - v e45f7d0  - avg 0.00025402ms
released          - v 4.6.4    - avg 0.00020516ms
previous major    - v 1.1.6    - avg 0.00017792ms
bson-ext          - v 4.0.2    - avg 0.00036723ms
```
^^ Deserialization with DV Allocation does show some slower performance when compared to previous versions. Since there must be a new allocation every time, this not resolvable with a global variable. 


```
testing: Many Doubles Deserialization
ALLOCATING ONCE                local          - v e45f7d0  - avg 0.00026782ms
ALLOCATING FOR EACH DOUBLE     local          - v e45f7d0  - avg 0.00026212ms
released       - v 4.6.4    - avg 0.00022623ms
previous major - v 1.1.6    - avg 0.00021576ms
bson-ext       - v 4.0.2    - avg 0.00040693ms
```
^^ With the deserialization of an array of many doubles, if we change the allocation to only happen once in the function, rather than once for every double, the performance does not change significantly.

etc/benchmarks/main.mjs
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
 import { performance } from 'perf_hooks';
+import bson from '../../lib/bson.js';
```
nbbeeken, line 3: We can remove this

nbbeeken (changes requested):

etc/benchmarks/main.mjs
aditi-khare-mongoDB: Got it!

nbbeeken (approved):

test/node/double_tests.js
```diff
@@ -35,4 +36,74 @@ describe('Double', function () {
       });
     }
   });
+
+  describe('specialValues', () => {
+    function twiceSerialized(value) {
```
dariakp, line 41: The function is called "twiceSerialized", but that's not really descriptive of what's actually done, since we're actually serializing once, and then deserializing it. What's done here is is generally referred to as "roundTripping", or you can be more explicit by saying `serializeThenDeserialize(value)` though even that is missing the last part of it, which is the extraction of the actual js value at the end.

It also wouldn't hurt to mention why we are only interested in promoteValues = true and what happens if promoteValues = false. I'm sure something can be done in a combination of the test case `description`, code comments, and the actual `it('should <behavior we are interested in verifying>')` to help make clear what we are testing and why in this set of tests.

test/node/double_tests.js
```diff
@@ -35,4 +36,74 @@ describe('Double', function () {
...
+      let newVal = twiceSerialized(orig);
+      expect(Object.is(newVal, -0)).to.be.true;
+    });
+
+    // TODO (NODE-4335): -0 should be serialized as double
+    it.skip('-0 serializes as Double', () => {
+      let value = -0;
+      let serializedDouble = BSON.serialize({ d: value });
+      let type = serializedDouble[5];
+      expect(type).to.not.equal(BSON.BSON_DATA_NUMBER);
```
dariakp, line 100: this isn't asserting that `-0` serializes to Double, it would be better to have the actual logic that shows that `-0` is a double that is appropriately roundtripped here

test/node/double_tests.js
```diff
@@ -35,4 +36,74 @@ describe('Double', function () {
...
+      expect(value).to.equal(newVal);
+    });
+
+    it('NaN', () => {
+      let value = NaN;
+      let newVal = twiceSerialized(value);
+      expect(Number.isNaN(newVal)).to.equal(true);
+    });
+
+    it('NaN with payload', function () {
```
dariakp, line 66: There's a lot of hard to follow logic in this test that doesn't make it clear what the assertions are really trying to accomplish, please clarify the description and the setup, possibly splitting things up into more than one test case as appropriate.

test/node/double_tests.js
```diff
@@ -35,4 +36,74 @@ describe('Double', function () {
...
+
+  describe('specialValues', () => {
+    function twiceSerialized(value) {
+      let serializedDouble = BSON.serialize({ d: value });
+      let deserializedDouble = BSON.deserialize(serializedDouble, { promoteValues: true });
+      let newVal = deserializedDouble.d;
+      return newVal;
+    }
+
+    it('inf', () => {
```
dariakp, line 48: My general recommendation for the test cases with repetitive setup is to make an array of test case names and values, and then generate the test cases based on that. It'll also make it more obvious if something requires special handling.

e.g.,
```js
const testCases = [
  { name: 'NaN', value: NaN }, 
  { name: 'Infinity', value: Infinity }
];

for (const { name, value } of testCases) {
  it(`should <describe what it should do> when value is: ${name}`, () => {
    const outcome = someFunction(value);
    expect(outcome).to.equal(value);
  })
}
```

dariakp (changes requested):

test/node/bson_corpus.spec.test.js
```diff
@@ -121,8 +122,7 @@ describe('BSON Corpus', function () {
         describe('valid-bson', function () {
           for (const v of valid) {
             it(v.description, function () {
-              if (v.description === 'NaN with payload') {
-                // TODO(NODE-3630): remove custom float parser so we can handle the NaN payload data
+              if (v.description === 'NaN with payload' && !isBrowser() && getNodeMajor() < 10) {
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: It originally wasn't working, but now it is, so I'll remove the if statement here!

test/node/double_tests.js
```diff
@@ -35,4 +36,74 @@ describe('Double', function () {
...
+    });
+
+    it('NaN', () => {
+      let value = NaN;
+      let newVal = twiceSerialized(value);
+      expect(Number.isNaN(newVal)).to.equal(true);
+    });
+
+    it('NaN with payload', function () {
+      if (!isBrowser() && getNodeMajor() < 10) {
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: same rationale as above

test/node/double_tests.js
```diff
@@ -35,4 +36,74 @@ describe('Double', function () {
       });
     }
   });
+
+  describe('specialValues', () => {
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Got it, I'll use the table testing descriptions that you provided moving forward

test/node/double_tests.js
```diff
@@ -35,4 +36,74 @@ describe('Double', function () {
...
+    function twiceSerialized(value) {
+      let serializedDouble = BSON.serialize({ d: value });
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Nope, I'll start using const for constant values

test/node/double_tests.js
aditi-khare-mongoDB: Sounds good, I've added more descripted titles for the test cases and function

test/node/double_tests.js
aditi-khare-mongoDB: Got it, I've changed it to check that the current implementation has it equalling an int and not equalling a Double.

test/node/double_tests.js
aditi-khare-mongoDB: That's right, I've refactored and added more descriptive variable names to make it more readable

test/node/double_tests.js
aditi-khare-mongoDB: Just added the table tests!

baileympearson (changes requested):

test/node/double_tests.js
```diff
@@ -3,36 +3,94 @@
...
+    it('should keep payload in serialize-deserialize roundtrip', function () {
+      expect(serializedNanPayloadDouble.subarray(7, 15)).to.deep.equal(NanPayloadBuffer);
+    });
+
+    it('should preserve NaN value in serialize-deserialize roundtrip', function () {
+      const { d: newVal } = BSON.deserialize(serializedNanPayloadDouble, { promoteValues: true });
+      expect(Number.isNaN(newVal)).to.equal(true);
+    });
+  });
+  it('NODE-4335: does not preserve -0 in serialize-deserialize roundtrip if JS number is used', () => {
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Sure, I flipped the test case logic and skipped the test case to be in line with the current convention.

test/node/tools/utils.js
```diff
@@ -158,6 +158,11 @@ exports.isNode6 = function () {
   return process.version.split('.')[0] === 'v6';
 };
 
+exports.getNodeMajor = function () {
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Got it!

test/node/double_tests.js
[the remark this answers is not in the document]
nbbeeken: I think we want to keep the test running here, so we should flip the assertion back and unskip the test. Bailey was saying that while we normally wouldn't have a test like this, it is a special case and he wants to make it excruciatingly clear that it's behavior that we ultimately do not desire but we also do not want to change until we're ready for a major version. Something like:
```
// TODO(NODE-XXX): xxx
// This test is demonstrating the behavior of -0 being serialized as an int32, something we do NOT want to unintentionally change, but may want to change in the future, which the above ticket serves to track.
```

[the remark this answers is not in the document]
baileympearson: 👍 neal was spot on.  It is an odd scenario where we've decided this bug can't be fixed without a breaking change, so we want to codify the actual buggy behavior in a test to ensure we don't fix it until we're ready (with the linked node ticket)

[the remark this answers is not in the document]
aditi-khare-mongoDB: Okay, sounds good

test/node/double_tests.js
```diff
@@ -3,36 +3,96 @@
...
+  ];
 
-    it('Number object', function (done) {
-      expect(new Double(new Number(value)).valueOf()).to.equal(value);
-      done();
+  for (const { name, input } of testCases) {
+    it(`should return Double from serialize-deserialize roundtrip when value is: ${name}`, () => {
+      const outcome = serializeThenDeserialize(input);
+      expect(outcome.value).to.equal(input.value);
+      expect(Object.is(outcome.value, input.value)).to.be.true;
```
dariakp, line 59: It may be good to add a comment why we are using two different ways of testing equality here (I assume it's to catch the `0, -0`), but that's not immediately obvious; actually, the equality check isn't necessary if you use `Object.is`, so we can probably get rid of that first assertion.

test/node/double_tests.js
```diff
@@ -3,36 +3,96 @@
...
+          const double = new Double(testNumber);
+          expect(double.toString(radix)).to.equal(testNumber.toString(radix));
+        });
+      }
     });
+  });
+
+  function serializeThenDeserialize(value) {
+    const serializedDouble = BSON.serialize({ d: value });
+    const deserializedDouble = BSON.deserialize(serializedDouble, { promoteValues: false });
```
dariakp, line 43: what's the significance of `promoteValues: false` here? what happens with `promoteValues: true`?

test/node/double_tests.js
```diff
@@ -3,36 +3,96 @@
...
+    { name: 'Zero', input: new Double(0) },
+    { name: 'Double (Negative Zero)', input: new Double(-0) }
+  ];
 
-    it('Number object', function (done) {
-      expect(new Double(new Number(value)).valueOf()).to.equal(value);
-      done();
+  for (const { name, input } of testCases) {
+    it(`should return Double from serialize-deserialize roundtrip when value is: ${name}`, () => {
+      const outcome = serializeThenDeserialize(input);
```
dariakp, line 57: Should we also be checking that the original input is preserved in all these cases? e.g., `input.value === Infinity` etc? If so, it might be better to use a `testValue` param in your table with the raw value, and then use `Object.is` on both the `testValue` and `doubleValue`, and then on the `doubleValue` and `roundTrippedValue` (or whatever you want to call the actual variables)

test/node/double_tests.js
```diff
@@ -3,36 +3,96 @@
...
+      expect(outcome.value).to.equal(input.value);
+      expect(Object.is(outcome.value, input.value)).to.be.true;
     });
+  }
+
+  it('should preserve NaN value in serialize-deserialize roundtrip', () => {
```
dariakp, line 63: you should be able to combine this with the table test if you make the other changes I am suggesting

test/node/double_tests.js
```diff
@@ -3,36 +3,96 @@
...
+    const deserializedDouble = BSON.deserialize(serializedDouble, { promoteValues: false });
+    return deserializedDouble.d;
+  }
+
+  const testCases = [
+    { name: 'Infinity', input: new Double(Infinity) },
+    { name: '-Infinity', input: new Double(-Infinity) },
+    { name: 'Number.EPSILON', input: new Double(Number.EPSILON) },
+    { name: 'Zero', input: new Double(0) },
+    { name: 'Double (Negative Zero)', input: new Double(-0) }
```
dariakp, line 52: -> just "Negative Zero" would be better here for consistency

test/node/double_tests.js
```diff
@@ -3,36 +3,96 @@
...
+  ];
 
-    it('Number object', function (done) {
-      expect(new Double(new Number(value)).valueOf()).to.equal(value);
-      done();
+  for (const { name, input } of testCases) {
+    it(`should return Double from serialize-deserialize roundtrip when value is: ${name}`, () => {
```
dariakp, line 56: -> 
```js 
`should preserve the input value ${name} in Double serialize-deserialize roundtrip` 
```

test/node/double_tests.js
```diff
@@ -3,36 +3,96 @@
...
+    const serializedNanPayloadDouble = BSON.serialize({ d: NanPayloadDouble });
+
+    it('should keep payload in serialize-deserialize roundtrip', function () {
+      expect(serializedNanPayloadDouble.subarray(7, 15)).to.deep.equal(NanPayloadBuffer);
     });
 
-    const testRadices = [2, 8, 10, 16, 22];
+    it('should preserve NaN value in serialize-deserialize roundtrip', function () {
+      const { d: newVal } = BSON.deserialize(serializedNanPayloadDouble, { promoteValues: true });
+      expect(Number.isNaN(newVal)).to.equal(true);
```
dariakp, line 85: I recommend the chai assertion
`expect(newVal).to.be.NaN`

dariakp (changes requested):
Just a bit of extra clean up on the tests - thanks a lot for making the other changes, much easier to read now!

test/node/double_tests.js
aditi-khare-mongoDB: This is so we get the raw BSON instead of the value as a double.

test/node/double_tests.js
```diff
@@ -3,36 +3,96 @@
...
-      expect(new Double(value).valueOf()).to.equal(value);
-      done();
+      it('Primitive number', function (done) {
+        expect(new Double(value).valueOf()).to.equal(value);
+        done();
+      });
+
+      it('Number object', function (done) {
+        expect(new Double(new Number(value)).valueOf()).to.equal(value);
+        done();
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Got it!

test/node/double_tests.js
aditi-khare-mongoDB: Alright

test/node/double_tests.js
aditi-khare-mongoDB: Got it

test/node/double_tests.js
aditi-khare-mongoDB: Got it

test/node/double_tests.js
aditi-khare-mongoDB: That sounds good, thanks

test/node/double_tests.js
aditi-khare-mongoDB: Thank you for your thorough comments once again!

test/node/double_tests.js
aditi-khare-mongoDB: That sounds good, I'll remove this function and add NaN to the table test cases

test/node/double_tests.js
aditi-khare-mongoDB: Yeah, the first equality is redundant, I'll remove it.

test/node/double_tests.js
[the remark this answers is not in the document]
aditi-khare-mongoDB: Removing the done() creates an error, I think the tests are actually async from the error
```

  2) BSON Double Precision
       class Double
         constructor()
           Primitive number:
     Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/aditi.khare/code/js-bson/test/node/double_tests.js)
      at listOnTimeout (node:internal/timers:559:17)
      at processTimers (node:internal/timers:502:7)

  3) BSON Double Precision
       class Double
         constructor()
           Number object:
     Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/aditi.khare/code/js-bson/test/node/double_tests.js)
      at listOnTimeout (node:internal/timers:559:17)
      at processTimers (node:internal/timers:502:7)

```

test/node/double_tests.js
aditi-khare-mongoDB: Sounds good!

test/node/double_tests.js
dariakp: could you add a comment above this line explaining that? also @nbbeeken is there a more precise option we can use than the general promoteValues?

test/node/double_tests.js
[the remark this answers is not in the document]
aditi-khare-mongoDB: Got it!

test/node/double_tests.js
nbbeeken: Unfortunately Doubles do not have a dedicated promote option, promoteValues: false is the only way to get the deserializer to return instanceof Double. _chaos_

nbbeeken (approved):

dariakp (approved):

PR #509, opened 2022-07-26
feat(NODE-4419): UUID class deserialization

aditi-khare-mongoDB:
### Description
When deserializing a UUID object, the user can use the boolean option ```promoteUUIDs``` to return an object of type UUID. This is part of an ongoing project to better integrate the use of the UUID class into js-bson.

#### What is changing?
There is a new option ```promoteUUIDs``` available to users when deserializing objections.

If the existing object is of the UUID subtype, and the flag is true (ex: ```{promoteUUIDs: true}```), then the resulting deserialization will return a UUID object. If the flag is false, or omitted it will return a binary object.

After deserializing, users can now have UUID objects. 

### DIFF
--- a/src/parser/deserializer.ts
+++ b/src/parser/deserializer.ts
@@ -35,6 +35,8 @@ export interface DeserializeOptions {
   promoteBuffers?: boolean;
   /** when deserializing will promote BSON values to their Node.js closest equivalent types. */
   promoteValues?: boolean;
+  /** when deserializing will return UUID type */
+  promoteUUIDs?: boolean;
   /** allow to specify if there what fields we wish to return as unserialized raw buffer. */
   fieldsAsRaw?: Document;
   /** return BSON regular expressions as BSONRegExp instances. */
@@ -135,6 +137,7 @@ function deserializeObject(
   const promoteBuffers = options['promoteBuffers'] == null ? false : options['promoteBuffers'];
   const promoteLongs = options['promoteLongs'] == null ? true : options['promoteLongs'];
   const promoteValues = options['promoteValues'] == null ? true : options['promoteValues'];
+  const promoteUUIDs = options['promoteUUIDs'] == null ? false : options['promoteUUIDs'];
 
   // Ensures default validation option if none given
   const validation = options.validation == null ? { utf8: true } : options.validation;
@@ -413,7 +416,9 @@ function deserializeObject(
             throw new BSONError('Binary type with subtype 0x02 contains too short binary size');
         }
 
-        if (promoteBuffers && promoteValues) {
+        if (promoteUUIDs && subType === 4) {
+          value = new Binary(buffer.slice(index, index + binarySize), subType).toUUID();
+        } else if (promoteBuffers && promoteValues) {
           value = buffer.slice(index, index + binarySize);
         } else {
           value = new Binary(buffer.slice(index, index + binarySize), subType);
@@ -440,7 +445,9 @@ function deserializeObject(
           _buffer[i] = buffer[index + i];
         }
 
-        if (promoteBuffers && promoteValues) {
+        if (promoteUUIDs && subType === 4) {
+          value = new Binary(_buffer, subType).toUUID();
+        } else if (promoteBuffers && promoteValues) {
           value = _buffer;
         } else {
           value = new Binary(_buffer, subType);
--- a/src/parser/deserializer.ts
+++ b/src/parser/deserializer.ts
@@ -35,7 +35,7 @@ export interface DeserializeOptions {
   promoteBuffers?: boolean;
   /** when deserializing will promote BSON values to their Node.js closest equivalent types. */
   promoteValues?: boolean;
-  /** when deserializing will return UUID type */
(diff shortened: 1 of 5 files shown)


src/parser/deserializer.ts
```diff
@@ -440,7 +445,9 @@ function deserializeObject(
           _buffer[i] = buffer[index + i];
         }
 
-        if (promoteBuffers && promoteValues) {
+        if (promoteUUIDs && subType === 4) {
+          value = new Binary(_buffer, subType).toUUID();
+        } else if (promoteBuffers && promoteValues) {
           value = _buffer;
```
nbbeeken, line 451: This ordering of logic is important, if promoteBuffers is enabled we will not return a node Buffer anymore if promoteUUIDs is enabled. I think that behavior makes sense and we should note it in the TS-doc for the option.
And then let's also add a test for when both options are passed, assert UUID / Buffer comes out with the combination of true/false for each.

src/parser/deserializer.ts
```diff
@@ -135,6 +137,7 @@ function deserializeObject(
   const promoteBuffers = options['promoteBuffers'] == null ? false : options['promoteBuffers'];
   const promoteLongs = options['promoteLongs'] == null ? true : options['promoteLongs'];
   const promoteValues = options['promoteValues'] == null ? true : options['promoteValues'];
+  const promoteUUIDs = options['promoteUUIDs'] == null ? false : options['promoteUUIDs'];
```
nbbeeken, line 140:
```suggestion
  const promoteUUIDs = options.promoteUUIDs == null ? false : options.promoteUUIDs;
```
We don't need to use hard coded computed property access, the `.` operator works fine. We should clean up the other options too, but let's save that for another PR.

src/parser/deserializer.ts
```diff
@@ -35,6 +35,8 @@ export interface DeserializeOptions {
   promoteBuffers?: boolean;
   /** when deserializing will promote BSON values to their Node.js closest equivalent types. */
   promoteValues?: boolean;
+  /** when deserializing will return UUID type */
```
nbbeeken, line 38: see other comment about adding to the docs.

nbbeeken (changes requested):

src/parser/deserializer.ts
aditi-khare-mongoDB: sounds good!

src/parser/deserializer.ts
aditi-khare-mongoDB: sounds good!

src/parser/deserializer.ts
aditi-khare-mongoDB: just added the tests! with flatMap!

test/node/uuid_tests.js
```diff
@@ -162,4 +162,42 @@ describe('UUID', () => {
...
+
+    it('should not promoteUUIDs when flag is false', () => {
+      const { uuid: unpromotedUUID } = BSON.deserialize(serializedUUID, { promoteUUIDs: false });
+      expect(unpromotedUUID._bsontype).to.equal('Binary');
+    });
+
+    it('should not promoteUUIDs when flag is omitted', () => {
+      const { uuid: omittedFlagUUID } = BSON.deserialize(serializedUUID);
+      expect(omittedFlagUUID._bsontype).to.equal('Binary');
+    });
```
nbbeeken, line 181: Discussed adding a test to assert that the toUUID errors will also get thrown from the deserializer when those unhappy paths are encountered

nbbeeken (changes requested):

test/node/uuid_tests.js
```diff
@@ -162,4 +162,42 @@ describe('UUID', () => {
     const uuid = new UUID(UPPERCASE_DASH_SEPARATED_UUID_STRING);
     expect(inspect(uuid)).to.equal(`new UUID("${LOWERCASE_DASH_SEPARATED_UUID_STRING}")`);
   });
+
+  describe('deserialize', () => {
+    const serializedUUID = BSON.serialize({ uuid: new BSON.UUID().toBinary() });
```
durran, line 167: As was mentioned by @dariakp in our virtual office time, can we also here save the UUID itself to a variable and then assert the deserialised value is also the same in these tests and not just has the correct `_bsontype`?

durran (changes requested):

test/node/uuid_tests.js
aditi-khare-mongoDB: That sounds good!

test/node/uuid_tests.js
aditi-khare-mongoDB: Sounds good!

durran (approved):

nbbeeken (approved):

aditi-khare-mongoDB:
> Have Evergreen failures on Node 10 and lower due to `flatMap` not being present. Should we remove those lower versions from the matrix as driver is only Node 12 and higher?

yeah, I've installed a new flatMap dependency that fixed this problem!

PR #512, opened 2022-08-05
feat(NODE-4506): Make UUID a subclass of binary

aditi-khare-mongoDB:
### Description
UUID is now a subclass of Binary!

#### What is changing?
UUID is now a subclass of Binary. UUID is no longer a _bsontype. `EJSON.parse`, `EJSON.stringify`,` BSON.serialize`, and `BSON.deserialize` are now supported for UUID.

Here is a [link ](https://docs.google.com/document/d/1Th8TgIMwwHztawg-4e4wfchJ95zKxxW4eS5ldWew2Kc/edit#heading=h.i9zc47oeau8l) to the section of the Design Documentation for Ease of Using UUIDs that outlines in more detail the changes made by this PR.

Yes, `EJSON.parse`, `EJSON.stringify`,` BSON.serialize`, and `BSON.deserialize` are now supported for UUID without calling the `toBinary() `on the UUID.

Make UUID more seamless to use.
So that UUID has the same functionality as Binary and other BSON classes, without any explicit conversion toBinary() first.  

TODO: add tests in BSON-EXT 


src/binary.ts
```diff
@@ -292,3 +293,204 @@ export class Binary {
...
+
+/**
+ * A class representation of the BSON UUID type.
+ * @public
+ */
+export class UUID extends Binary {
+  static cacheHexString: boolean;
+
+  /** UUID Bytes @internal */
+  private [kId]!: Buffer;
```
addaleax, line 313: Is this needed, given that the parent class has a `buffer` property that provides the same information?

src/parser/serializer.ts
```diff
@@ -1047,8 +1043,6 @@ export function serializeInto(
         index = serializeFunction(buffer, key, value, index, checkKeys, depth, serializeFunctions);
       } else if (value['_bsontype'] === 'Binary') {
         index = serializeBinary(buffer, key, value, index);
-      } else if (value['_bsontype'] === 'UUID') {
-        index = serializeBinary(buffer, key, value.toBinary(), index);
```
addaleax, line 1051: I think changing `_bsontype` is a good thing, especially since UUID serialization/deserialization has not been released yet, so there's no need to worry about interoperability with existing bson package releases 👍

test/node/uuid_tests.js
```diff
@@ -166,24 +164,6 @@ describe('UUID', () => {
   });
 
   describe('serialize', () => {
-    it('should have a valid UUID _bsontype with Object input without error', () => {
```
nbbeeken, line 169: I think these tests should still work, no?

src/binary.ts
```diff
@@ -292,3 +293,204 @@ export class Binary {
 }
 
 Object.defineProperty(Binary.prototype, '_bsontype', { value: 'Binary' });
+
+/** @public */
+export type UUIDExtended = {
+  $uuid: string;
+};
+const BYTE_LENGTH = 16;
```
nbbeeken, line 301:
```suggestion
const UUID_BYTE_LENGTH = 16;
```

Since this is no longer in a UUID specific file

nbbeeken (changes requested):

src/binary.ts
```diff
@@ -337,7 +333,7 @@ export class UUID extends Binary {
       );
     }
     super(bytes, BSON_BINARY_SUBTYPE_UUID_NEW);
-    this[kId] = bytes;
+    this.buffer = bytes;
```
nbbeeken, line 336: We can remove this altogether now since the buffer will be assigned in super, right?

nbbeeken (changes requested):

src/binary.ts
```diff
@@ -292,3 +293,200 @@ export class Binary {
...
+
+    if (UUID.cacheHexString) {
+      this.__id = bufferToUuidHexString(value);
+    }
+  }
+
+  /**
+   * Generate a 16 byte uuid v4 buffer used in UUIDs
+   */
+
```
addaleax, line 359: This seems like a comment that got lost somewhere, but should probably be removed at some point?
```suggestion
```

src/binary.ts
```diff
@@ -292,3 +293,200 @@ export class Binary {
...
+        return false;
+      }
+
+      try {
+        // get this byte as hex:             xxxxxxxx-xxxx-XXxx-xxxx-xxxxxxxxxxxx
+        // check first part as uuid version: xxxxxxxx-xxxx-Xxxx-xxxx-xxxxxxxxxxxx
+        return parseInt(input[6].toString(16)[0], 10) === Binary.SUBTYPE_UUID;
+      } catch {
+        return false;
+      }
```
addaleax, line 464: Feel free to ignore since this is existing code, but this is ... just a bit odd? The fact that `Binary.SUBTYPE_UUID` happens to also match the value of a specific digit in UUID hex expansions is pure coincidence, and combining `.toString()` with `parseInt()` is an unnecessary roundtrip.

This could be replaced with something like

```suggestion
      return (input[6] & 0xf0) === 0x40;
```

to match the code in `generate()`, or, if we want to match that check even more strictly,

```suggestion
      return (input[6] & 0xf0) === 0x40 && (bytes[8] & 0x80) === 0x80;
```

src/binary.ts
aditi-khare-mongoDB: Sounds good!

src/binary.ts
aditi-khare-mongoDB: Sounds good!

src/binary.ts
aditi-khare-mongoDB: I'm not sure since UUID.generate also uses the bytes variables

addaleax (approved):

nbbeeken (approved):

baileympearson (approved):

PR #513, opened 2022-08-11
feat(NODE-4535): automatically promote UUIDs when deserializing and parsing UUIDs

aditi-khare-mongoDB:
### Description
If a Binary bytes or EJSON strings have subtype 4, then BSON.deserialize and EJSON.parse, respectively will automatically promote it to UUID Type.

#### What is changing?
Removed promoteUUIDs as an option, and added automatic promotion. In addition, this PR also removes promoteUUID related tests that were failing in previous commits in the parent branch. 

Example with EJSON Parse:
Previous:
```
EJSON.parse(`{"u":{"$uuid":"bc3ceb30-f100-40e1-a403-cc0650d48ec7"}}`);
// {u: new Binary(Buffer.from("bc3ceb30f10040e1a403cc0650d48ec7", "hex"), 4)}
```
Proposed: 
```
EJSON.parse(`{"u":{"$uuid":"bc3ceb30-f100-40e1-a403-cc0650d48ec7"}}`);
//  {u: new UUID("bc3ceb30f10040e1a403cc0650d48ec7")}
```

Yes


durran (changes requested):
@aditi-khare-mongoDB what was the reason for the deletion of the valid compliance tests?

aditi-khare-mongoDB:
> @aditi-khare-mongoDB what was the reason for the deletion of the valid compliance tests?

The tests included buggy inputs of incorrect length for UUIDs that were not previously exposed until we started automatically promoting UUIDs when deserializing.

durran (dismissed):

src/binary.ts
```diff
@@ -278,7 +278,7 @@ export class Binary {
     if (!data) {
       throw new BSONTypeError(`Unexpected Binary Extended JSON format ${JSON.stringify(doc)}`);
     }
-    return new Binary(data, type);
+    return type === 4 ? new UUID(data) : new Binary(data, type);
```
nbbeeken, line 281: Should we use the constant here? just for consistency

nbbeeken (changes requested):

src/parser/deserializer.ts
```diff
@@ -445,12 +443,14 @@ function deserializeObject(
...
-          value = new Binary(_buffer, subType).toUUID();
-        } else if (promoteBuffers && promoteValues) {
+        if (promoteBuffers && promoteValues) {
           value = _buffer;
         } else {
           value = new Binary(_buffer, subType);
+          value = new Binary(buffer.slice(index, index + binarySize), subType);
+          if (subType === constants.BSON_BINARY_SUBTYPE_UUID_NEW) {
+            value = value.toUUID();
+          }
```
nbbeeken, line 453: Another consistency comment, lets use either an if stmt or a ternary in both places, really minor but might as well have the code match up

nbbeeken (changes requested):

test/node/extended_json_tests.js
```diff
@@ -773,5 +773,15 @@ Converting circular structure to EJSON:
       );
       expect(parsedUndashedInput).to.deep.equal(parsedDashedInput);
     });
+
+    it('should return UUID object when deserializing UUID subtype', () => {
+      const exampleUUID = new BSON.UUID('878dac12-01cc-4830-b271-cbc8518e63ad');
+      const stringifiedUUID = EJSON.stringify({ uuid: exampleUUID });
+      const parsedUUID = EJSON.parse(stringifiedUUID);
+      const expectedResult = {
+        uuid: new UUID('878dac1201cc4830b271cbc8518e63ad')
```
baileympearson, line 782: not blocking but could we use the same format for the UUID string in both cases?  In the first we use `-` and the second we don't.

test/node/uuid_tests.js
```diff
@@ -193,63 +193,14 @@ describe('UUID', () => {
...
-          } else if (uuid._bsontype === 'Binary') {
-            expect(uuid.buffer).to.deep.equal(originalUUID.id);
-          } else if (uuid._bsontype === undefined) {
-            expect(uuid).to.deep.equal(originalUUID.id);
-          } else {
-            expect.fail('Unexpected _bsontype');
-          }
-        });
-      }
+    it('should return UUID object when deserializing UUID subtype', () => {
```
baileympearson, line 196: same comment here

baileympearson (approved):
two small, non-blocking comments. otherwise LGTM

test/node/uuid_tests.js
aditi-khare-mongoDB: sounds good! I originally had the dashes and without the dashes because UUID should technically treat those two inputs as the same, but I understand how that is not within the scope of these tests / could be confusing for developers looking at this code later.

src/parser/deserializer.ts
aditi-khare-mongoDB: I added an if else since the ternary would be a bit long / hard to read!

src/binary.ts
aditi-khare-mongoDB: yes!

baileympearson (approved):

nbbeeken (approved):

PR #516, opened 2022-10-18
perf(NODE-4726): bench many strings in large batch in parallel

nbbeeken:
#### What is changing?

There's still some cruft in here, but its intentional in case someone else want to pull it down and check my work.
The issue reported does reproduce but only when we try to simulate parallelism. I don't think it is the parallelism that effects our BSON library it just offers enough load to be able to measure the difference. 

The goal here isn't to fix anything but some hypothesis, some ruled out:
- getValidatedString moved the buffer.toString to a helper, undoing this did not demonstrate a change
- removing all the UTF8 key validation logic may have introduced overhead of allocating a Set, checking has etc.
  - removing it did not appear to have an impact
- The analysis that I would recommend in a follow up is taking a close look at globals / new structures added to deserializeObject that don't exist in v1. Some known ones are:
  - DBRef shape checking
  - DataView for double parsing
  - UTF8 stuff ^

This benchmark is meant to demonstrate the perf difference so we can follow up with fixes that can be proven to be effective.

### DIFF
--- a/etc/benchmarks/lib_runner.mjs
+++ b/etc/benchmarks/lib_runner.mjs
@@ -54,15 +54,15 @@ export function getCurrentLocalBSON(libs) {
 
 export async function getLibs() {
   return await Promise.all([
-    // (async () => {
-    //   const { stdout } = await exec('git rev-parse --short HEAD');
-    //   const hash = stdout.trim();
-    //   return {
-    //     name: 'local',
-    //     lib: await import('../../lib/bson.js'),
-    //     version: hash
-    //   };
-    // })(),
+    (async () => {
+      const { stdout } = await exec('git rev-parse --short HEAD');
+      const hash = stdout.trim();
+      return {
+        name: 'local',
+        lib: await import('../../lib/bson.js'),
+        version: hash
+      };
+    })(),
     (async () => ({
       name: 'released',
       lib: await import('../../node_modules/bson_latest/lib/bson.js'),
@@ -76,6 +76,8 @@ export async function getLibs() {
         version: (await readJSONFile('../../node_modules/bson_legacy/package.json')).version
       };
     })()
+    // BSON-EXT is EOL so we do not need to keep testing it, and it has issues installing it
+    // in this no-save way on M1 currently that are not worth fixing.
     // (async () => ({
     //   name: 'bson-ext',
     //   lib: await import('../../node_modules/bson_ext/lib/index.js'),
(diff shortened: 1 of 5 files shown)


dariakp (changes requested):
This PR needs a rebase and a fresh re-install of dependencies.

Let's make sure that we capture the suggested approach in the description in a comment on the follow up ticket (either NODE-4283 or a new one that captures a more specific perf improvement)

etc/benchmarks/lib_runner.mjs
```diff
@@ -52,15 +54,15 @@ export function getCurrentLocalBSON(libs) {
...
-    (async () => {
-      const { stdout } = await exec('git rev-parse --short HEAD');
-      const hash = stdout.trim();
-      return {
-        name: 'local',
-        lib: await import('../../lib/bson.js'),
-        version: hash
-      };
-    })(),
+    // (async () => {
```
dariakp, line 57: if we're going to keep commented code, we should add a description of its purpose (same goes for the code in L78)

dariakp (changes requested):
Just one minor thing here

etc/benchmarks/lib_runner.mjs
nbbeeken: Uncommented importing the current local BSON, and commented on BSON-EXT

package.json
```diff
@@ -68,7 +69,8 @@
     "tsd": "^0.24.1",
     "typescript": "^4.8.4",
     "typescript-cached-transpile": "0.0.6",
-    "uuid": "^9.0.0"
+    "uuid": "^9.0.0",
+    "v8-profiler-next": "^1.9.0"
```
nbbeeken, line 73: This seems to have added a lot to the lock file I'm happy to make this a "must install before launching benchmaks" style dep, much like the other versions of the BSON lib are installed with --no-save

dariakp: no strong preferences here since it's a dev dependency; though if we're intending to run perf tests in the pipeline, may be worth keeping just to avoid surprise failures

dariakp (approved):

etc/benchmarks/main.mjs
```diff
@@ -98,6 +98,37 @@ await runner({
...
+    await Promise.all(
+      Array.from(
+        { length: 100 },
+        (_, i) =>
+          new Promise(resolve => {
+            setTimeout(() => {
+              resolve(bson.lib.deserialize(document, { validation: { utf8: false } }));
+            }, 20);
+          })
+      )
```
baileympearson, line 127: What is the purpose of the setTimeout here?  And why did you choose a setTimeout of 20ms specifically?

nbbeeken: It manufactures the delay that the driver does when we promise.all many many finds, 20ms is arbitrary, kept it low so the bench runs in a reasonable amount of time

PR #517, opened 2022-10-19
feat!(NODE-4435): drop support for nodejs versions below 14

nbbeeken:
#### What is changing?

- Drops EOL nodejs versions.
- Update dependencies, minor fixes to get everything working again


.eslintrc.json
```diff
@@ -63,5 +63,17 @@
...
+  },
+  "overrides": [
+    {
+      "parser": "espree",
+      "files": [
+        "*.mjs"
+      ],
+      "parserOptions": {
+        "sourceType": "module",
+        "ecmaVersion": "latest"
```
nbbeeken, line 75: This was to get eslint to be happy about our rollup config (not strictly necessary, but ew! redlines!) even latest isn't latest enough yet for the json import assertion, but at least that's the only error left.

.evergreen/install-dependencies.sh
```diff
@@ -55,5 +55,7 @@ registry=https://registry.npmjs.org
 EOT
 
 # install node dependencies
-npm install # npm prepare runs after install and will compile the library
+# npm prepare runs after install and will compile the library
+# TODO(NODE-3555): rollup dependencies for node polyfills have broken peerDeps. We can remove this flag once we've removed them.
+npm install --legacy-peer-deps
```
nbbeeken, line 60: We're going to be removing the rollup plugins that are causing peer dep issues in the next ticket, so this is a temp patch to get install to keep working.

.evergreen/run-checks.sh
```diff
@@ -6,4 +6,4 @@ NODE_ARTIFACTS_PATH="${PROJECT_DIRECTORY}/node-artifacts"
 export NVM_DIR="${NODE_ARTIFACTS_PATH}/nvm"
 [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
 
-npm run lint
+npm run check:lint
```
nbbeeken, line 9: driver scripts alignment

.evergreen/run-tests.sh
```diff
@@ -13,9 +13,10 @@ export NVM_DIR="${NODE_ARTIFACTS_PATH}/nvm"
 
 case $1 in
   "node")
-    npm run test-node
+    npm run check:coverage
```
nbbeeken, line 16: Getting out in front of the changes in coming in the many future ticket, I wanna make sure we don't drop as tests might get some minimal refactoring along with src as needed.

.nycrc.json
```diff
@@ -1,4 +1,12 @@
...
-  "reporter": ["lcovonly", "text-summary", "html"]
+  "reporter": [
+    "lcovonly",
+    "text-summary",
+    "html"
+  ],
+  "statements": 53,
+  "branches": 38,
+  "functions": 52,
+  "lines": 54
```
nbbeeken, line 11: These numbers are -1 the current values, trying to prevent regression for now, we can work on improving this as we go.

package.json
```diff
@@ -91,22 +89,20 @@
     "./dist/bson.esm.js": "./dist/bson.browser.esm.js"
   },
   "engines": {
-    "node": ">=6.9.0"
+    "node": ">=14.20.1"
```
nbbeeken, line 92: Current 14.x release. This is what we did when we picked the 12 ver for the driver we set it to the minor that was available at the time, we only ever pull in the latest minor for our testing so I think this is effective at communicating we test on 14.20+ (I can drop the patch version if we want)

.evergreen/config.yml
```diff
@@ -162,16 +122,17 @@ tasks:
...
-    tags: ["browser"]
-    commands:
-      - func: fetch source
-        vars:
-          NODE_MAJOR_VERSION: 16
-      - func: install dependencies
-      - func: run tests
-        vars:
-          TEST_TARGET: browser
+  # TODO(NODE-3555): Karma tests pass locally still after these changes, rhel80 is just missing chrome
```
nbbeeken, line 125: We do not have chrome set up on rhel80 so I'm skipping these in CI (we intend to remove them) but I wanna keep the rest of the karma set up around b/c it does work locally and demonstrates the updates did not break that build

.nycrc.json
durran: Just curious on what was the issue here and why this was needed or what was the potential regression?

nbbeeken: We just didn't use the coverage in CI, specifying these numbers along with the `--check-coverage` flag in the npm script will make nyc exit with failure if we start testing less, so there's no regression here I want to start testing these since we're kicking off a number of changes for the major

dariakp (approved):

PR #518, opened 2022-10-21
feat!(NODE-4802): Refactor BSON to work with cross platform JS APIs

nbbeeken:
### Description

NODE-3555 / NODE-4802

#### What is changing?

Refactors BSON to rely on common JS globals and runs tests for both node and non node global scenarios.

##### `ByteUtils`

A new collection of functions that change their implementation based on the environment the BSON library is imported into. If Buffer is available the ByteUtils wraps the same Buffer functions we have been using in BSON thus far. If Buffer is not available ByteUtils uses a combination of existing JS TypedArray APIs as well as some original implementations to get the needed format translations BSON requires. 

##### Using VM context

In order to test how the BSON library behaves when Buffer exists or not, we use `vm.runInContext` and read in the umd bundle as a string. We can then "run" the library with our own definition of a global object. All BSON tests are run both with and without Buffer being defined. The only globals required are: `TextEncoder`, `TextDecoder`, `atob`, `btoa`.

##### `buffer.copy`

Buffer.copy calls have been translated to `dst.set(src.subarray(), offset)` calls. As seen here: https://github.com/nodejs/node/blob/main/lib/buffer.js?rgh-link-date=2022-10-26T15%3A06%3A09Z#L247-L262 this is the same approach taken internally by node we just have inputs that we can trust are the correct type.

##### `isArray` flag

The `isArray` flag internal to the the serializer has been removed, it didn't provide functionality other than switching the key encoding to ASCII, but that doesn't actually change the output for a stringified number which is always "ascii". Additionally, we were mistakenly setting this flag with the wrong variable `serializeFunctions`, which would cause functions with names beyond the ascii range to be serialized with broken keys. It has to be fixed in this PR b/c of typescript, but I've filed NODE-4771 so we can make a separate entry in the changelog about this.

##### chai `.throws`

When running the BSON library from another context, instanceof checks on errors will fail so I've enhanced the throws assertion to permit errors whose `.name` property match the expected error name.

##### `ObjectId.toString` / `UUID.toString` / `Binary.toString`

The above methods all supported passing an encoding through to the Buffer.toString implementation, but the various encodings supported are unlikely to be useful and would be out of scope to maintain for this library. The underlying buffer can be fetched and passed to the Node.js Buffer .toString implementation if so desired. However, we will continue to support a common subset of the encodings that are likely to be useful, and the TS asserts that the encodings are only the ones we do accept now. (base64/utf8/hex)

Maintainability


src/parser/serializer.ts
```diff
@@ -42,30 +41,28 @@ export interface SerializeOptions {
 const regexp = /\x00/; // eslint-disable-line no-control-regex
 const ignoreKeys = new Set(['$db', '$ref', '$id', '$clusterTime']);
 
+function setUtf8String(buffer: Uint8Array, stringToSet: string, offset: number): number {
```
nbbeeken, line 44: Created this as a replacement for the `isArray ? utf8 : ascii` logic for keys. There is no difference between utf8 and ascii serialization so the condition was unnecessary. The logic is also the same for setting strings inside an existing buffer so I use it for the BSON values that are strings as well.

src/parser/serializer.ts
```diff
@@ -1040,7 +918,7 @@ export function serializeInto(
           ignoreUndefined
         );
       } else if (typeof value === 'function' && serializeFunctions) {
-        index = serializeFunction(buffer, key, value, index, checkKeys, depth, serializeFunctions);
+        index = serializeFunction(buffer, key, value, index, checkKeys, depth); // Oh lordy, bug!
```
nbbeeken, line 921: Same mention as above, I can clean up the comment, just noted it as I was fixing the API calls.

src/utils/node_byte_utils.ts
```diff
@@ -0,0 +1,79 @@
...
+  isBuffer(value: unknown): value is NodeJsBuffer;
+  prototype: {
+    toString: (this: Uint8Array, encoding: NodeJsEncoding) => string;
+    equals: (this: Uint8Array, other: Uint8Array) => boolean;
+  };
+};
+
+// This can be nullish, but we gate the nodejs functions on being exported whether or not this exists
+// Node.js global
+declare const Buffer: NodeJsBuffer;
```
nbbeeken, line 18: So this may or may not exist depending on env, but I'm relying on the check to be made in byte_utils.ts so that we never enter these helpers if there is no global Buffer constructor.

src/utils/node_byte_utils.ts
```diff
@@ -0,0 +1,79 @@
...
+
+function bytesToString(buffer: Uint8Array, encoding: NodeJsEncoding) {
+  return Buffer.prototype.toString.call(nodeJsByteUtils.toLocalBufferType(buffer), encoding);
+}
```
nbbeeken, line 22: We have to call toLocalBufferType before passing the buffer to the prototype call so wrapping this up in a helper seemed logical

src/binary.ts
```diff
@@ -216,17 +216,18 @@ export class Binary {
...
-    return this.buffer.toString('base64');
+    return ByteUtils.toBase64(this.buffer);
   }
 
-  toString(format?: string): string {
-    return this.buffer.toString(format);
+  // TODO: BREAKING
+  toString(): string {
+    return ByteUtils.toText(this.buffer);
   }
```
nbbeeken, line 225: This wasn't predicted in our set of breaking changes but I think its an additional one we'll want to make, we shouldn't implement every format that Node supports. The workaround is to call to string on the buffer directly.

src/objectid.ts
```diff
@@ -73,17 +72,14 @@ export class ObjectId {
...
     } else if (typeof workingId === 'string') {
       if (workingId.length === 12) {
-        const bytes = Buffer.from(workingId);
-        if (bytes.byteLength === 12) {
-          this[kId] = bytes;
-        } else {
-          throw new BSONTypeError('Argument passed in must be a string of 12 bytes');
-        }
+        throw new BSONError(
+          'TODO, we are not supporting this anymore and the behavior change is not worth fixing'
```
nbbeeken, line 79: I'm breaking this now since I ran into issues with testing but, We can clean it up proper in the follow up ticket to remove the binary strings

src/objectid.ts
```diff
@@ -189,12 +185,12 @@ export class ObjectId {
 
   /**
    * Converts the id into a 24 character hex string for printing
-   *
-   * @param format - The Buffer toString format parameter.
+   * TODO: THIS IS A BREAKAGE.. But implementing this in web is overkill
+   * @param format - hex or base64
    */
-  toString(format?: string): string {
+  toString(format?: 'hex' | 'base64'): string {
```
nbbeeken, line 191: Here again we should not support every format nodejs supported, work around is to get the buffer and call toString directly, continuing to support hex and base64 are common and easy since we have that logic in-house anyway

src/uuid_utils.ts
```diff
@@ -8,26 +8,26 @@ const VALIDATION_REGEX =
...
-      '-' +
-      buffer.toString('hex', 6, 8) +
-      '-' +
-      buffer.toString('hex', 8, 10) +
-      '-' +
-      buffer.toString('hex', 10, 16)
-    : buffer.toString('hex');
+export function bufferToUuidHexString(buffer: Uint8Array, includeDashes = true): string {
+  if (includeDashes) {
+    return [
```
nbbeeken, line 24: I'm happy to revert this change, I made it when there was a bunch of type errors in the initial refactoring of all the byteUtils, but it's kinda clean this way 🤷🙂

test/node/bson_older_versions_tests.js
```diff
@@ -223,9 +228,7 @@ describe('Mutual version and distribution compatibility', function () {
 
             try {
               // Check that both BSON versions serialize to equal Buffers
-              expect(toBSON.serialize({ object }).toString('hex')).to.equal(
-                fromBSON.serialize({ object }).toString('hex')
-              );
+              expect(toBSON.serialize({ object })).to.deep.equal(fromBSON.serialize({ object }));
```
nbbeeken, line 231: The toString('hex') isn't always available now

test/node/error_test.js
```diff
@@ -23,10 +29,15 @@ describe('BSONTypeError', function () {
 describe('BSONError', function () {
   it('should evaluate true on instanceof BSONError and Error', function () {
     const bsonErr = new BSONError();
-    expect(bsonErr instanceof BSONError).to.be.true;
-    expect(bsonErr instanceof Error).to.be.true;
-    expect(bsonErr).to.be.instanceOf(BSONError);
-    expect(bsonErr).to.be.instanceOf(Error);
+    if (__isWeb__) {
+      expect(bsonErr).to.have.property('name', 'BSONError');
```
nbbeeken, line 33: `instanceof` doesn't work since the "web" tests are run in a different context, so name prop check will suffice

test/register-bson.js
```diff
@@ -1,7 +1,65 @@
...
+ * since the error classes are declared in a different realm (?)
+ * So here we augment the chai assertion to fallback on checking the name property of the error instance
+ */
+chai.use(function (chai) {
+  const throwsAssertion = chai.Assertion.prototype.throw;
+  chai.Assertion.addMethod('throw', function (...args) {
+    try {
+      throwsAssertion.call(this, ...args);
+    } catch (assertionError) {
+      if (assertionError.actual?.name === assertionError.expected) {
```
nbbeeken, line 24: If the name of the actual equals the stringified expected, then we can pass the assertion. This is to make instanceof checks that fail still work in the "web" mode

test/register-bson.js
```diff
@@ -1,7 +1,65 @@
...
+        return;
+      }
+      throw assertionError;
+    }
+  });
+});
+
+// Register expect globally, this is no longer necessary since we do not use karma / bundle everything together
+globalThis.expect = chai.expect;
```
nbbeeken, line 33: We can now import chai per test file as per usual, but lets do it in another PR

test/register-bson.js
```diff
@@ -1,7 +1,65 @@
...
+
+// Controls whether to run BSON library declaration in an node or "web" environment
+const web = process.env.WEB === 'true';
+console.error(inspect({ web }, { colors: true }));
```
nbbeeken, line 37: Following our driver pattern here of printing env info at the top, not much else to print than whether the web mode is on or not

test/register-bson.js
```diff
@@ -1,7 +1,65 @@
...
+
+// Node.js 16 was when the atob and btoa globals were introduced, so we need replacements for testing on 14
+const shim_btoa = input => Buffer.prototype.toString.call(Buffer.from(input), 'base64');
+const shim_atob = input => Buffer.from(input, 'base64').toString('binary');
```
nbbeeken, line 41: Instead of figuring out skip logic on node 14 it's easier to throw these shims on the context object

test/register-bson.js
```diff
@@ -1,7 +1,65 @@
...
+
+let BSON;
+if (web) {
+  // TODO(NODE-4713): Using the umd for now since it works well as a Node.js import
+  // Switch to the .cjs rollup planned for NODE-4713
+  const filename = path.resolve(__dirname, '../dist/bson.browser.umd.js');
```
nbbeeken, line 47: Using the umd bundle simply because it is easy, but we can change this as the comment notes.

test/register-bson.js
```diff
@@ -1,7 +1,65 @@
...
+  const code = fs.readFileSync(filename, { encoding: 'utf8' });
+  // These are the only globals BSON strictly depends on
+  // an optional global is crypto
```
nbbeeken, line 50: Did not include crypto b/c our BSON library works without it (falling back to Math.random()), but this could be an interesting way to test that fallback explicitly (we are now since I don't include crypto)

rollup.config.mjs
```diff
@@ -1,11 +1,11 @@
...
-import nodeBuiltins from 'rollup-plugin-polyfill-node';
+import commonjs from '@rollup/plugin-commonjs';
 import { nodeResolve } from '@rollup/plugin-node-resolve';
 import { babel } from '@rollup/plugin-babel';
 import typescript from '@rollup/plugin-typescript';
-import nodeGlobals from 'rollup-plugin-node-globals';
 import replace from '@rollup/plugin-replace';
+import { readFile } from 'fs/promises';
+
+const pkg = JSON.parse(await readFile('./package.json', { encoding: 'utf8' }));
```
nbbeeken, line 8: eslint was complaining about the import assertion

src/objectid.ts
```diff
@@ -223,15 +219,15 @@ export class ObjectId {
...
+      return ByteUtils.equals(this.id, ByteUtils.fromISO88591(otherId));
     }
 
     if (typeof otherId === 'string' && ObjectId.isValid(otherId) && otherId.length === 24) {
       return otherId.toLowerCase() === this.toHexString();
     }
 
     if (typeof otherId === 'string' && ObjectId.isValid(otherId) && otherId.length === 12) {
-      return Buffer.from(otherId).equals(this.id);
+      return ByteUtils.equals(ByteUtils.fromISO88591(otherId), this.id);
```
addaleax, line 230: I think this and the similar one above are currently unreachable because of https://github.com/mongodb/js-bson/pull/518/files#r1003675008?

src/parser/serializer.ts
```diff
@@ -935,7 +814,7 @@ export function serializeInto(
...
```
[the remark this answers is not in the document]
nbbeeken: https://jira.mongodb.org/browse/NODE-4771
Nice glad we could catch that, I'll make sure its in the changelog here ^ kinda hard to separate from this PR for TS reasons.

src/utils/node_byte_utils.ts
```diff
@@ -0,0 +1,119 @@
...
+type NodeJsEncoding = 'base64' | 'hex' | 'utf8' | 'binary';
+type NodeJsBuffer = {
+  alloc: (size: number) => Uint8Array;
+  from(array: number[]): Uint8Array;
+  from(array: Uint8Array): Uint8Array;
+  from(array: ArrayBuffer): Uint8Array;
+  from(array: ArrayBuffer, byteOffset: number, byteLength: number): Uint8Array;
+  from(base64: string, encoding: NodeJsEncoding): Uint8Array;
+  write(string: string, offset: number, length: undefined, encoding: 'utf8'): number;
+  copy(target: Uint8Array, targetStart: number, sourceStart: number, sourceEnd: number): number;
```
addaleax, line 14: Not sure how strict you’d like to get with the types here, but these are also prototype methods, not static ones

src/utils/node_byte_utils.ts
```diff
@@ -0,0 +1,119 @@
...
+  toUTF8(buffer) {
+    return bytesToString(buffer, 'utf8');
+  },
+
+  utf8ByteLength(input) {
+    return Buffer.byteLength(input, 'utf8');
+  },
+
+  encodeUTF8Into(buffer, source, byteOffset) {
+    return (buffer as unknown as NodeJsBuffer).write(source, byteOffset, undefined, 'utf8');
```
addaleax, line 102: It’s kind of valid that TS complains here – there’s nothing guaranteeing that `buffer` is actually a Node.js buffer from a TS perspective. Not sure how far you want to go into making the types work as expected, though.

src/parser/serializer.ts
```diff
@@ -42,30 +41,28 @@ export interface SerializeOptions {
 const regexp = /\x00/; // eslint-disable-line no-control-regex
 const ignoreKeys = new Set(['$db', '$ref', '$id', '$clusterTime']);
 
+function setUtf8String(buffer: Uint8Array, stringToSet: string, offset: number): number {
+  const keyAsBytes = ByteUtils.fromText(stringToSet);
+  buffer.set(keyAsBytes, offset);
```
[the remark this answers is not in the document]
nbbeeken: I guess I shouldn't make changes without actually benching. 😅 That makes sense for write ty! but `.set` doesn't let you bound the amount to copy out from the source without creating the subarray first. I'll test and see for myself but any sense if constructing subarrays before .set calls is worth avoiding? 

I'm thinking an extreme case (and probably the only one) is copying a subset of the 17MB serialize buffer to the buffer returned to the caller. 
I'm also making the assumption that the subarray -> set isn't how copy works under the hood 🤔

src/utils/node_byte_utils.ts
nbbeeken: broke up the types to now be a Ctor and an instance, tells more of the truth

src/parser/serializer.ts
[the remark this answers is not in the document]
nbbeeken: https://github.com/nodejs/node/blob/main/lib/buffer.js#L247-L262
Well whaddya know 😁 okay so subarray and set can actually save us some time on all the arg type checking etc

src/utils/web_byte_utils.ts
```diff
@@ -0,0 +1,138 @@
...
+const HEX_DIGIT = /(\d|[a-f])/i;
+
+export const webByteUtils = {
+  toLocalBufferType(
+    potentialUint8array: Uint8Array | ArrayBufferViewWithTag | ArrayBuffer
+  ): Uint8Array {
+    let stringTag = potentialUint8array?.[Symbol.toStringTag];
+    if (typeof stringTag !== 'string') {
+      stringTag = Object.prototype.toString.call(potentialUint8array);
+      stringTag = stringTag.slice(8, stringTag.indexOf(']'));
```
dariakp, line 40: same comment here regarding the tag

src/utils/web_byte_utils.ts
```diff
@@ -0,0 +1,138 @@
...
+  toISO88591(uint8array: Uint8Array): string {
+    return Array.from(Uint16Array.from(uint8array), b => String.fromCharCode(b)).join('');
+  },
+
+  fromHex(hex: string): Uint8Array {
+    const evenLengthHex = hex.length % 2 === 0 ? hex : hex.slice(0, hex.length - 1);
+    const buffer = [];
+
+    for (let i = 0; i < evenLengthHex.length; i += 2) {
+      if (!HEX_DIGIT.test(evenLengthHex[i]) && !HEX_DIGIT.test(evenLengthHex[i + 1])) {
```
dariakp, line 115: is this the same as the node internal implementation? to silently ignore any non-hex portions or whole non-hex strings?

src/utils/web_byte_utils.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  fromBase64(base64: string): Uint8Array {
+    return Uint8Array.from(atob(base64), c => c.charCodeAt(0));
+  },
+
+  toBase64(uint8array: Uint8Array): string {
+    return btoa(webByteUtils.toISO88591(uint8array));
+  },
+
+  fromISO88591(codePoints: string): Uint8Array {
```
dariakp, line 94: can we group the companion `to` method here, and mark them with the TODO ticket to remove?
and can we similarly group the other methods like we do in the node file, so that the `from` and corresponding `to` ones are next to each other?

test/node/to_bson_test.js
```diff
@@ -133,19 +131,24 @@ describe('toBSON', function () {
...
+      if (__isWeb__) {
+        // These prototype modification do not make it
+        // to the BSON lib that was run in a separate context
+        return this.skip();
+      }
       Number.prototype.toBSON = () => 'hello';
       String.prototype.toBSON = () => 'hello';
       Boolean.prototype.toBSON = () => 'hello';
-      if (BigInt) BigInt.prototype.toBSON = () => 'hello';
+      BigInt.prototype.toBSON = () => 'hello';
```
dariakp, line 143: is the purpose of these extensions to just not throw? or is the output checked somewhere? if it *is* checked, then I think it would be better for the toBSON on the different types to return different strings, but otherwise it doesn't matter, though in that case, empty string or void would do just as well, no?

test/node/object_id_tests.js
```diff
@@ -171,18 +172,22 @@ describe('ObjectId', function () {
...
   it('should correctly create ObjectId undefined or null', function () {
     const objNull = new ObjectId(null);
     const objNoArg = new ObjectId();
     const objUndef = new ObjectId(undefined);
-    expect(objNull.id).to.be.instanceOf(Buffer);
-    expect(objNoArg.id).to.be.instanceOf(Buffer);
-    expect(objUndef.id).to.be.instanceOf(Buffer);
+    expect(isBufferOrUint8Array(objNull.id)).to.be.true;
+    expect(isBufferOrUint8Array(objNoArg.id)).to.be.true;
+    expect(isBufferOrUint8Array(objUndef.id)).to.be.true;
```
dariakp, line 190: can we add an expect fail message, since we no longer have an obvious `instanceOf` assertion?

test/register-bson.js
dariakp: Can you file and link a follow up ticket?

test/register-bson.js
dariakp: can you add a TODO to remove when we drop node 14 in V6? actually, can you file that drop 14 ticket in the BSON V6 epic - you can name it generically to "drop unsupported node versions" but then link it here; it'll make clean up either if we identify version-specific logic ahead of time

dariakp (changes requested):
Comments for everything but the byte utils test

src/utils/byte_utils.ts
```diff
@@ -0,0 +1,56 @@
...
+  fromNumberArray: (array: number[]) => Uint8Array;
+  /** Create a Uint8Array from a base64 string */
+  fromBase64: (base64: string) => Uint8Array;
+  /** Create a base64 string from bytes */
+  toBase64: (buffer: Uint8Array) => string;
+  /** Create a Uint8Array from a hex string */
+  fromHex: (hex: string) => Uint8Array;
+  /** Create a hex string from bytes */
+  toHex: (buffer: Uint8Array) => string;
+  /** @deprecated binary strings are a legacy method of data transfer */
```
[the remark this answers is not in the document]
nbbeeken: We may not remove these since they are necessary for base64 encoding on the web side (at least with the current implementation), so I've mentioned the ticket, the `@deprecation` was to get the strike through so we don't add more usage of this encoding. But that's a lot of implied meaning 😅 so let me just speak it in the comment.

src/utils/node_byte_utils.ts
```diff
@@ -0,0 +1,103 @@
...
+        potentialBuffer.buffer,
+        potentialBuffer.byteOffset,
+        potentialBuffer.byteLength
+      );
+    }
+
+    let stringTag = potentialBuffer?.[Symbol.toStringTag];
+    if (typeof stringTag !== 'string') {
+      stringTag = Object.prototype.toString.call(potentialBuffer);
+      stringTag = stringTag.slice(8, stringTag.indexOf(']'));
```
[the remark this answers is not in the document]
nbbeeken: Good call, I'm actually taking a step further, it is only exactly 4 cases and being a hot code path I want to keep the conditions streamlined, we can just type out the `'[object X]'` cases.

test/node/byte_utils.test.ts
```diff
@@ -0,0 +1,479 @@
...
+    }
+  }
+];
+const fromBase64Tests: ByteUtilTest<'fromBase64'>[] = [
+  {
+    name: 'should create buffer from base64 input',
+    inputs: [Buffer.from('abc\u{1f913}', 'utf8').toString('base64')],
+    expectation({ web, output, error }) {
+      if (!isNode16OrLater && web) {
+        // Skip reason: btoa and atob were not made globals until node 16
```
dariakp, line 192: for all the cases in this file that are skipped, is there a way to actually skip them or to alter the test case name to indicate what is actually tested; because here, for example, it seems that we actually won't be able to create the buffer from base 64 on Node 14 - is that true? or is it a test setup issue only?

src/utils/web_byte_utils.ts
nbbeeken: Yep I should have tests to demonstrate this, I actually didn't look at the node implementation, but the tests demonstrate it matches, node always ignores the last character in an uneven lengthed input. And Node will stop short when it reaches a character that is not hex. I assume the idea is to mirror parseInt, which does similiar early stoping.

test/node/byte_utils.test.ts
```diff
@@ -0,0 +1,479 @@
...
+      expect(error).to.be.null;
+      expect(output).to.have.property('byteLength', 0);
+    }
+  },
+  {
+    name: 'should use leading valid hex characters',
+    inputs: ['abxxcd'],
+    expectation({ output, error }) {
+      expect(error).to.be.null;
+      expect(output).to.deep.equal(Buffer.from('abxxcd', 'hex'));
```
dariakp, line 263: would this also be the same as `Buffer.from('ab')`?

test/node/byte_utils.test.ts
```diff
@@ -0,0 +1,479 @@
...
+    }
+  },
+  {
+    name: 'should slice input strings down to nearest even length (len: 1)',
+    inputs: ['a'],
+    expectation({ output, error }) {
+      expect(error).to.be.null;
+      expect(output).to.deep.equal(Buffer.from('a', 'hex'));
```
dariakp, line 271: it's not clear how the test is testing the "slice down to nearest even length" since it's using the result of `Buffer.from` in the assertion on the exact same string; and I have a similar comment for the case below that, too

test/node/byte_utils.test.ts
```diff
@@ -0,0 +1,479 @@
...
+    }
+  },
+  {
+    name: 'should slice input strings down to nearest even length (len: 5)',
+    inputs: ['abcde'],
+    expectation({ output, error }) {
+      expect(error).to.be.null;
+      expect(output).to.deep.equal(Buffer.from('abcde', 'hex'));
+    }
```
dariakp, line 280: I think it would be worth adding cases for a string with:
1) no valid hex characters `xx`
2) a string containing a valid hex symbol in the odd position immediately followed by an invalid one `abax`
3) a string containing a valid hex symbol in the even position immediately preceded by an invalid one `abxa`

src/utils/web_byte_utils.ts
nbbeeken: Yea they should be made to be the same order, which should also match the order in the `ByteUtils` type now.

test/node/to_bson_test.js
nbbeeken: It is being checked on line 157 inside `testToBSONFor` this is a bit of a table test without the loop, I was removing the if stmt here since we're testing on node versions that always have BigInt so it was never not testing the BigInt case. But I think that's a good enhancement to the test!

test/node/byte_utils.test.ts
```diff
@@ -0,0 +1,479 @@
...
+  ['toHex', toHexTests],
+  ['fromISO88591', fromISO88591Tests],
+  ['toISO88591', toISO88591Tests],
+  ['fromUTF8', fromUTF8Tests],
+  ['toUTF8', toUTF8Tests],
+  ['utf8ByteLength', utf8ByteLengthTests]
+]);
+
+describe('ByteUtils', () => {
+  it('should be set to the nodeJsByteUtils', () => {
```
dariakp, line 414: I thought it got set interchangeably depending on the env?

dariakp (changes requested):
test review

test/node/object_id_tests.js
nbbeeken: Yes, added to all the cases where we use this helper

test/register-bson.js
nbbeeken: https://jira.mongodb.org/browse/NODE-4786

test/register-bson.js
nbbeeken: done!

src/utils/web_byte_utils.ts
nbbeeken: done!

src/utils/node_byte_utils.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  fromBase64(base64: string): NodeJsBuffer {
+    return Buffer.from(base64, 'base64');
+  },
+
+  toBase64(buffer: Uint8Array): string {
+    return nodeJsByteUtils.toLocalBufferType(buffer).toString('base64');
+  },
+
+  fromISO88591(codePoints: string): NodeJsBuffer {
```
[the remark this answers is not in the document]
nbbeeken: done!

test/node/byte_utils.test.ts
nbbeeken: Added cases (and fixed bug that they revealed!)

test/node/byte_utils.test.ts
```diff
@@ -0,0 +1,514 @@
...
+    expectation({ output, error }) {
+      expect(error).to.be.null;
+      expect(output).to.deep.equal(Buffer.from('abcx', 'hex'));
+      expect(output).to.have.property('byteLength', 1);
+      expect(output).to.have.property('0', 0xab);
+    }
+  },
+  {
+    name: 'should ignore double digit hex subsequence that starts with invalid character',
+    inputs: ['abcx'],
```
dariakp, line 294: wrong test data here, should be `abxc`

docs/upgrade-to-v5.md
```diff
@@ -0,0 +1,37 @@
...
+1. a brief statement of what is breaking (brief as in "x will now return y instead of z", or "x is no longer supported, use y instead", etc
+2. a brief statement of why we are breaking it (bug, not useful, inconsistent behavior, better alternative, etc)
+3. if applicable, an example of suggested syntax change (can be included in (1) )
+-->
+
+### Remove reliance on Node.js Buffer
+
+> **TL;DR**: Web environments return Uint8Array; Node.js environments return Buffer
+
+For those that use the BSON library on Node.js, there is no change the BSON APIs will still return and accept instances of Node.js Buffer. Since we no longer depend on the Buffer web shim for compatibility with browsers, in non-Node.js environments a Uint8Array will be returned instead.
```
dariakp, line 18: can we mention which APIs this actually affects? and can we say in the summary instead of what? (i.e. `<what API is affected> in web environments return Uint8Array instead of <...>`

docs/upgrade-to-v5.md
```diff
@@ -0,0 +1,37 @@
...
+
+This allows the BSON library to be better at platform independence while keeping its behavior consistent cross platform. The Buffer shim served the library well but brought in more than was necessary for the concerns of the code here.
+
+### `ObjectId.toString` / `UUID.toString` / `Binary.toString`
+
+> **TL;DR**: These `toString` methods only support the following encodings: 'hex', 'base64', 'utf8'
+
+The methods: `ObjectId.toString`, `UUID.toString`, and `Binary.toString` took encodings that were passed through to the Node.js Buffer API. As a result of no longer relying on the presence of `Buffer` we can no longer support every encoding that Node.js did. We continue to support `'hex'` and `'base64'` on all three methods and additionally `'utf-8' | 'utf8'` on `Binary.toString`. If any of the other encodings are desired the underlying buffer for all these classes are publicly accessible and while in Node.js will be stored as a Node.js buffer:
```
dariakp, line 26: any notable examples that are no longer supported? (at least one?)

docs/upgrade-to-v5.md
```diff
@@ -0,0 +1,37 @@
...
+
+```typescript
+const oid = new ObjectId();
+oid.id.toString('utf16le') // Returns string interpreting the bytes as 'utf16le'
```
dariakp, line 30: what is this an example of? what did it return before? can we be clear about what this is demonstrating?

docs/upgrade-to-v5.md
durran, line 18:
```suggestion
For those that use the BSON library on Node.js, there is no change - the BSON APIs will still return and accept instances of Node.js Buffer. Since we no longer depend on the Buffer web shim for compatibility with browsers, in non-Node.js environments a Uint8Array will be returned instead.
```

durran (changes requested):

docs/upgrade-to-v5.md
nbbeeken: done!

docs/upgrade-to-v5.md
nbbeeken: Not really, the other encodings generally don't make sense for ObjectId or UUID, for Binary (since it's black box bytes) it could be used to store text formats other than utf8, but it would be arbitrary to list any subset. But I've linked to node's list of encodings now

test/node/byte_utils.test.ts
nbbeeken: fixed

src/utils/node_byte_utils.ts
```diff
@@ -0,0 +1,108 @@
...
+    if (ArrayBuffer.isView(potentialBuffer)) {
+      return Buffer.from(
+        potentialBuffer.buffer,
+        potentialBuffer.byteOffset,
+        potentialBuffer.byteLength
+      );
+    }
+
+    const stringTag =
+      potentialBuffer?.[Symbol.toStringTag] ?? Object.prototype.toString.call(potentialBuffer);
```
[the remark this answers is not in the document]
nbbeeken: Yea I don't see a way to cover this explicitly, if I manipulate the input to not have a string tag it will also break the Object.prototype.toString call since that is how it builds the result of that call

docs/upgrade-to-v5.md
nbbeeken: https://jira.mongodb.org/browse/NODE-4794

That list would be looking at all the public APIs that changed in this PR and listing them out, in the interest of moving this along we will be cleaning up this guide when we reach the end of the EPIC and the diff of this PR will still be available to us then for reference. So let's defer to NODE-4794, I've added a note about creating a list of APIs changed.

src/utils/node_byte_utils.ts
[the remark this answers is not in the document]
nbbeeken: @addaleax Any background knowledge on this subject of whether we should keep the fallback proto.ToString or just use the toStringTag getter and fail if it's not present?

docs/upgrade-to-v5.md
dariakp: this isn't about making sure we have a well documented list, it's about knowing what these *specific* code changes actually affect, so I don't think it's out of scope of this PR to identify what will be changing. This work will still need to be done, and it is easier to do it piecemeal with the corresponding code changes than in one big block, where we might miss something that was affected due to the sheer volume of changes.

docs/upgrade-to-v5.md
```diff
@@ -15,19 +15,28 @@ for nodejs and web platforms.
 
 > **TL;DR**: Web environments return Uint8Array; Node.js environments return Buffer
 
-For those that use the BSON library on Node.js, there is no change the BSON APIs will still return and accept instances of Node.js Buffer. Since we no longer depend on the Buffer web shim for compatibility with browsers, in non-Node.js environments a Uint8Array will be returned instead.
+For those that use the BSON library on Node.js, there is no change - the BSON APIs will still return and accept instances of Node.js Buffer. Since we no longer depend on the Buffer web shim for compatibility with browsers, in non-Node.js environments a Uint8Array will be returned instead.
 
 This allows the BSON library to be better at platform independence while keeping its behavior consistent cross platform. The Buffer shim served the library well but brought in more than was necessary for the concerns of the code here.
 
 ### `ObjectId.toString` / `UUID.toString` / `Binary.toString`
```
dariakp, line 22:
```suggestion
### Restrict supported encodings in `ObjectId.toString` / `UUID.toString` / `Binary.toString`
```

docs/upgrade-to-v5.md
```diff
@@ -15,19 +15,28 @@ for nodejs and web platforms.
...
 
 > **TL;DR**: These `toString` methods only support the following encodings: 'hex', 'base64', 'utf8'
```
dariakp, line 24: if the 'utf8' encoding isn't supported on `ObjectId.toString` or `UUID.toString`, then this summary is incorrect

I actually think that what you currently have as a "tl;dr" can be simply the sub-heading for the corresponding set of changes. I think it would also make the table of contents easier to generate. In the case of this TL;DR, it can be removed entirely if you update L22 the way I am suggesting; then the paragraph underneath can elaborate on the specifics

Then the TL;DR of the section above (L16) can be promoted to a subheading, with both of these sections being under the title "Remove reliance on Node.js buffer", which would be the grouping heading.

So in this guide, the grouping headings will indicate the themes of the changes, and the subheadings in the groups will be the actual breaking change from the user's perspective. (What I mean is, "remove reliance on nodejs buffer" isn't the breaking change, it's the cause of the breaking changes whose summaries follow)

As for the 3rd example, I think we can group it under "Other" for now

dariakp (changes requested):

src/utils/node_byte_utils.ts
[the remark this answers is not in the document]
addaleax: @nbbeeken My intuition would have been to keep the `Object.prototype.toString.call`, it’s imo the more idiomatic variant and otherwise these two should just be equivalent, yeah.

lucsoft (approved):
I hope this is not off-topic, but as a third party maintainer of the deno_mongo project (also web_bson), this is awesome!

dariakp (approved):

PR #521, opened 2022-11-21
test(NODE-4803): add test for toStringTag fallback

nbbeeken:
#### What is changing?

Adds a test for toStringTag fallback to Object.prototype.toString.call logic.

Tests that we don't change the behavior that is carefully crafted for unusual or legacy js systems.


test/node/byte_utils.test.ts
```diff
@@ -465,6 +476,14 @@ describe('ByteUtils', () => {
         expect(bufferOut).to.be.an.instanceOf(Buffer);
         expect(bufferOut.buffer).to.equal(arrayBufferIn);
       });
+
+      it('should attempt fallback to Object.prototype.toString call if toStringTag does not exist on ArrayBuffer', () => {
```
dariakp, line 480: I would prefer we check this more directly via seeing if Object.prototype.toString is called

dariakp (changes requested):

dariakp (changes requested):
Let's add a happy path test, too

dariakp (approved):

dariakp (approved):

PR #523, opened 2022-11-21
feat!(NODE-4712): remove unused Map polyfill

nbbeeken:
#### What is changing?

Removes unused Map implementation, and removes the export.
Adds an entry point test modeled after the one we have in our driver: [here](https://github.com/mongodb/node-mongodb-native/blob/main/test/unit/index.test.ts)

Migration guide updated.

ES Map is available on all supported runtimes.


test/node/tools/utils.js
```diff
@@ -176,3 +176,17 @@ const getSymbolFrom = function (target, symbolName, assertExists) {
 };
 
 exports.getSymbolFrom = getSymbolFrom;
+
+module.exports.byStrings = (a, b) => {
+  const res = `${a}`.localeCompare(`${b}`);
+  return res < 0 ? -1 : res > 0 ? 1 : 0;
```
addaleax, line 182: Nothing wrong with this, but why not …?

```suggestion
  return `${a}`.localeCompare(`${b}`);
```

addaleax (dismissed):

test/node/tools/utils.js
nbbeeken: It's copy pasta-ed from our driver where this is written in TS and localeCompare returns number not the required union of `0 | 1 | -1` so that's why we have this little coercion ternary. it would clean it up but we've copied this function in 3 places (its in mongodb-legacy too) so might as well keep it identical despite the extra logic

addaleax: Interesting! I’d be curious to know more, the TS definition for `Array.prototype.sort()` is `sort(compareFn?: (a: T, b: T) => number): this`, so it should work out fine to use `localeCompare` directly in the driver too? But obviously this is getting into more detail than necessary :)

nbbeeken: No totally good to keep me honest here about motivations. I think this strict union requirement was inspired by the bug we had to fix across all drivers relating to setVersion/electionId comparison, for [our implementation in particular](https://github.com/mongodb/node-mongodb-native/blob/main/src/sdam/topology_description.ts#L377-L381) we have this carefully written logic for determining greaterThan/lessThan/equal that's easier to follow when you can name variable representing `=== 0` or `=== -1`. (of course we could've still made those variables using `> 0` or `< -1`) Now I've just brought that TS requirement with me when writing compare functions for other use cases since it proved useful there. It is a burden brought on by own design 😭 😅 

Personal anecdote, compare functions and I have never been friends, going back to my OS class I wrote my [compare function to only return true or false](https://gitlab.com/moon-project/moon/-/blob/develop/libc/string.c#L18-27) 🫣 since (for whatever reason) I believed that to be more logical than `-1,1,0` after all isn't equality all we care about?? (of course not) thus breaking our libc implementation 😭

test/node/exports.test.ts
```diff
@@ -0,0 +1,82 @@
+import * as BSON from '../register-bson';
+import { sorted, byStrings } from './tools/utils';
+
+const EXPECTED_EXPORTS = [
+  // This is our added web indicator not a real export but a small exception for this test.
```
baileympearson, line 5: Can you explain further why this is necessary?  I don't understand why __isWeb__ is added for our tests

test/node/map.test.ts
```diff
@@ -0,0 +1,68 @@
+import { expect } from 'chai';
+import * as BSON from '../register-bson';
+
+describe('ES Map support in serialize()', () => {
+  it('should serialize a Map to BSON document', () => {
```
baileympearson, line 5: How thoroughly do we want to test this functionality?  The tests cases seem random (a map with one key, nested map and numbered keys?).  I'd argue that if, for the purposes of this PR, we only care that we still support maps, then it should be sufficient to just serialize an empty map and confirm 1) there's no error thrown and 2) it serializes as a document type.  If we want to add testing to our map serialization, then these test cases feel insufficient and random.

Also, the nested document test feels random.  If we test nested maps, shouldn't we test all valid nesting scenarios we can think of?  A map nested in an array or an array in a map in an array, ect?  My feeling is that it's probably unnecessary to test these nested scenarios at all, but I'm curious to hear others' opinions.

Lastly, the test for ordering of keys feels insufficient.  If we test that case, shouldn't we test that insertion order is respected for the following cases (and maybe more, if we can think of more):

- numeric keys
- string keys
- numeric and string keys
- string keys that are numeric (i.e., "1", "2")
- numeric keys and string keys that are numeric

---

I'm curious what @dariakp thinks too

baileympearson (changes requested):

test/node/exports.test.ts
nbbeeken: It is to identify when the export was created in an environment without Buffer (web-like, only TextEncoder/Decoder and btoa/atob globals). So we can change assertions based on that condition: for example, in the `error_test.js` file we can check instanceof on Node and `.name` prop in the web.

test/node/map.test.ts
dariakp: Well, we should determine what the *purpose* of the tests is (i.e., what is *important* for this functionality to do); if it's a sanity check, then an empty map and a non-empty map would be sufficient (empty is always an edge case, so I wouldn't advise testing solely with empty input). We should also look at the removed tests and determine if any of those are worth preserving (e.g., the object that looks like a map). 

To help determine the purpose, we can see what it is we are actually implementing on top of what the built in construct already does and look for any branching conditions. I can't give more concrete guidance without digging through the code, hopefully @nbbeeken has enough context to answer these questions.

nbbeeken: A lot more thoroughly than we currently do, we have some smattering of Map tests already in the bson_test.js file, my goal was to bring us some organized sanity checks using our more modern testing practices. So we don't need new tests at all since we do have the coverage over there. I kept them simple to make the bytewise assertions reasonable, but they are arbitrary, including the nested Map test, it was only inspired by being an "interesting" case of using Maps.

What we really need to file a ticket for is fuzz testing of some kind that is much more structured about generating arbitrarily shaped objects/maps/arrays with all various types, which I can do!

The key ordering test is specifically trying to prove the use case we've relied on Maps for (the ability to keep numeric keys in chronological order) within the driver code (sorts and index definitions). This test provides an early warning if that were to change as we work more on the BSON library. So yes we should test insertion order for all those things, with some knowledge that there's diminishing return since we'd be testing ES Map behavior. The stringified numeric keys is an "interesting" case worth keeping here since it is so easy to break, as soon as the keys from the Map land in an object we've lost the ordering information. (ex. `Object.fromEntries(map)` seemingly harmless line of code actually can have a big impact on BSON output)

docs/upgrade-to-v5.md
```diff
@@ -50,3 +50,7 @@ TODO(NODE-4771): serializeFunctions bug fix makes function names outside the asc
 We have set our typescript compilation target to `es2020` which aligns with our minimum supported Node.js version 14+. The following is from the typescript release notes on es2020 support, so it's some of the syntax that can be expected to be preserved after compilation:
 
 > This will preserve newer ECMAScript 2020 features like optional chaining, nullish coalescing, export * as ns, and dynamic import(...) syntax. It also means bigint literals now have a stable target below esnext.
+
+### Remove `Map` export
+
+BSON no longer polyfills es Map and so the export was removed. The removal should reveal any instances where the global `Map` constructor should be used in place of the one BSON used to export.
```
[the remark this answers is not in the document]
nbbeeken: Indeed, I was saying more than necessary here, clarified

test/node/map.test.ts
baileympearson: I was thinking along the lines of what Daria said - that just simple cases would suffice for this PR (thanks for adding the empty document test case).  I'm okay with extra testing too.

Also - thanks for adding the bson fuzzing ticket

baileympearson (approved):

test/node/map.test.ts
```diff
@@ -0,0 +1,76 @@
...
+  it('serialize a empty Map to BSON document - { }', () => {
+    const map = new Map();
+    const emptyBSON = new Uint8Array([5, 0, 0, 0, 0]);
+    const result = BSON.serialize(map);
+    expect(result).to.have.property('byteLength', 5);
+    expect(result, 'byteLength must be 5 followed by null terminator').to.deep.equal(emptyBSON);
+  });
+
+  it('serialize a Map with one key to BSON document - { a: 2 }', () => {
+    const map = new Map([['a', new BSON.Int32(2)]]);
```
dariakp, line 14: per slack discussion, let's update the assertions to be easier to read by providing an annotated `bufferFromHexArray` instead of a byte by byte sequence of assertions

dariakp (changes requested):

test/node/map.test.ts
nbbeeken: done!

docs/upgrade-to-v5.md
```diff
@@ -79,3 +79,8 @@ We have set our typescript compilation target to `es2020` which aligns with our
 > **TL;DR**: TODO
 
 TODO(NODE-4771): serializeFunctions bug fix makes function names outside the ascii range get serialized correctly
+> This will preserve newer ECMAScript 2020 features like optional chaining, nullish coalescing, export * as ns, and dynamic import(...) syntax. It also means bigint literals now have a stable target below esnext.
+
+### Remove `Map` export
+
+This library no longer polyfills [ES Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) and the export "Map" was removed. Users should migrate to using the global Map constructor available in all supported JS environments.
```
baileympearson, line 86:
```suggestion
This library no longer polyfills [ES Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) and the export "Map" was removed. If users are instantiating the exported `Map` object from the BSON library instead of the global map object, users should migrate to using the global Map constructor available in all supported JS environments.
```

baileympearson (approved):

dariakp (approved):

PR #525, opened 2022-11-22
feat!(NODE-4704): remove deprecated ObjectId methods

nbbeeken:
#### What is changing?

Removing deprecated ObjectId methods.

Methods are deprecated, alternatives exist.

### DIFF
--- a/docs/upgrade-to-v5.md
+++ b/docs/upgrade-to-v5.md
@@ -93,12 +93,17 @@ This library no longer polyfills [ES Map](https://developer.mozilla.org/en-US/do
 The following deprecated methods have been removed:
 
 - `ObjectId.prototype.generate`
-  - Instead, generate a new ObjectId with the constructor: `new ObjectId()`
+  - Instead, generate a new ObjectId with the constructor: `new ObjectId()` or using the `static generate(time?: number)` method.
+- `ObjectId.prototype.generationTime`
+  - Instead, use `static createFromTime()` and `getTimestamp()` to set and inspect these values on an `ObjectId()`
 
 - `ObjectId.prototype.getInc`
 - `ObjectId.prototype.get_inc`
 - `ObjectId.get_inc`
-  - The `static getInc()` is private now since it has a side effect of changing the value of the next `ObjectId` generated, using `new ObjectId()` and inspecting the increment section of the bytes can provide insight into the current increment value.
+  - The `static getInc()` is private since invoking it increments the next `ObjectId` index, instead users should inspect the counter on newly created ObjectId if it is needed.
+  - See the following code snippet for how to obtain the counter field of an ObjectId:
 
-- `ObjectId.prototype.generationTime`
-  - Instead, use `static createFromTime()` and `getTimestamp()` to set and inspect these values on an `ObjectId()`
+```ts
+const oid = new ObjectId();
+const counterField = oid.id[9] << 16 | oid.id[10] << 8 | oid.id[11];
+```
--- a/docs/upgrade-to-v5.md
+++ b/docs/upgrade-to-v5.md
@@ -87,3 +87,18 @@ This library no longer polyfills [ES Map](https://developer.mozilla.org/en-US/do
 ### `Decimal128` `toObject()` mapper support removed
 
 `Decimal128` can no longer have a `toObject()` method added on to its prototype for mapping to a custom value. This feature was undocumented and inconsistent with the rest of our BSON types. At this time there is no direct migration: cursors in the driver support transformations via `.map`, otherwise the `Decimal128` instances will require manual transformation. There is a plan to provide a better mechanism for consistently transforming BSON values tracked in [NODE-4680](https://jira.mongodb.org/browse/NODE-4680), please feel free to add a vote or comment with a use case to help us land the feature in the most useful form.
+
+### Remove deprecated ObjectId methods
+
+The following deprecated methods have been removed:
+
+- `ObjectId.prototype.generate`
+  - Instead, generate a new ObjectId with the constructor: `new ObjectId()`
+
+- `ObjectId.prototype.getInc`
+- `ObjectId.prototype.get_inc`
+- `ObjectId.get_inc`
+  - The `static getInc()` is private now since it has a side effect of changing the value of the next `ObjectId` generated, using `new ObjectId()` and inspecting the increment section of the bytes can provide insight into the current increment value.
+
+- `ObjectId.prototype.generationTime`
+  - Instead, use `static createFromTime()` and `getTimestamp()` to set and inspect these values on an `ObjectId()`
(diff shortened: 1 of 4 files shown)


test/node/object_id_tests.js
```diff
@@ -9,18 +9,30 @@ const getSymbolFrom = require('./tools/utils').getSymbolFrom;
 const isBufferOrUint8Array = require('./tools/utils').isBufferOrUint8Array;
 
 describe('ObjectId', function () {
-  it('should correctly handle objectId timestamps', function (done) {
-    const a = ObjectId.createFromTime(1);
-    expect(Buffer.from([0, 0, 0, 1])).to.deep.equal(a.id.slice(0, 4));
-    expect(1000).to.equal(a.getTimestamp().getTime());
+  describe('static createFromTime()', () => {
+    it('should create an objectId with user defined value in the timestamp field', function () {
```
durran, line 13:
```suggestion
    it('creates an objectId with user defined value in the timestamp field', function () {
```

test/node/object_id_tests.js
```diff
@@ -9,18 +9,30 @@ const getSymbolFrom = require('./tools/utils').getSymbolFrom;
...
+    });
+  });
 
-    const b = new ObjectId();
-    b.generationTime = 1;
-    expect(Buffer.from([0, 0, 0, 1])).to.deep.equal(b.id.slice(0, 4));
-    expect(1).to.equal(b.generationTime);
-    expect(1000).to.equal(b.getTimestamp().getTime());
+  describe('getTimestamp()', () => {
+    it('should fetch the big endian int32 leading the Oid and create a Date instance', function () {
```
durran, line 22:
```suggestion
    it('fetches the big endian int32 leading the Oid and create a Date instance', function () {
```

test/node/object_id_tests.js
```diff
@@ -9,18 +9,30 @@ const getSymbolFrom = require('./tools/utils').getSymbolFrom;
...
+      const a = new ObjectId('00000002' + '00'.repeat(8));
+      expect(a.id.slice(0, 4)).to.deep.equal(Buffer.from([0, 0, 0, 2]));
+      expect(Object.prototype.toString.call(a.getTimestamp())).to.equal('[object Date]');
+      expect(a.getTimestamp()).to.deep.equal(new Date(2 * 1000));
+      expect(a.getTimestamp().getTime()).to.equal(2000);
+    });
+  });
 
-    done();
+  it('should create an objectId with user defined value in the timestamp field', function () {
```
durran, line 31:
```suggestion
  it('creates an objectId with user defined value in the timestamp field', function () {
```

durran (changes requested):
Just minor test description changes per our test patterns.

durran (approved):

docs/upgrade-to-v5.md
```diff
@@ -50,3 +50,18 @@ TODO(NODE-4771): serializeFunctions bug fix makes function names outside the asc
...
+
+The following deprecated methods have been removed:
+
+- `ObjectId.prototype.generate`
+  - Instead, generate a new ObjectId with the constructor: `new ObjectId()`
+
+- `ObjectId.prototype.getInc`
+- `ObjectId.prototype.get_inc`
+- `ObjectId.get_inc`
+  - The `static getInc()` is private now since it has a side effect of changing the value of the next `ObjectId` generated, using `new ObjectId()` and inspecting the increment section of the bytes can provide insight into the current increment value.
```
dariakp, line 64: I think it's worth putting this change on its own line, because it's another breaking-ish change; and if we are making the users do this manually now, we should have sample code pointing out how

dariakp (changes requested):

src/objectid.ts
```diff
@@ -33,7 +33,7 @@ export class ObjectId {
   }
 
   /** @internal */
-  static index = Math.floor(Math.random() * 0xffffff);
+  private static index = Math.floor(Math.random() * 0xffffff);
```
baileympearson, line 36: love the user of ts-private variables.  is this a new pattern we want to use?

docs/upgrade-to-v5.md
```diff
@@ -50,3 +50,18 @@ TODO(NODE-4771): serializeFunctions bug fix makes function names outside the asc
 We have set our typescript compilation target to `es2020` which aligns with our minimum supported Node.js version 14+. The following is from the typescript release notes on es2020 support, so it's some of the syntax that can be expected to be preserved after compilation:
 
 > This will preserve newer ECMAScript 2020 features like optional chaining, nullish coalescing, export * as ns, and dynamic import(...) syntax. It also means bigint literals now have a stable target below esnext.
+
+### Remove deprecated ObjectId methods
+
+The following deprecated methods have been removed:
+
+- `ObjectId.prototype.generate`
+  - Instead, generate a new ObjectId with the constructor: `new ObjectId()`
```
dariakp, line 59: in the ticket description, we mention the static generate, and it's also the suggestion in the corresponding deprecation message

dariakp (changes requested):

src/objectid.ts
baileympearson: nvm it looks like we're already using it.  But i love it anyways

nbbeeken: Yea seems worth applying where it's relevant, this was internal already so hidden from users (with types) this hides it from the rest of the BSON library too

docs/upgrade-to-v5.md
nbbeeken: Sure thing, and just to capture here, the reason I'm leaving this as something to be done manually is inspired by the spec: https://github.com/mongodb/specifications/blob/master/source/objectid.rst#counter it states drivers MUST not have an accessor for this field.

docs/upgrade-to-v5.md
nbbeeken: updated

docs/upgrade-to-v5.md
```diff
@@ -79,8 +79,27 @@ We have set our typescript compilation target to `es2020` which aligns with our
...
+
+- `ObjectId.prototype.getInc`
+- `ObjectId.prototype.get_inc`
+- `ObjectId.get_inc`
+  - The `static getInc()` is private since invoking it increments the next `ObjectId` index, instead users should inspect the counter on newly created ObjectId if it is needed.
+  - See the following code snippet for how to obtain the counter field of an ObjectId:
+
+```ts
+const oid = new ObjectId();
+const counterField = oid.id[9] << 16 | oid.id[10] << 8 | oid.id[11];
```
baileympearson, line 104: completely optional - should we consider just adding a simple getter that does this for users?

```typescript
  getCounter() : number { return this.id[9] << 16 | this.id[10] << 8 | this.id[11]; }
```

nbbeeken: Per the spec we should not have one: https://github.com/mongodb/specifications/blob/master/source/objectid.rst#counter

baileympearson (dismissed):

dariakp (changes requested):
in the spirit of the spec, I think we shouldn't include the instructions for obtaining the counter (especially if we ever want to change the internal implementation); otherwise lgtm

dariakp (dismissed):

dariakp (approved):

baileympearson (approved):

PR #527, opened 2022-11-22
feat!(NODE-4410): only enumerate own properties

nbbeeken:


test/node/extended_json.test.ts
```diff
@@ -784,4 +791,12 @@ Converting circular structure to EJSON:
       expect(parsedUUID).to.deep.equal(expectedResult);
     });
   });
+
+  it('should only enumerate own property keys from input objects', () => {
+    const input = { a: 1 };
+    Object.setPrototypeOf(input, { b: 2 });
+    const string = EJSON.stringify(input);
+    expect(string).to.include(`"a":`);
+    expect(string).to.not.include(`"b":`);
```
baileympearson, line 800: same comment here as in serializer.test.ts

baileympearson (changes requested):

src/parser/deserializer.ts
```diff
@@ -314,23 +314,16 @@ function deserializeObject(
...
-        arrayOptions = {};
-        for (const n in options) {
-          (
-            arrayOptions as {
-              [key: string]: DeserializeOptions[keyof DeserializeOptions];
-            }
-          )[n] = options[n as keyof DeserializeOptions];
-        }
-        arrayOptions['raw'] = true;
+        arrayOptions = { ...options, raw: true };
```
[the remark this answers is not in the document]
nbbeeken: The issue is related because it has similar surprising side effects when there are keys that you didn't specify yourself effecting the outcomes of your deserialize calls. I tried demonstrating the possible buggy behavior in the test in `test/node/parser/deserializer.test.ts`. If the global object prototype has been polluted with unexpected keys they can control the behavior of the BSON library without actually needing access to the options object a user passes in.

test/node/extended_json.test.ts
nbbeeken: This is a little different, I can hardcode the JSON as is done in other tests here but the issue I've seen crop up a number of times when working on EJSON is the spacing changes which doesn't actually mean the format really changed. Currently this test isn't space sensitive, ideally I'd like us to add a JSON equal checker that is better at being agnostic about that stuff, we'll likely need one when we fix the EJSON corpus issues.

src/parser/deserializer.ts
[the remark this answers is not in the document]
baileympearson: I see.  This impacts all other options too, no? 

```typescript
import { deserialize, serialize } from "./src/bson";

const bytes = serialize({ someKey: { 
    nested: "value"
} });

const opts = {} 
Object.setPrototypeOf(opts, { raw: true });

const result = deserialize(bytes, opts);

console.log(result); 
// { someKey: <Buffer 17 00 00 00 02 6e 65 73 74 65 64 00 06 00 00 00 76 61 6c 75 65 00 00> }
```

It seems like what we'd want to do is only look for own properties on the options objects too when accessing individual fields (not in scope for this work).  My 2-cents would be to remove this from this PR and file a follow up to only consider options that are set as own properties (probably an easy fix by making a shallow clone of the options before we begin any work deserializing / serializing).

If you want to leave this in this PR, that's also okay because technically the scope of the ticket is "Use Object.keys to enumerate keys on a JS object passed to BSON", and this is an instance where we're using a for-in loop.  If we leave this work in this PR, we should still file a follow up ticket I think.  And I'll comment below, but I think we can make the test easier to understand.

test/node/parser/deserializer.test.ts
```diff
@@ -0,0 +1,13 @@
...
+import { expect } from 'chai';
+
+describe('deserializer()', () => {
+  it('should only enumerate own property keys from input options', () => {
+    const bytes = BSON.serialize({ someKey: [1] });
+    const options = { fieldsAsRaw: { someKey: true } };
+    Object.setPrototypeOf(options, { promoteValues: false });
+    const result = BSON.deserialize(bytes, options);
+    expect(result).to.have.property('someKey').that.is.an('array');
+    expect(result.someKey[0]).to.not.have.property('_bsontype', 'Int32');
```
baileympearson, line 11: related to the comment above - this test kind of confusing.  It's not immediately apparent why `fieldsAsRaw` is necessary when looking at the test.  I had to read the code to realize that we only use the for-in loop when `fieldsAsRaw` is set to true, so we need that option to trigger the offending code.

```typescript
context('when the fieldsAsRaw options is present and has a value that corresponds to a key in the object', () => {
  it('ignores non-own properties set on the options object', () => {
	const bytes = BSON.serialize({ someKey: [1] });
    const options = { fieldsAsRaw: { someKey: true } };
    Object.setPrototypeOf(options, { promoteValues: false });
    const result = BSON.deserialize(bytes, options);
    expect(result).to.have.property('someKey').that.is.an('array');
    expect(result.someKey[0], 'expected promoteValues option set on options object prototype to be ignored, but it was not').to.not.be.instanceOf(Int32);
  });
});
```

src/parser/deserializer.ts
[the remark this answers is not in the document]
baileympearson: Works for me.  I left a comment about the test below

test/node/extended_json.test.ts
baileympearson: So stringified ejson is valid json right?  The concern is circularly testing our bson library by using ejson.parse to test ejson.stringify and vice-versa.  But could we use JSON.parse?

```typescript
const input = { a: 1 };
Object.setPrototypeOf(input, { b: 2 });
const string = EJSON.stringify(input);
const parsed = JSON.pase(string)
expect(parsed).to.deep.equal({ a: 1 });
```

nbbeeken: Yea for this test that's fine, I'll update. but generally JSON.parse is lossy in certain scenarios that do matter for EJSON/BSON correctness, so it's not always viable

baileympearson: interesting, do you have an example?

nbbeeken: I kept the assertion that checks the b string doesn't exist, since that's what we really care about seems worth keeping, and of course json parse + deep equal confirms it

nbbeeken: Numbers that exceed the precision supported by javascript

docs/upgrade-to-v5.md
```diff
@@ -100,3 +100,7 @@ The following deprecated methods have been removed:
 - `ObjectId.prototype.get_inc`
 - `ObjectId.get_inc`
   - The `static getInc()` is private since invoking it increments the next `ObjectId` index, so invoking would impact the creation of subsequent ObjectIds.
+
+### BSON Element names are now fetched only from object's own properties
+
+Previously objects passed to the `BSON.serialize`, `BSON.calculateObjectSize`, and `EJSON.stringify` API would have the element names enumerated with a `for-in` loop which will emit keys defined on the prototype. Since this is likely surprising, especially if a globally shared prototype has been modified we are now using `Object.keys` to enumerate the element names from a js object.
```
baileympearson, line 106:
```suggestion
`BSON.serialize`, `EJSON.stringify` and `BSON.calculateObjectSize` now only consider own properties and do not consider properties defined on the prototype of the object when serializing objects.

Example:
```typescript
const object = { a: 1 };
Object.setPrototypeOf(object, { b: 2 });
deserialize(serialize(object)); // { a: 1 } in 5.0, { a: 1, b: 2 } in 4x
```

baileympearson (changes requested):
one small upgrade guide suggestion

baileympearson (dismissed):

baileympearson (approved):

PR #528, opened 2022-11-23
feat!(NODE-4710): remove capital D ObjectID export

nbbeeken:


test/node/extended_json_tests.js
```diff
@@ -381,6 +363,18 @@ describe('Extended JSON', function () {
...
+      // Old deserializer needs `copy` to exist
+      newBsonObject.decimal128.bytes.copy = function (buffer, offset, start, end) {
+        return buffer.set(this.subarray(start, end), offset)
+      }
+      newBsonObject.objectId.id.copy = function (buffer, offset, start, end) {
+        return buffer.set(this.subarray(start, end), offset)
+      }
+      newBsonObject.objectID.id.copy = function (buffer, offset, start, end) {
+        return buffer.set(this.subarray(start, end), offset)
+      }
```
addaleax, line 376: Curious, why wasn’t this necessary before? And doesn’t the fact that it requires monkey-patching basically mean that proper interop is gone anyway?

nbbeeken: Yea that was a question I was going to bring up, IIRC we were concerned about the interop b/c of a mix of bson versions in compass? with the removal of buffer the serialize functions in bson 1.x can't work with the types from 4.x when run in a non-nodejs env. 

This wasn't failing before because these tests would attempt to import legacy bson but fallback to the current one so they ran but didn't compare anything unless you manually installed legacy bson, I set it up to always be available, uncovering these issues.

addaleax: Ah right, I was mixing this up with the interop tests in `test/node/bson_older_versions_tests.js` – those still pass, so maybe this isn't really all that problematic? It probably makes sense to adjust the `usesBufferPolyfill` logic there, in any case.

> concerned about the interop b/c of a mix of bson versions in compass

It's not just Compass, it's about being able to use a 3.x or lower driver with bson 4.x or above. That's the situation Compass was in for a long time, but I assume it's not that uncommon for users in general to not use both latest driver and latest bson? Right now, this isn't an issue for us, since we're on the 4.x driver with 4.x bson.

nbbeeken: Well because the each driver depends directly on its BSON library version it should be tied together but understandably folks can add BSON as a separate dep and use types from there, which would break against an older driver. That's why the driver reexports BSON types but it doesn't reexport BSON APIs (serialize etc.) and maybe that's the short coming that would inspire folks to install bson themselves 🤔 
I don't see how its possible that we make changes to 4.x that we don't also backport to 1.x to keep cross compat going, maybe that's something we have to consider, this is something we need clear messaging on regardless so I'll put this on our agenda for tech discussion. 

Yea your set of tests that put all the bundles against each other etc still pass but only on nodejs and that's b/c we're still using Buffer when in that in env. And maybe it's okay for this incompatibility to exist in the "web" mode of the library. (Or maybe its worth patching 1.x to use `.set` 👀)

addaleax: > backport to 1.x to keep cross compat going, maybe that's something we have to consider

My gut feeling is that you don’t have to worry about cross compat support with 1.x and 5.x. You _may_ have to worry about cross compat support with 4.x with 5.x, though?

> And maybe it's okay for this incompatibility to exist in the "web" mode of the library.

I would say so, yeah. The driver is strictly a Node.js library, so the concern of using a driver that uses a different bson version than the one that’s being used directly doesn’t really apply to the web mode variant.

nbbeeken: I created https://jira.mongodb.org/browse/NODE-4843 we can add v4 as a new version to verify against in the matrix that "bson_older_versions_tests" checks.

src/extended_json.ts
```diff
@@ -273,7 +273,10 @@ const BSON_TYPE_MAPPINGS = {
   MaxKey: () => new MaxKey(),
   MinKey: () => new MinKey(),
   ObjectID: (o: ObjectId) => new ObjectId(o),
-  ObjectId: (o: ObjectId) => new ObjectId(o), // support 4.0.0/4.0.1 before _bsontype was reverted back to ObjectID
+  // The _bsontype for ObjectId is spelled with a capital "D", to the mapping above will be used (most of the time)
```
durran, line 276: Why can we not just eradicate the capital "D" completely in the _bsontype? It's a runtime value - is this because we think that users may be storing the properties of the ObjectId and we need to be backwards compatible? (I'm thinking Compass preferences as an example).

PR #529, opened 2022-11-28
feat!(NODE-4850): serialize negative zero to double

nbbeeken:
#### What is changing?

When negative zero is detected it will be serialized to an 8 byte double instead of an Int32 to preserve the sign information.

Remove lossy serialization

- [ ] Migration guide updated


nbbeeken:
NODE-4850

test/node/parser/serializer.test.ts
```diff
@@ -0,0 +1,80 @@
...
+        input: 2.3,
+        output: bufferFromHexArray([doubleType, keyA, '6666666666660240'])
+      },
+      {
+        title: `fractional looking number (100000e-5) but is not fractional is serialized to int32`,
+        input: 100000e-5,
+        output: bufferFromHexArray([int32Type, keyA, '01000000'])
+      },
+      {
+        title: `fractional arithmetic (1.2 + 0.8) that sums to and int is serialized to int32`,
```
baileympearson, line 66:
```suggestion
        title: `fractional arithmetic (1.2 + 0.8) that sums to an int is serialized to int32`,
```

test/node/parser/serializer.test.ts
```diff
@@ -0,0 +1,80 @@
+import * as BSON from '../../register-bson';
+import { expect } from 'chai';
+import { bufferFromHexArray } from '../tools/utils';
+import { ByteUtils } from '../../../src/utils/byte_utils';
+
+describe('serialize()', () => {
+  describe('javascript number', () => {
```
baileympearson, line 7: Can you clarify why these tests are necessary?

Here are my thoughts.  You did refactor how we serialize numbers, so we need to make sure that the refactor is adequately tested.  But either our existing test coverage is adequate and no reason for new tests here (except for new negative zero tests) or our existing test coverage is inadequate.  If our existing test coverage is inadequate, then we should either revert the refactor in our number parsing so we're not making untested refactors, or add more testing here so our coverage is complete and leave the refactor.

docs/upgrade-to-v5.md
```diff
@@ -100,3 +100,7 @@ The following deprecated methods have been removed:
 - `ObjectId.prototype.get_inc`
 - `ObjectId.get_inc`
   - The `static getInc()` is private since invoking it increments the next `ObjectId` index, so invoking would impact the creation of subsequent ObjectIds.
+
+### Negative Zero is now serialized to Double
+
+While use cases for negative zero may be rare it is important that the value is preserved on round trips through the BSON serialization layer. Prior to this change negative zero was possible to preserve with the BSON `new Double(-0)` type, such code can be updated to use `-0` directly.
```
baileympearson, line 106:
```suggestion
4x versions of BSON incorrectly serialized the literal -0 as an Int32, which does not support -0.  This results in the following behavior: 

```typescript
deserialize(serialize({ number: -0 })); // { number: 0 }
deserialize(serialize({ number: new Double(-0) })) // { number: Double(-0) }
```

BSON now automatically serializes -0 to a Double.
```typescript
deserialize(serialize({ number: -0 })); // { number: Double(-0) }
```

Any code previously serializing -0 as a double can now use the literal `-0` syntax safely.
```

**Review by baileympearson — CHANGES_REQUESTED** [review]

File: test/node/parser/serializer.test.ts (hunk 2 above)
Comment (nbbeeken) [reply test/node/parser/serializer.test.ts]: they aren't necessary

File: docs/upgrade-to-v5.md (hunk 3 above)
Comment (nbbeeken) [reply docs/upgrade-to-v5.md]: updated, I kept it short and simple

File: docs/upgrade-to-v5.md [hunk 4]
```diff
@@ -100,3 +100,17 @@ The following deprecated methods have been removed:
...
+
+Previously it was required to use the `Double` type class to preserve `-0`:
+```ts
+BSON.serialize({ d: new Double(-0) })
+```
+
+Now `-0` can be used directly
+```ts
+BSON.serialize({ d: -0 })
+```
```
baileympearson, line 116: Maybe we should highlight the difference in behavior here?
```suggestion
Previously it was required to use the `Double` type class to preserve `-0`:
```ts
BSON.deserialize(BSON.serialize({ d: new Double(-0) })); // { d: Double(-0) }
BSON.deserialize(BSON.serialize({ d: new -0})); // { d: 0 }
```

Now `-0` can be used directly
```ts
BSON.deserialize(BSON.serialize({ d: new -0})); // { d: Double(-0) }
```
```

nbbeeken: The syntax here is a little broken, did you intend `new -0`? and deserialize does not return a Double instance unless you disable `promoteValues`. There were no changes to deserialize because it will return a negative zero js number if one is encoded in the BSON bytes so there's nothing to showcase with deserialize here

nbbeeken: updated

baileympearson (approved):

PR #530, opened 2022-11-28
ci(NODE-4204): add support for latest node version in CI

baileympearson:
#### What is changing?

Adds latest node into CI.

Nope.

### DIFF
--- a/.evergreen/setup-environment.sh
+++ b/.evergreen/setup-environment.sh
@@ -8,14 +8,20 @@ else
 fi
 export PROJECT_DIRECTORY="$(pwd)"
 
-if [ "$NODE_MAJOR_VERSION" == 'latest' ]; then 
-    NODE_VERSION='node'
-    echo "using latest node version"
-else 
+get_node_version() {
+    local NODE_DOWNLOAD_URI="nodejs.org/download/release/latest-v${NODE_MAJOR_VERSION}.x/SHASUMS256.txt"
+
+    if [ "$NODE_MAJOR_VERSION" == 'latest' ]; then 
+        NODE_DOWNLOAD_URI="nodejs.org/download/release/latest/SHASUMS256.txt"
+    fi
+
     # get the latest version of node for given major version
-    NODE_VERSION=$(curl -sL nodejs.org/download/release/latest-v${NODE_MAJOR_VERSION}.x/SHASUMS256.txt -o - | head -n 1 | tr -s ' ' | cut -d' ' -f2 | cut -d- -f2 | cut -dv -f2)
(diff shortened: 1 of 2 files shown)


.evergreen/setup-environment.sh
```diff
@@ -0,0 +1,31 @@
+#! /bin/bash
```
nbbeeken, line 1:
```suggestion
#! /usr/bin/env bash
```
just for [cross compat reasons](https://stackoverflow.com/questions/16365130/what-is-the-difference-between-usr-bin-env-bash-and-usr-bin-bash)

.evergreen/setup-environment.sh
```diff
@@ -0,0 +1,31 @@
...
+PROJECT_DIRECTORY: "$PROJECT_DIRECTORY"
+NODE_VERSION: "$NODE_VERSION"
+PREPARE_SHELL: |
+    set -o errexit
+    set -o xtrace
+    export PROJECT_DIRECTORY="$PROJECT_DIRECTORY"
+    export NODE_VERSION="$NODE_VERSION"
+EOT
+# See what we've done
+cat expansion.yml
```
nbbeeken, line 31: Super minor, can we put the newline at the end here?

.evergreen/setup-environment.sh
```diff
@@ -0,0 +1,31 @@
...
+export PROJECT_DIRECTORY="$(pwd)"
+
+if [ "$NODE_MAJOR_VERSION" == 'latest' ]; then 
+    NODE_VERSION='node'
+    echo "using latest node version"
+else 
+    # get the latest version of node for given major version
+    NODE_VERSION=$(curl -sL nodejs.org/download/release/latest-v${NODE_MAJOR_VERSION}.x/SHASUMS256.txt -o - | head -n 1 | tr -s ' ' | cut -d' ' -f2 | cut -d- -f2 | cut -dv -f2)
+    echo "LATEST NODE ${NODE_MAJOR_VERSION}.x = $NODE_VERSION"
+fi
```
nbbeeken, line 18: I don't think we need this if stmt, the curl URL can be set to `nodejs.org/download/release/latest/SHASUMS256.txt` then the nvm commands will be using an explicit version number, only for perhaps better logs and just no longer relying on defaults

nbbeeken (changes requested):

.evergreen/setup-environment.sh
baileympearson: per discussion from slack - aligned the download behavior with the driver

nbbeeken (approved):

PR #531, opened 2022-11-28
fix(NODE-4464): stringify negative zero to $numberDouble: -0.0

nbbeeken:
#### What is changing?

EJSON when in `relaxed: false` mode (cannonical) is supposed to serialize negative zero to a double.

EJSON implementation correctness


jimmywarting:
( issues/discussion where closes... so wiring it here instead... )

I created this test page to test out different encoders / decoders https://jimmywarting.github.io/3th-party-structured-clone-wpt/ and wanted to share this with you...

you maybe want to try to increase the score a little better...?

nbbeeken:
Hi @jimmywarting, if you have not already please feel free to [make a JIRA account](https://account.mongodb.com/account/register) to bring up any specific issues you think would be valuable to consider. I understand it is yet another account to make (but I think you can OAuth with github, convenient!). Over there we can better get more specific about the goals and track their progress. Also, we can make sure downstream projects are aware of the changes. JIRA is how we stay organized so we can best deliver fixes and feats. Regardless, thank you for sharing!

jimmywarting:
Don't want to create a JIRA account, did never like it (used it for work)... much more prefer github (and i think that you should move your issues to Github instead where you also have your code, works better with permalinks and refering issues across different github repos etc)

src/extended_json.ts
```diff
@@ -216,16 +217,16 @@ function serializeValue(value: any, options: EJSONSerializeOptions): any {
   }
 
   if (typeof value === 'number' && (!options.relaxed || !isFinite(value))) {
-    // it's an integer
-    if (Math.floor(value) === value) {
-      const int32Range = value >= BSON_INT32_MIN && value <= BSON_INT32_MAX,
-        int64Range = value >= BSON_INT64_MIN && value <= BSON_INT64_MAX;
+    if (Number.isInteger(value) && !Object.is(value, -0)) {
+      const int32Range = value >= BSON_INT32_MIN && value <= BSON_INT32_MAX;
+      const int64Range = value >= BSON_INT64_MIN && value <= BSON_INT64_MAX;
```
baileympearson, line 222: optional
```suggestion
      const isInInt32Range = value >= BSON_INT32_MIN && value <= BSON_INT32_MAX;
      const isInInt64Range = value >= BSON_INT64_MIN && value <= BSON_INT64_MAX;
```

baileympearson: (and corresponding usages ofc)

src/double.ts
```diff
@@ -43,7 +43,7 @@ export class Double {
   }
 
   toString(radix?: number): string {
-    return this.value.toString(radix);
+    return radix == null && Object.is(this.value, -0) ? '-0' : this.value.toString(radix);
```
baileympearson, line 46: why is the radix guard necessary here?  I'd expect that in any radix, negative zero is returned as `-0`.

src/extended_json.ts
```diff
@@ -75,8 +75,9 @@ function deserializeValue(value: any, options: EJSON.Options = {}) {
 
     // if it's an integer, should interpret as smallest BSON integer
     // that can represent it exactly. (if out of range, interpret as double.)
-    if (Math.floor(value) === value) {
+    if (Number.isInteger(value) && !Object.is(value, -0)) {
       if (value >= BSON_INT32_MIN && value <= BSON_INT32_MAX) return new Int32(value);
+      // TODO(NODE-4377): EJSON js number handling diverges from BSON
```
baileympearson, line 80: Do we have a test for this behavior somewhere?  If not, should we add a test demonstrating now that `-0` will deserialize to a double instead of an integer?

baileympearson (changes requested):

src/double.ts
nbbeeken: We shouldn't make this change (at least not here) I made it while debugging something sorry it slipped in

src/extended_json.ts
nbbeeken: made the two blocks that do this logic look identical now

src/extended_json.ts
nbbeeken: added!

baileympearson:
Can we change the title of the PR to accurately reflect the contents?  The PR now handles parsing ejson with negative zeroes as well.

baileympearson (approved):

PR #524, opened 2022-11-22
feat!(NODE-4701): remove default export and add named namespace export

nbbeeken:
#### What is changing?

Removes deprecated default export, adds a namespace export.

```js
// typescript and es module users:
import * as BSON from 'bson'; // either
import { BSON, ObjectId } from 'bson'; // or

// nodejs commonjs users:
const BSON = require('bson');
```

The following is no longer supported:
```js
import BSON from 'bson';
```

#### What is the motivation for this change?

The default had to be maintained explicitly, did not have access to the type definitions nor constants so it wasn't an equivalent import. default export has a lot of behavioral corner cases that we want to avoid.

### Double check the following

- [x] Ran `npm run lint` script
- [x] Self-review completed using the [steps outlined here](https://github.com/mongodb/node-mongodb-native/blob/HEAD/CONTRIBUTING.md#reviewer-guidelines)
- [x] PR title follows the correct format: `<type>(NODE-xxxx)<!>: <description>`
- [x] Changes are covered by tests
- [x] New TODOs have a related JIRA ticket

PR #532, opened 2022-12-02
feat!(NODE-4862): add BSONType enum and remove internal constants from export

nbbeeken:
#### What is changing?

- removes internal constants from export, 
- Adds the BSONType enum from the driver

The exports are a blocker for #524 passing typescript tests

The enum is a nice to have, it provides the same information that constants did but in a consistent way the driver already does, locating the enum here puts it closer to the code that it relates to, its not used by the BSON library but if someone needs type indicators this is a consistent way to make it available. 


test/node/constants.test.ts
```diff
@@ -0,0 +1,180 @@
...
+    });
+    it('Min key should be 0xFF', () => {
+      expect(constants.BSON_DATA_MIN_KEY).to.equal(0xff);
+    });
+    it('Max key should be 0x7F', () => {
+      expect(constants.BSON_DATA_MAX_KEY).to.equal(0x7f);
+    });
+  });
+
+  describe('BSONType enum', () => {
```
nbbeeken, line 143: Tests above are old, no change except for the import
Tests here are new for the enum

src/constants.ts
```diff
@@ -108,3 +108,31 @@ export const BSON_BINARY_SUBTYPE_COLUMN = 7;
...
+  int: 16,
+  timestamp: 17,
+  long: 18,
+  decimal: 19,
+  minKey: -1,
+  maxKey: 127
+} as const);
+
+/** @public */
+export type BSONType = typeof BSONType[keyof typeof BSONType];
```
nbbeeken, line 138: Using the enum pattern here, it makes BSONType the annotation equal to a union of the values.
BSONType the runtime value is an object with properties.

durran (commented):
Are we planning on removing the export in https://github.com/mongodb/node-mongodb-native/blob/main/src/mongo_types.ts to just re-export `BSONType` from here as part of NODE-4701?

durran (approved):

nbbeeken:
I expect we'll change to a re-export in this ticket: https://jira.mongodb.org/browse/NODE-4867

durran (approved):

PR #533, opened 2022-12-02
refactor(NODE-4859): move random bytes implementations to platform specific byte utils

nbbeeken:
#### What is changing?

Moved the randomBytes logic to web and node byte utils

The ticket we have to make sense of the bundling ran into an issue with how require works. TODO will elaborat.


src/utils/node_byte_utils.ts
```diff
@@ -23,6 +23,25 @@ type NodeJsBufferConstructor = Omit<Uint8ArrayConstructor, 'from'> & {
...
+export function mathRandomBytes(byteLength: number) {
+  return nodeJsByteUtils.fromNumberArray(
+    Array.from({ length: byteLength }, () => Math.floor(Math.random() * 256))
+  );
+}
+
+/** @internal */
+const localRandomBytes = (() => {
+  try {
+    // What about nodejs es module users.........
```
nbbeeken, line 36: Have to discuss this, may change the goals of NODE-4860

src/utils/node_byte_utils.ts
```diff
@@ -23,6 +23,25 @@ type NodeJsBufferConstructor = Omit<Uint8ArrayConstructor, 'from'> & {
 // Node.js global
 declare const Buffer: NodeJsBufferConstructor;
 
+/** @internal */
+export function mathRandomBytes(byteLength: number) {
```
baileympearson, line 27: Why duplicate this function?  We could easily put this in the shared utils file:
```typescript
export function mathRandomBytes(byteUtils: ByteUtils, byteLength: number) {
  return byteUtils.fromNumberArray(
    Array.from({ length: byteLength }, () => Math.floor(Math.random() * 256))
  );
}
```

test/node/byte_utils.test.ts
```diff
@@ -424,6 +425,26 @@ const utf8ByteLengthTests: ByteUtilTest<'utf8ByteLength'>[] = [
...
+  },
+  {
+    name: 'when byteLength is 10 returns ten byteLength buffer',
+    inputs: [10],
+    expectation({ output, error }) {
+      expect(error, error?.message).to.be.null;
+      expect(output).to.be.instanceOf(Uint8Array);
+      expect(output).to.have.property('byteLength', 10);
+    }
+  }
```
baileympearson, line 446: Why 10?  Is it just generally a non-zero positive integer?  If so, maybe the test should be more generic:

```typescript
    name: 'when byteLength is a non-zero positive integer returns ten byteLength buffer',
```

Also, do we care about cases where number could be negative or greater than the maximum array length?  I assume it would throw in these scenarios

test/node/byte_utils.test.ts
```diff
@@ -544,6 +566,32 @@ describe('ByteUtils', () => {
...
+      bsonWithNoCrypto = loadBSONWithGlobal({
+        crypto: null,
+        // if we don't add a copy of Math here then we cannot spy on it for the test
+        Math: {
+          pow: Math.pow,
+          floor: Math.floor,
+          random: Math.random
+        }
+      });
+    });
```
baileympearson, line 581: Isn't there different behavior here for both web and nodejs?  NodeJS falls back, web only falls back if it is not react native.  Should we add platform specific tests for these behaviors somewhere (if we don't have them)?

test/node/byte_utils.test.ts
```diff
@@ -544,6 +566,32 @@ describe('ByteUtils', () => {
...
+
+    after(function () {
+      sinon.restore();
+      bsonWithNoCrypto = null;
+    });
+
+    it('should fall back to Math.random implementation for random bytes if crypto is not present', () => {
+      const randomSpy = sinon.spy(bsonWithNoCrypto.Math, 'random');
+      new bsonWithNoCrypto.BSON.UUID();
+      expect(randomSpy).to.have.callCount(16);
```
baileympearson, line 591: Why 16?  That's not entirely clear.  Is it not enough to simply assert that it was called, and that `crypto.randomBytes` was not called?

src/utils/global.ts
```diff
@@ -1,22 +0,0 @@
...
-declare const self: unknown;
-declare const global: unknown;
-
-function checkForMath(potentialGlobal: any) {
-  // eslint-disable-next-line eqeqeq
-  return potentialGlobal && potentialGlobal.Math == Math && potentialGlobal;
-}
-
-// https://github.com/zloirock/core-js/issues/86#issuecomment-115759028
-export function getGlobal<T = Record<string, unknown>>(): T {
```
baileympearson, line 13: I assume this code was unused and is safe to delete?

baileympearson (changes requested):

src/utils/node_byte_utils.ts
nbbeeken: This is going to be a problem for the way we planned on bundling the library in NODE-4860, so it does not block this PR. I'll bring it up to the team when that ticket / PR is ready, I'll clean up my comment here for now, it just occurred to me when I wrote that line, didn't want to forget.

src/utils/node_byte_utils.ts
nbbeeken: This was to avoid a circular dependency on ByteUtils

test/node/byte_utils.test.ts
nbbeeken: Because 16 is the the length of a UUID. I would change it but `crypto` is null so it does not have a randomBytes method to not be called, so we cannot exactly assert that, any ideas?

src/utils/global.ts
nbbeeken: Yep

test/node/byte_utils.test.ts
nbbeeken: updated

test/node/byte_utils.test.ts
nbbeeken: Added tests for nodejs and react native

test/node/byte_utils.test.ts
```diff
@@ -424,6 +431,48 @@ const utf8ByteLengthTests: ByteUtilTest<'utf8ByteLength'>[] = [
...
+    }
+  },
+  {
+    name: 'when byteLength is beyond the supported length for arrays',
+    inputs: [4294967296],
+    expectation({ output, error }) {
+      if (isNode19OrHigher) {
+        // Node 19 on linux (not macos) throws a QuotaExceededError: The requested length exceeds 65,536 bytes
+        // At least we can assert an Error is thrown
+        expect(error).to.exist;
```
nbbeeken, line 468: https://evergreen.mongodb.com/task_log_raw/node_bson_linux_node_tests_latest_patch_a469e919bbee8704d0bbc18d435d7a4d27f65027_638e4637d1fe073c01975b7a_22_12_05_19_27_52/0?type=T#L2723

test/node/byte_utils.test.ts
```diff
@@ -544,6 +594,102 @@ describe('ByteUtils', () => {
...
+      after(function () {
+        sinon.restore();
+        bsonWithNoCrypto = null;
+      });
+
+      it('should fall back to Math.random implementation for random bytes if crypto is not present', () => {
+        const randomSpy = sinon.spy(bsonWithNoCrypto.Math, 'random');
+        new bsonWithNoCrypto.BSON.UUID();
+        // 16 is the length of a UUID
+        expect(randomSpy).to.have.callCount(16);
```
baileympearson, line 621:
```suggestion
        expect(randomSpy.called).to.be.true;
```
optional - the current assertion doesn't feel quite right.  What we care about is that `randomSpy` is called - we don't care about how many times it is called.  If it were me, I'd only assert that it was called

baileympearson: this would apply to the other tests in this file, if you choose to take the suggestion

baileympearson (dismissed):

test/node/byte_utils.test.ts
nbbeeken: A UUID will never change it's length and I'd want to know if Math.random was called more times than necessary, it does expand the scope of the test but it could catch something if the function is changed down the road to accidentally be invoked 1000s of times just to be used for 16 bytes of random values

nbbeeken: Ah yes fair point, let me update the title

nbbeeken: updated!

baileympearson (approved):

PR #534, opened 2022-12-05
feat(NODE-4713)!: modernize bundling

nbbeeken:
#### What is changing?

Consolidating bundles and adding a rollup plugin to support `mjs` in the browser and node


test/register-bson.js
```diff
@@ -40,7 +40,7 @@ console.error(inspect({ web }, { colors: true }));
 
 let BSON;
 if (web) {
-  BSON = loadBSONWithGlobal().BSON;
+  BSON = loadBSONWithGlobal().exports;
```
nbbeeken, line 43: now loading cjs, plus changes to the loadBSON helpers

test/register-bson.d.ts
```diff
@@ -1 +1,4 @@
-export * from '../src/index'
+// Register BSON should export precisely what the library exports
+// plus the web mode indicator
+export * from '..'
+export const __isWeb__: boolean;
```
nbbeeken, line 4: type checks tests in VSCode

src/extended_json.ts
```diff
@@ -347,116 +345,110 @@ function serializeDocument(doc: any, options: EJSONSerializeOptions) {
...
+ * console.log(EJSON.parse(text, { relaxed: false }));
+ *
+ * // prints { int32: 10 }
+ * console.log(EJSON.parse(text));
+ * ```
  */
-// the namespace here is used to emulate `export * as EJSON from '...'`
-// which as of now (sept 2020) api-extractor does not support
-// eslint-disable-next-line @typescript-eslint/no-namespace
-export namespace EJSON {
```
nbbeeken, line 356: Sorry about this change looking large, removing the namespace in favor of the explicit export is favorable b/c the code generated in the bundling process is under our control rather than the tooling's

src/bson.ts
```diff
@@ -31,8 +30,9 @@ export type { ObjectIdExtended, ObjectIdLike } from './objectid';
 export type { BSONRegExpExtended, BSONRegExpExtendedLegacy } from './regexp';
 export type { BSONSymbolExtended } from './symbol';
 export type { LongWithoutOverrides, TimestampExtended, TimestampOverrides } from './timestamp';
-export { LongWithoutOverridesClass } from './timestamp';
+export type { LongWithoutOverridesClass } from './timestamp';
```
nbbeeken, line 33: This is typescript hackery to support Timestamp subclassing Long, this isn't a real value that should be exported, removing this change came about as part of cleaning up the bundle output

.mocharc.json
```diff
@@ -12,5 +12,8 @@
   "timeout": 10000,
   "failZero": true,
   "sort": true,
-  "color": true
+  "color": true,
+  "node-option": [
+    "experimental-vm-modules"
```
nbbeeken, line 17: Makes sinon tests for mjs possible (need to flag on node version)

.evergreen/install-dependencies.sh
```diff
@@ -43,21 +43,8 @@ nvm use --lts
 
 set -o xtrace
 
-
-
-# setup npm cache in a local directory
-cat <<EOT > .npmrc
```
nbbeeken, line 49: these settings are unnecessary and just caused warnings to print out in CI

package.json
```diff
@@ -9,10 +9,8 @@
   "files": [
     "lib",
     "src",
-    "dist",
```
nbbeeken, line 12: The bundles now output to the lib folder

test/load_bson.js
```diff
@@ -5,33 +5,56 @@ const fs = require('node:fs');
...
     // Putting this last to allow caller to override default globals
     ...globals
   });
+
   vm.runInContext(code, context, { filename });
-  return context;
+  return { context, exports: context.exports };
+}
+
+async function load_ESM_BSON_WithGlobal(globals) {
```
durran, line 42: Is there a reason we are breaking JS naming conventions here? I would assume we'd use camel case only here.

durran (changes requested):
First round of high level notes.

test/load_bson.js
nbbeeken: No reason I can fix this, just a lot of capital letters

src/utils/node_byte_utils.ts
```diff
@@ -30,17 +31,15 @@ export function nodejsMathRandomBytes(byteLength: number) {
   );
 }
 
-/** @internal */
+/** @internal REQUIRE WILL BE REWRITTEN: Modify this code with caution */
```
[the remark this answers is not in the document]
nbbeeken: Expanded on the reasoning in a comment that mentions the plugin, the plugin fails if something changes (hopefully) so I think we can't change this by mistake, but let me know if we can make it any more clear

src/extended_json.ts
durran: Just comfirming there are no changes here just a diff because of the ns removal.

durran (approved):

src/extended_json.ts
KuSh: There's at least the removal of SerializableTypes type which is not listed on the release note (I've just come into it during migration)

nbbeeken: Hey @KuSh, thanks for calling this out, and sorry about the surprise removal. I have filed a ticket for us to update the migration guide. At this time the workaround would be declaring the type in your code and casting, do you have a use case for that type that you could share (does it help with narrowing or something else)? I will capture what it provides so if we reintroduce it or something similar it remains useful. Thanks in advance!

https://jira.mongodb.org/browse/NODE-6302

KuSh: We were using those types for some function parameters and return type that were using parse and stringify internally. I have just changed them to unknown for now (as our policy forbids any) but this is clearly less strict than before and I'm not sure those fonctions accept more form of content so potentially more runtime errors than before

PR #535, opened 2022-12-05
feat!(NODE-4703): remove strict flag from parse and stringify

nbbeeken:
#### What is changing?

Removing deprecated strict flag from EJSON

Removing a deprecation


src/extended_json.ts
```diff
@@ -377,19 +372,15 @@ export namespace EJSON {
    * ```
    */
   export function parse(text: string, options?: EJSON.Options): SerializableTypes {
-    const finalOptions = Object.assign({}, { relaxed: true, legacy: false }, options);
-
-    // relaxed implies not strict
-    if (typeof finalOptions.relaxed === 'boolean') finalOptions.strict = !finalOptions.relaxed;
-    if (typeof finalOptions.strict === 'boolean') finalOptions.relaxed = !finalOptions.strict;
+    options = { relaxed: true, legacy: false, ...options };
```
durran, line 375: I'd prefer we don't mutate the input options here and assign to a new const.

durran (changes requested):

durran (dismissed):

docs/upgrade-to-v5.md
```diff
@@ -130,3 +130,7 @@ Now `-0` can be used directly
 BSON.deserialize(BSON.serialize({ d: -0 }))
 // type preservation, returns { d: -0 }
 ```
+
+### Extended JSON `strict` flag removed
+
+Extended JSON `parse` and `stringify` APIs ignore the deprecated `strict` option and it has been removed from the type definitions.
```
baileympearson, line 136: This is not clear.  Did they used to ignore the option or do they now?  We should make the difference between the old and new behaviors explicit.

Maybe something like:
```
In BSON 4x, the `strict` option is deprecated in the extended JSON `parse` and `stringify` functions.  

In BSON 5x, this option is removed.
```

durran (approved):

PR #537, opened 2022-12-07
feat(NODE-1921)!: validate serializer root input

nbbeeken:
#### What is changing?

- Added validation to the root input for serialize
- Adjusted the internal fn prototypes to no longer have optional arguments (avoids the bug we had with serializeFunctions)
- Fixed lack of cyclic reference check for Code w/ scope and DBRef fields ~(TODO: need to add a test)~
- Added cyclic ref checks that test a wider array of object arrangements
- Moved the EJSON circular tests to the same file, organizational change only
- Fixed breakage introduced in #527 where nullish input caused an error

Arrays cannot be root BSON documents, see migration guide.

### DIFF
--- a/src/parser/serializer.ts
+++ b/src/parser/serializer.ts
@@ -20,8 +20,7 @@ import {
   isDate,
   isMap,
   isRegExp,
-  isUint8Array,
-  normalizedFunctionString
+  isUint8Array
 } from './utils';
 
 /** @public */
@@ -388,14 +387,7 @@ function serializeDouble(buffer: Uint8Array, key: string, value: Double, index: 
   return index;
 }
 
-function serializeFunction(
-  buffer: Uint8Array,
-  key: string,
-  value: Function,
-  index: number,
-  _checkKeys = false,
-  _depth = 0
-) {
+function serializeFunction(buffer: Uint8Array, key: string, value: Function, index: number) {
   buffer[index++] = constants.BSON_DATA_CODE;
   // Number of written bytes
   const numberOfWrittenBytes = ByteUtils.encodeUTF8Into(buffer, key, index);
@@ -403,7 +395,7 @@ function serializeFunction(
   index = index + numberOfWrittenBytes;
   buffer[index++] = 0;
   // Function string
-  const functionString = normalizedFunctionString(value);
+  const functionString = value.toString();
 
   // Write the string
   const size = ByteUtils.encodeUTF8Into(buffer, functionString, index + 4) + 1;
@@ -444,7 +436,7 @@ function serializeCode(
 
     // Serialize the function
     // Get the function string
-    const functionString = typeof value.code === 'string' ? value.code : value.code.toString();
+    const functionString = value.code;
     // Index adjustment
     index = index + 4;
     // Write string into buffer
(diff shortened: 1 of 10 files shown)


src/parser/serializer.ts
```diff
@@ -608,18 +620,39 @@ function serializeDBRef(
...
+      buffer[3] = 0x00;
+      // All documents end with null terminator
+      buffer[4] = 0x00;
+      return 5;
+    }
+
+    if (Array.isArray(object)) {
+      throw new BSONError('serialize does not support an array as the root input');
+    }
+    if (typeof object !== 'object') {
```
durran, line 647: @nbbeeken I was curious since this line didn't seem it would catch all expectations in my opinion and added some tests to pass a `Long`, `Int32`, etc and it does indeed serialize them as our custom BSON types will pass this check. So it felt odd to me that `BSON.serialize(1)` would error but `BSON.serialize(new Int32('1'))` would not. This ticket explicitly states to error on "non object inputs" but I just felt the above example unexpected.

Since this is the last ticket in the v5 epic I thought (against or normal process in this special case) that maybe we decide what to do about that here? It just feels incomplete that we might still roundtrip some single values as a result of this change, which might be confusing to users. Thoughts?

durran: If I add one more check after this lin (probably a more elegant way) then I do get my expectation of all single values not getting round tripped:

```
    if ('_bsontype' in object) {
      throw new BSONError('serialize does not support non-object as the root input');
    }
```

nbbeeken: Added a check for `_bsontype` and that it is a string, this aligns with our typescript definition in the driver for not recursing a type: [here](https://github.com/mongodb/node-mongodb-native/blob/main/src/mongo_types.ts#L520)

Which in looking at that type, should we also be looking to prevent those other object types that we consider bson value from being passed as the root? Currently both `Date` and `RegExp` result in an empty document, Buffer will turn into (3 bytes in size) `{ '0': 0, '1': 0, '2': 0, parent: {}, offset: 0 }`. Thoughts?

test/node/parser/serializer.test.ts
```diff
@@ -14,4 +14,57 @@ describe('serialize()', () => {
...
+    const emptyDocumentNull = BSON.serialize({ a: null });
+    expect(emptyDocumentNull).to.deep.equal(nestedNull);
+  });
+
+  describe('validates input types', () => {
+    it('does not permit arrays as the root input', () => {
+      expect(() => BSON.serialize([])).to.throw(/does not support an array/);
+    });
+
+    it('does not permit objects with a _bsontype string to be serialized at the root', () => {
```
durran, line 44: Can we add one more test for non-string `_bsontype`?

durran (changes requested):

test/node/parser/serializer.test.ts
nbbeeken: Done, and added more prevention the common types we discussed

durran (approved):

src/parser/serializer.ts
```diff
@@ -410,7 +419,8 @@ function serializeCode(
   checkKeys = false,
```
baileympearson, line 419: perhaps someday we should consider a context object for bson serialization too.  9 position arguments, some with defaults, feels excessive.  just a thought

baileympearson (approved):
I looked at the migration guide and how we're validating inputs into serialize.  LGTM

PR #539, opened 2022-12-07
feat(NODE-4711)!: remove evalFunctions option

nbbeeken:
#### What is changing?

- The `evalFunctions` option and friends are removed
- deserialize always returns a Code instance
- Migration guide is updated
- Just for clarity: there are no changes to the `serializeFunctions` option

Evaluating javascript strings is not a common practice and since the BSON code type is generally used for legacy operations (pre aggregation pipeline) it generally not javascript that would be useful to be invoked client side. 

There is a migration example if this feature is still desired.


durran (changes requested):

src/code.ts
```diff
@@ -16,19 +16,27 @@ export class Code {
...
-    this.code = code;
-    this.scope = scope;
+  constructor(code: string | Function, scope?: Document | null) {
+    this.code = code.toString();
+    this.scope = scope ?? null;
   }
 
-  toJSON(): { code: string | Function; scope?: Document } {
-    return { code: this.code, scope: this.scope };
+  toJSON(): { code: string; scope?: Document } {
```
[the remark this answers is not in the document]
nbbeeken: Added testing, the code before was relying on the fact that undefined is omitted from JSON.stringify, so it just defined both keys and let the stringify call delete the scope property. I figured it made sense to make the toJSON and toExtendedJSON logic look exactly the same and be explict about omitting a scope property

durran (dismissed):

durran (approved):

docs/upgrade-to-v5.md
```diff
@@ -179,3 +179,33 @@ EJSON.parse("...",  { strict: false }); /* migrate to */ EJSON.parse("...",  { r
 // stringify
 EJSON.stringify({}, { strict: true  }); /* migrate to */ EJSON.stringify({}, { relaxed: false });
 EJSON.stringify({}, { strict: false }); /* migrate to */ EJSON.stringify({}, { relaxed: true });
+
+### `Code` only stores a string
```
addaleax, line 183: It doesn’t look like support for `Code` with `scope` has gone away here, so this might be a confusing wording?

nbbeeken: good call, fixed up the wording here

addaleax (approved):

PR #541, opened 2022-12-14
feat(NODE-4870): Support BigInt serialization

W-A-James:
#### What is changing?

Adding capability to serialize `BigInt`s

Yes. This documentation for this change will be captured in NODE-4871.


test/node/bigint_tests.ts
```diff
@@ -0,0 +1,123 @@
+import * as BSON from '../register-bson';
```
nbbeeken, line 1: Can you rename the file to `bigint.test.ts`, it's the name format that gets picked up by VSCode, and the pattern we use in the driver

test/node/bigint_tests.ts
```diff
@@ -0,0 +1,123 @@
...
+describe('BSON BigInt serialization Support', function () {
+  // Index for the data type byte of a BSON document with a single element
+  const DATA_TYPE_OFFSET = 4;
+  before(function () {
+    try {
+      BigInt(0);
+    } catch (_) {
+      this.skip('JS VM does not support BigInt');
+    }
+  });
```
nbbeeken, line 14: We no longer support runtimes that do not have `BigInt` so we can delete this

test/node/bigint_tests.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  it('Serializes bigints with the correct BSON type', function () {
+    const testDoc = { a: 0n };
+    const serializedDoc = BSON.serialize(testDoc);
+    expect(serializedDoc[DATA_TYPE_OFFSET]).to.equal(CONSTANTS.BSON_DATA_LONG);
+  });
+
+  it('Serializes bigints into little-endian byte order', function () {
+    const testDoc = { a: BigInt(0x1234567812345678n) };
```
nbbeeken, line 23:
```suggestion
    const testDoc = { a:0x1234567812345678n };
```

No need for the cast and `n` syntax, just redundant, I won't comment on every instance but there are more

test/node/bigint_tests.ts
```diff
@@ -0,0 +1,123 @@
...
+    expect(new Long(BigInt('0')).toString()).to.equal('0');
+    expect(new Long(BigInt('-1')).toString()).to.equal('-1');
+    expect(new Long(BigInt('-1'), true).toString()).to.equal('18446744073709551615');
+    expect(new Long(BigInt('123456789123456789')).toString()).to.equal('123456789123456789');
+    expect(new Long(BigInt('123456789123456789'), true).toString()).to.equal('123456789123456789');
+    expect(new Long(BigInt('13835058055282163712')).toString()).to.equal('-4611686018427387904');
+    expect(new Long(BigInt('13835058055282163712'), true).toString()).to.equal(
+      '13835058055282163712'
+    );
+  });
```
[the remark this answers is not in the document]
nbbeeken: Feels like these tests belong in long_tests.js (which you can rename to `long.test.ts`) so yea we can move em over

test/node/bigint_tests.ts
```diff
@@ -0,0 +1,123 @@
...
+    const expectedSerialization = byteUtils.bufferFromHexArray([
+      '12', //int64
+      Buffer.from('test\x00', 'utf8').toString('hex'),
+      '0100000000000000'
+    ]);
+    expect(serializedDoc).to.deep.equal(expectedSerialization);
+  });
+
+  it('Calls toBSON on a BigInt value if it exists', function () {
+    BigInt.prototype.toBSON = () => `hello`;
```
nbbeeken, line 98: If a error occurs within this test we'll have left a toBSON method on the BigInt prototype which could effect other tests in positive or negative ways. So to avoid that, adding/removing this method should be done in a beforeEach/afterEach hook. Not all bigint tests need to be in this file it may make more sense to test this with the other concerns of toBSON. But positioning of tests isn't critical as long as they're written in a safe way (don't impact other tests) and communicate intention.

nbbeeken (changes requested):
Quick first round

test/node/bigint_tests.ts
W-A-James: I see what you mean. Also, it seems that there is already a test for this in `test/node/to_bson_test.js:133-186`, so I can just go ahead and delete this test here altogether.

src/parser/serializer.ts
```diff
@@ -783,7 +797,7 @@ export function serializeInto(
       } else if (type === 'number') {
         index = serializeNumber(buffer, key, value, index);
       } else if (type === 'bigint' || isBigInt64Array(value) || isBigUInt64Array(value)) {
```
nbbeeken, line 799: The condition left behind here checks for more than it should, this should just be `type === 'bigint'`

test/node/bigint.test.ts
```diff
@@ -0,0 +1,89 @@
...
+    const serializedMinPositiveInt64 = BSON.serialize(minPositiveInt64);
+    const expectedSerializationForMinPositiveInt64 = bufferFromHexArray([
+      '12', // int64
+      '6100', // 'a' key with key null terminator
+      '0000000000000080'
+    ]);
+    expect(serializedMinPositiveInt64).to.deep.equal(expectedSerializationForMinPositiveInt64);
+  });
+
+  it('Correctly truncates a BigInt that is larger than a 64-bit int', function () {
```
nbbeeken, line 79: Just a nit about the test titles:

- you can always start a test title off with a lower case
- you can omit "correctly" from a name, and just focus on stating what it does "truncates a bigint"

test/node/bigint.test.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  it('Serializes bigints into little-endian byte order', function () {
+    const testDoc = { a: 0x1234567812345678n };
+    const serializedDoc = BSON.serialize(testDoc);
+    const expectedResult = bufferFromHexArray([
+      '12', // int64 type
+      '6100', // 'a' key with key null terminator
+      '7856341278563412'
+    ]);
+    expect(serializedDoc).to.deep.equal(expectedResult);
```
nbbeeken, line 23: Since we know the `DATA_TYPE_OFFSET` is always consistent, then the offset of the actual int is consistent too, can we assert that if we put the `expectedResult` in a DataView and use getBigInt64 with the LE flag set to true that it equals the input?

test/node/long.test.ts
```diff
@@ -0,0 +1,24 @@
...
+    expect(new Long('00').toString()).to.equal('0');
+    expect(new Long('-1').toString()).to.equal('-1');
+    expect(new Long('-1', true).toString()).to.equal('18446744073709551615');
+    expect(new Long('123456789123456789').toString()).to.equal('123456789123456789');
+    expect(new Long('123456789123456789', true).toString()).to.equal('123456789123456789');
+    expect(new Long('13835058055282163712').toString()).to.equal('-4611686018427387904');
+    expect(new Long('13835058055282163712', true).toString()).to.equal('13835058055282163712');
+  });
+
+  it('Accept BigInts in Long constructor', function () {
```
nbbeeken, line 15:
```suggestion
  it('accepts BigInts in Long constructor', function () {
```

src/parser/serializer.ts
```diff
@@ -676,7 +690,7 @@ export function serializeInto(
       } else if (typeof value === 'number') {
         index = serializeNumber(buffer, key, value, index);
       } else if (typeof value === 'bigint') {
-        throw new BSONTypeError('Unsupported type BigInt, please use Decimal128');
+        index = serializeBigInt(buffer, key, value, index);
```
nbbeeken, line 693: If you run `npm run check:coverage` you can actually view what lines of code get hit by our testing, and we currently don't hit this or the line in the Map section either, can we add two tests covering each code path?

nbbeeken (changes requested):

test/node/bigint.test.ts
```diff
@@ -1,89 +1,124 @@
...
+import { BSONDataView } from '../../src/utils/byte_utils'
 
-describe('BSON BigInt serialization Support', function () {
-  // Index for the data type byte of a BSON document with a single element
-  const DATA_TYPE_OFFSET = 4;
+describe('BSON BigInt serialization Support', function() {
+  // Index for the data type byte of a BSON document with a 
+  // NOTE: These offsets only apply for documents with the shape {a : <n>}
+  // where n is a BigInt
+  class SerializedDocParts {
```
nbbeeken, line 10:
```suggestion
  type SerializedDocParts = {
```

Since we're not actually using the class we can make this a TS only description of the object shape

test/node/bigint.test.ts
```diff
@@ -1,89 +1,124 @@
...
+    const dataView = BSONDataView.fromUint8Array(serializedDoc);
+    const keySlice = serializedDoc.slice(KEY_OFFSET);
+
+    let keyLength = 0;
+    while (keySlice[keyLength++] !== 0) { }
+
+    const valueOffset = KEY_OFFSET + keyLength;
+    const key = Buffer.from(serializedDoc.slice(KEY_OFFSET, KEY_OFFSET + keyLength)).toString('utf8');
+
+    const result = {
```
nbbeeken, line 33:
```suggestion
    return {
```

Just a nit, one less var

test/node/bigint.test.ts
```diff
@@ -1,89 +1,124 @@
...
+    key: string;
+    value: BigInt;
+  }
 
-  it('Serializes bigints with the correct BSON type', function () {
+  // NOTE: this function operates on serialized BSON documents with the shape { <key> : <n>}
+  // where n is some int64
+  // This function assumes that keys are properly encoded with the necessary null byte at the
+  // end and only at the end of the key
+  function getSerializedDocParts(serializedDoc: Uint8Array): SerializedDocParts {
```
nbbeeken, line 20:
```suggestion
  /** 
   * NOTE: this function operates on serialized BSON documents with the shape { <key> : <n>}
   * where n is some int64
   * This function assumes that keys are properly encoded with the necessary null byte at the
   * end and only at the end of the key
   */
  function getSerializedDocParts(serializedDoc: Uint8Array): SerializedDocParts {
```

Makes the comment visible in editor hovers

nbbeeken (changes requested):

nbbeeken (approved):

durran (approved):

PR #542, opened 2022-12-15
feat(NODE-4871): Add support for int64 deserialization to BigInt

W-A-James:
#### What is changing?

Adding capability to deserialize BSON `Long` to `BigInt`

Yes. Documentation changes in README and inline docs


src/parser/deserializer.ts
```diff
@@ -334,8 +345,11 @@ function deserializeObject(
         (buffer[index++] << 16) |
         (buffer[index++] << 24);
       const long = new Long(lowBits, highBits);
+      if (useBigInt64) {
+        value = BigInt(lowBits) | (BigInt(highBits) << 32n);
+      }
```
nbbeeken, line 350: Can we instead lean on the native API here, create and ask a DataView to getBigInt64 at this offset. I want to avoid the peculiarities of translating bit shifting to bigint creation. The logic you went with looks perfectly correct (+ tests agree) 

Also I never came up with a way to test this effectively, its something a linter should do, but I couldn't find an eslint rule off the bat, we want to avoid shipping `n` notation inside the BSON code 1. we don't need it, 2. those who aren't in environments that support it won't crash on import, instead they may crash on use. If we weren't changing to use DataView we'd want to change `<< 32n` to `<< BigInt(32)`. This requirement is called out in the design doc

test/node/bigint.test.ts
```diff
@@ -0,0 +1,111 @@
+import { BSON } from '../register-bson';
+import { bufferFromHexArray } from './tools/utils';
+import { expect } from 'chai';
+
+describe('BSON BigInt deserialization support', function () {
+  describe('BSON.deserialize()', function () {
+    let testSerializedDoc: Buffer;
+    before(function () {
+      testSerializedDoc = bufferFromHexArray(['12', '6100', '2300000000000000']);
+    });
```
nbbeeken, line 10: No need for a hook here, you can create this inline

test/node/bigint.test.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    it('deserializes int64 to Number when useBigInt64,promoteValues, promoteLongs are default', function () {
```
nbbeeken, line 12: This is a perfect opportunity for making table tests, we want to make sure we cover the cases where all three variables are set to true/false/undefined (unset) so we can generate that along with the expected outcome. I'll slack you some examples.

nbbeeken (changes requested):

test/node/bigint.test.ts
```diff
@@ -0,0 +1,102 @@
...
+    const useBigInt64Values = [true, false, undefined];
+    const promoteLongsValues = [true, false, undefined];
+    const promoteValuesValues = [true, false, undefined];
+
+    const testTable = useBigInt64Values.flatMap(useBigInt64 => {
+      return promoteLongsValues.flatMap(promoteLongs => {
+        return promoteValuesValues.flatMap(promoteValues => {
+          const useBigInt64Set = useBigInt64 ?? false;
+          const promoteLongsSet = promoteLongs ?? true;
+          const promoteValuesSet = promoteValues ?? true;
```
W-A-James, line 28: Should probably add comments here explaining that this is done to mirror5 the defaults used in the deserializer

nbbeeken: Should we let them remain undefined instead? that way we test the defaults set in the src

test/node/bigint.test.ts
```diff
@@ -0,0 +1,102 @@
...
+              shouldThrow,
+              expectedResult,
+              expectedErrorMessage
+            }
+          ];
+        });
+      });
+    });
+
+    it('meta test: should test table correctly', () => {
```
nbbeeken, line 53:
```suggestion
    it('meta test: generates 27 tests 5 error cases and 22 success cases', () => {
```

Just a bit more informational, if this were to ever change we'd change the title to go with it

test/node/bigint.test.ts
```diff
@@ -0,0 +1,102 @@
...
+        return `deserializes int64 to ${entry.expectedResult.a.constructor.name} when ${flagString}`;
+      }
+    }
+
+    function generateTest(test: TestTableEntry) {
+      const options = test.options;
+      const deserialize = () => {
+        return BSON.deserialize(testSerializedDoc, options);
+      };
+      if (test.shouldThrow) {
```
nbbeeken, line 83: Great idea to create the test callback, much better than putting the control flow logic into the test code itself 💡

nbbeeken (changes requested):

test/node/bigint.test.ts
W-A-James: The only place these `xxxSet` flags are used in the test is to determine what our expected result it. We directly pass the value of the option when creating the test. (See lines 41-48)

nbbeeken: Oh! missed the `Set` I agree with adding a comment, maybe more so to call out that these indicate Set/not Set

src/parser/deserializer.ts
```diff
@@ -334,8 +347,11 @@ function deserializeObject(
         (buffer[index++] << 16) |
         (buffer[index++] << 24);
       const long = new Long(lowBits, highBits);
+      if (useBigInt64) {
+        value = dataview.getBigInt64(0, true);
+      }
       // Promote the long if possible
-      if (promoteLongs && promoteValues === true) {
+      else if (promoteLongs && promoteValues === true) {
```
nbbeeken, line 354: Super nit, it's kinda weird to have the comment break the else if, can you either delete it or move it inside the `promoteLongs && promoteValues === true` condition

nbbeeken (changes requested):
last comment sorry teeny fmt fix

nbbeeken (dismissed):

nbbeeken (approved):

PR #543, opened 2022-12-15
feat(NODE-4892)!: error on bson types not from this version

nbbeeken:
#### What is changing?

- Added new BSONValue super class that all BSON type classes now extend from
- Added a version number to all BSON types via inheritance
- Added assertions to serialize, stringify, calculateObjectSize that all fail if the version is not the same as the constant

Explicitly breaks cases where old BSON types fed into this version of the library may be silently ignored / serialized in a corrupted form.


nbbeeken:
This PR still needs a migration guide update, but code is ready

src/binary.ts
```diff
@@ -31,6 +31,10 @@ export class Binary {
   get _bsontype(): 'Binary' {
     return 'Binary';
   }
+  /** @internal */
```
durran, line 34: Will all types always have the same version or do we expect in the future each type could have a different version for, say, the minimum serializer version it matches? The reason I ask is that if every type will always have the same version could we dry this up so it's only defined in one place and all the types use that?

durran (dismissed):

durran (approved):
Yeah I think I like this now even with the `super()` calls. Nice work.

src/parser/calculate_size.ts
```diff
@@ -77,9 +78,19 @@ function calculateElement(
     case 'boolean':
       return (name != null ? ByteUtils.utf8ByteLength(name) + 1 : 0) + (1 + 1);
     case 'object':
-      if (value == null || value['_bsontype'] === 'MinKey' || value['_bsontype'] === 'MaxKey') {
+      if (
+        value != null &&
+        typeof value._bsontype === 'string' &&
+        value[Symbol.for('@@mdb.bson.version')] !== constants.BSON_MAJOR_VERSION
+      ) {
+        throw new BSONError('Unsupported BSON version, bson types must be from bson 5.0 or later');
```
baileympearson, line 86: I've seen discussion about creating an error hierarchy for BSON.

Could we consider starting that now?  This seems like a good candidate for a specialized error, `BSONTypeVersionMismatchError`.

test/node/bson_type_classes.test.ts
```diff
@@ -0,0 +1,100 @@
...
+      it(`defines a toExtendedJSON() method`, () =>
+        expect(TypeClass.prototype).to.have.property('toExtendedJSON').that.is.a('function'));
+
+      it(`defines an inspect() method`, () =>
+        expect(TypeClass.prototype).to.have.property('inspect').that.is.a('function'));
+
+      it(`defines a [Symbol.for('nodejs.util.inspect.custom')]() method`, () =>
+        expect(TypeClass.prototype)
+          .to.have.property(Symbol.for('nodejs.util.inspect.custom'))
+          .that.is.a('function'));
```
baileympearson, line 97: If we're testing and asserting that all these properties exist on every single bson type, should we consider making them abstract properties on the BSONType base class?

test/node/cross_compat.test.ts
```diff
@@ -0,0 +1,83 @@
...
+  },
+  UUID: () => {
+    return new UUID('74e65f2f-6fdb-4c56-8785-bddb8ad79ea2');
+  }
+};
+
+describe('Prevent previous major versions from working with BSON v5 serialize and stringify', function () {
+  for (const [typeName, typeMaker] of Object.entries(BSONTypeClasses)) {
+    it(`serialize throws if ${typeName} is missing a version symbol`, () => {
+      const type = typeMaker();
```
baileympearson, line 69:
```suggestion
  for (const [typeName, typeFactory] of Object.entries(BSONTypeClasses)) {
    it(`serialize throws if ${typeName} is missing a version symbol`, () => {
      const type = typeFactory();
```

really optional nit - I usually see `Factory` used

test/node/bson_type_classes.test.ts
```diff
@@ -0,0 +1,100 @@
...
+  MaxKey,
+  ObjectId,
+  BSONRegExp,
+  BSONSymbol,
+  Timestamp,
+  UUID
+];
+
+const BSONTypeClassCtors = new Map<string, () => BSONValue>([
+  ['Binary', () => new Binary()],
```
baileympearson, line 39: Is there a reason that these need to be factory functions and we can't just hard-code instances of the classes?

src/symbol.ts
```diff
@@ -8,16 +10,17 @@ export interface BSONSymbolExtended {
  * @public
  * @category BSONType
  */
-export class BSONSymbol {
-  get _bsontype(): 'Symbol' {
-    return 'Symbol';
+export class BSONSymbol extends BSONValue {
+  get _bsontype(): 'BSONSymbol' {
+    return 'BSONSymbol';
   }
```
baileympearson, line 16: This is a public API we're changing, let's make sure we mention this in the migration guide.

baileympearson (changes requested):
We need to update the migration guide as well.

test/node/bson_type_classes.test.ts
nbbeeken: I've added what is possible to the abstract class (`_bsontype`, `inspect`, and `toExtendedJSON`). Static properties and non-wellknown symbols aren't enforceable by typescript

src/symbol.ts
nbbeeken: This is tricky, it is not public API it's only function is to power the serializer, but we also cannot mark it as internal because there would be a lot of TS that would complain b/c of it.

But yea I can add it to the migration guide it just is a bit confusing it was never meant to be consumed outside the library / driver

test/node/cross_compat.test.ts
nbbeeken: brb filing a ticket to make a type factory factory

test/node/bson_type_classes.test.ts
nbbeeken: If the constructors start throwing the tests crash on launch instead of inside the it block.

src/parser/calculate_size.ts
nbbeeken: Done

test/node/extended_json.test.ts
```diff
@@ -300,14 +300,16 @@ describe('Extended JSON', function () {
     expect(serialized).to.equal('{"a":10}');
   });
 
-  it('should throw if invalid BSON types are input to EJSON serializer', function () {
+  it.skip('should throw if invalid BSON types are input to EJSON serializer', function () {
+    // TODO This doesn't throw for the reason you'd expect it to
```
nbbeeken, line 304: Needs JIRA ticket, EJSON does not have the same assertions BSON does, this was passing for the wrong reason.

package.json
```diff
@@ -13,7 +13,7 @@
     "etc/prepare.js"
   ],
   "types": "bson.d.ts",
-  "version": "4.7.0",
+  "version": "5.0.0-alpha.2",
```
baileympearson, line 16: We need to revert package.json and package-lock changes too.

baileympearson (changes requested):

baileympearson (changes requested):
@nbbeeken migration guide should be updated here

src/symbol.ts
nbbeeken: > we also cannot mark it as internal

The driver inspects _bsontype at runtime and uses it in the `Filter` type logic

package.json
nbbeeken: We'll probably want to do an alpha.3 and get the driver PR settled first so might as well keep this so the next bump is correct

test/node/extended_json.test.ts
nbbeeken: https://jira.mongodb.org/browse/NODE-4952

src/constants.ts
```diff
@@ -1,3 +1,8 @@
+/** @internal */
+export type BSON_MAJOR_VERSION = 5;
+/** @internal */
+export const BSON_MAJOR_VERSION: BSON_MAJOR_VERSION = 5;
```
dariakp, line 4: why do we need the type and const gymnastics here? more specifically, why do we need the explicit `type` export for this?

src/error.ts
```diff
@@ -43,3 +43,14 @@ export class BSONError extends Error {
...
 }
+
+/** @public */
+export class BSONVersionError extends BSONError {
+  get name(): 'BSONVersionError' {
+    return 'BSONVersionError';
+  }
+
+  constructor() {
+    super('Unsupported BSON version, bson types must be from bson 5.0 or later');
```
dariakp, line 54: can we change this to use the const so that it can auto-update as needed when we do the next major version?

src/max_key.ts
```diff
@@ -8,7 +10,7 @@ export interface MaxKeyExtended {
  * @public
  * @category BSONType
  */
-export class MaxKey {
+export class MaxKey extends BSONValue {
```
dariakp, line 13: does this work without a constructor/super call?

src/min_key.ts
```diff
@@ -8,7 +10,7 @@ export interface MinKeyExtended {
  * @public
  * @category BSONType
  */
-export class MinKey {
+export class MinKey extends BSONValue {
```
dariakp, line 13: same question as with MaxKey

test/node/symbol.test.ts
```diff
@@ -0,0 +1,44 @@
...
+      '0E', // bson symbol
+      Buffer.from('sym\x00', 'utf8').toString('hex'),
+      '07000000', // 6 bytes
+      Buffer.from('symbol\x00').toString('hex')
+    ]);
+
+    expect(BSON.deserialize(bytes)).to.have.property('sym', 'symbol');
+  });
+
+  it('deserializes to BSONSymbol if promoteValues is false', () => {
```
dariakp, line 33: why is `promoteValues` significant here?

test/node/parser/serializer.test.ts
```diff
@@ -88,5 +91,13 @@ describe('serialize()', () => {
       expect(() => BSON.serialize(Buffer.alloc(2))).to.throw(/cannot be BSON documents/);
       expect(() => BSON.serialize(new Uint8Array(3))).to.throw(/cannot be BSON documents/);
     });
+
+    it(`throws if Symbol.for('@@mdb.bson.version') is the wrong version`, () => {
```
dariakp, line 95: how about if the symbol is not defined? (as it presumably wouldn't be for older versions of bson)

test/node/parser/calculate_size.test.ts
```diff
@@ -1,10 +1,19 @@
...
+import { BSONVersionError } from '../../register-bson';
 
 describe('calculateSize()', () => {
   it('should only enumerate own property keys from input objects', () => {
     const input = { a: 1 };
     Object.setPrototypeOf(input, { b: 2 });
     expect(BSON.calculateObjectSize(input)).to.equal(12);
   });
+
+  it(`throws if Symbol.for('@@mdb.bson.version') is the wrong version`, () => {
```
dariakp, line 12: how about if it's not defined?

docs/upgrade-to-v5.md
```diff
@@ -288,3 +288,9 @@ try {
   throw error;
 }
 ```
+
+### Explicit cross version incompatibility
+
+Starting with v5.0.0 of the BSON library instances of types from previous versions will throw an error when passed to the serializer.
```
dariakp, line 294: can we add a code example of what users would expect to see? (might be nice to call out the new error type)

dariakp (changes requested):

src/constants.ts
nbbeeken: This made more sense when we had the many getters approach and I wanted TS to check the return type strictly. fixed!

src/error.ts
nbbeeken: done!

src/max_key.ts
nbbeeken: Yep, constructor is optional, if one is added both JS runtime / TS will require the author to call super first.

src/min_key.ts
nbbeeken: same answer

test/node/parser/calculate_size.test.ts
nbbeeken: Good call, added a test case

test/node/symbol.test.ts
nbbeeken: promoteValues controls whether a string or BSONSymbol instance is returned. I wrote these test b/c tests were failing in the driver but not here in BSON early on in this PR

test/node/parser/serializer.test.ts
nbbeeken: added case

docs/upgrade-to-v5.md
nbbeeken: added! let me know if we should say more about the `BSONVersionError`

docs/upgrade-to-v5.md
```diff
@@ -288,3 +288,19 @@ try {
...
+This is to ensure that types are always serialized correctly and that there is no unexpected silent BSON serialization mistakes that could occur when mixing versions.
+It's unexpected for any applications to have more than one version of the BSON library but with nested dependencies and re-exporting, this new error will illuminate those incorrect combinations.
```
baileympearson, line 296:
```suggestion
```

Are these lines really necessary?  I don't think providing a justification for this feature is necessary

dariakp: I think we should keep the lines in and then have Alex pare down the content if it seems extraneous (in the bson v5 guide docs ticket). Err on the side of more context rather than less when documenting the individual changes.

dariakp (approved):

src/bson_value.ts
```diff
@@ -0,0 +1,18 @@
+import { BSON_MAJOR_VERSION } from './constants';
+
+/** @public */
+export abstract class BSONValue {
```
dariakp, line 4: forgot we still have this as a TBD, let's confirm with team before merging

dariakp (changes requested):

src/bson_value.ts
dariakp: consensus to keep the current name

dariakp (approved):

docs/upgrade-to-v5.md
baileympearson: Okay, that's fine

baileympearson (approved):

PR #547, opened 2023-01-03
feat(NODE-4873): support EJSON stringify from BigInt to $numberLong

W-A-James:
#### What is changing?

Adding capability to stringify BigInt

No

### DIFF
--- a/src/extended_json.ts
+++ b/src/extended_json.ts
@@ -194,8 +194,8 @@ function serializeValue(value: any, options: EJSONSerializeOptions): any {
 
       throw new BSONError(
         'Converting circular structure to EJSON:\n' +
-        `    ${leadingPart}${alreadySeen}${circularPart}${current}\n` +
-        `    ${leadingSpace}\\${dashes}/`
+          `    ${leadingPart}${alreadySeen}${circularPart}${current}\n` +
+          `    ${leadingSpace}\\${dashes}/`
       );
     }
     options.seenObjects[options.seenObjects.length - 1].obj = value;
@@ -236,17 +236,9 @@ function serializeValue(value: any, options: EJSONSerializeOptions): any {
 
   if (typeof value === 'bigint') {
     if (!options.relaxed) {
-      // Interpret as smallest BSON integer type that can represent the number exactly
-      if (value >= BSON_INT32_MIN && value <= BSON_INT32_MAX) {
-        return { $numberInt: value.toString() };
-      }
-      if (value >= BSON_INT64_MIN && value <= BSON_INT64_MAX) {
-        return { $numberLong: value.toString() };
-      }
-      // Fallback to double if number is out of signed-64-bit int range
-      return { $numberDouble: value.toString() };
+      return { $numberLong: BigInt.asIntN(64, value).toString() };
     }
-    return Number(BigInt.asIntN(64, value)); 
+    return Number(BigInt.asIntN(64, value));
   }
 
   if (value instanceof RegExp || isRegExp(value)) {
--- a/src/extended_json.ts
+++ b/src/extended_json.ts
@@ -194,8 +194,8 @@ function serializeValue(value: any, options: EJSONSerializeOptions): any {
 
       throw new BSONError(
         'Converting circular structure to EJSON:\n' +
-          `    ${leadingPart}${alreadySeen}${circularPart}${current}\n` +
-          `    ${leadingSpace}\\${dashes}/`
+        `    ${leadingPart}${alreadySeen}${circularPart}${current}\n` +
+        `    ${leadingSpace}\\${dashes}/`
       );
     }
     options.seenObjects[options.seenObjects.length - 1].obj = value;
@@ -232,6 +232,10 @@ function serializeValue(value: any, options: EJSONSerializeOptions): any {
       }
     }
     return { $numberDouble: Object.is(value, -0) ? '-0.0' : value.toString() };
+  }
+
+  if (typeof value === 'bigint' && (!options.relaxed)) {
+    return { $numberLong: value.toString() };
   }
 
   if (value instanceof RegExp || isRegExp(value)) {
--- a/src/extended_json.ts
+++ b/src/extended_json.ts
@@ -234,8 +234,11 @@ function serializeValue(value: any, options: EJSONSerializeOptions): any {
     return { $numberDouble: Object.is(value, -0) ? '-0.0' : value.toString() };
   }
 
-  if (typeof value === 'bigint' && (!options.relaxed)) {
-    return { $numberLong: value.toString() };
+  if (typeof value === 'bigint') {
+    if (!options.relaxed) {
+      return { $numberLong: value.toString() };
+    }
+    return Number(BigInt.asIntN(64, value)); // FIXME(NODE-4873): This is yucky
   }
 
   if (value instanceof RegExp || isRegExp(value)) {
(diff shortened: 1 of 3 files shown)


src/extended_json.ts
```diff
@@ -234,6 +234,13 @@ function serializeValue(value: any, options: EJSONSerializeOptions): any {
     return { $numberDouble: Object.is(value, -0) ? '-0.0' : value.toString() };
   }
 
+  if (typeof value === 'bigint') {
+    if (!options.relaxed) {
+      return { $numberLong: BigInt.asIntN(64, value).toString() };
```
W-A-James, line 239: This is done to truncate the BigInt to 64bits to mirror the new behaviour of `BSON.serialize` for BigInt

test/node/extended_json.test.ts
```diff
@@ -180,6 +180,30 @@ describe('Extended JSON', function () {
...
+    const serialized = EJSON.stringify(number, { relaxed: true });
+    expect(serialized).to.equal('{"a":10000}');
+  });
+
+  it('loses precision when serializing bigint values outside of range [Number.MIN_SAFE_INTEGER, Number.MAX_SAFE_INTEGER] in relaxed mode', function () {
+    const numbers = { a: -(2n ** 53n) - 1n, b: 2n ** 53n + 2n };
+    const serialized = EJSON.stringify(numbers, { relaxed: true });
+    expect(serialized).to.equal('{"a":-9007199254740992,"b":9007199254740994}');
+  });
+
```
nbbeeken, line 206: The truncation being the same in BSON and EJSON is an "interesting" case, can we add a test that runs serialize and stringify and we get the "same" results. And same here will mean same in value interpretation, one is going to be bytes and the other will be a string. 
Turning the bytes back into a bigint within the test and checking its stringified value should result in the same string EJSON emits.

nbbeeken (changes requested):

nbbeeken (approved):

nbbeeken (changes requested):
We discussed on slack one more test about loose equality working, and js string == EJSON string

src/extended_json.ts
```diff
@@ -234,6 +234,13 @@ function serializeValue(value: any, options: EJSONSerializeOptions): any {
...
+    }
+    return Number(BigInt.asIntN(64, value));
```
nbbeeken, line 241: Let's add examples of this to the design doc so we can be sure we're all aligned on the choice we made here.

nbbeeken (changes requested):

nbbeeken (approved):
LGTM

test/node/extended_json.test.ts
```diff
@@ -181,6 +182,92 @@ describe('Extended JSON', function () {
     expect(EJSON.parse(serialized)).to.deep.equal(numbers);
   });
 
+  it('truncates bigint values when they are outside the range [BSON_INT64_MIN, BSON_INT64_MAX] in canonical mode', function () {
```
W-A-James, line 185: Should I move these tests to live with the other BigInt support tests in `bigint.test.ts`?

nbbeeken (dismissed):

src/extended_json.ts
```diff
@@ -234,6 +234,13 @@ function serializeValue(value: any, options: EJSONSerializeOptions): any {
     return { $numberDouble: Object.is(value, -0) ? '-0.0' : value.toString() };
   }
 
+  if (typeof value === 'bigint') {
+    if (!options.relaxed) {
```
baileympearson, line 238: so this is pretty last minute in the review process, so this is more of a suggestion / general idea.  if you want to make the refactor, we should run it by the team since it deviates from the existing pattern here and we'd want to make sure everyone is alright with it.

An alternate approach here could be converting the bigint into a long and then stringify-ing the long:

```typescript
return Long.fromBigInt(value).toExtendedJSON();
```

this isn't exactly as straightforward as the code snippet because

1. we'd want to do the truncation in `Long.fromBigInt`, but that's already public api so we can't change the behavior (although, we _could_, just a larger scope)
2. so we may want to create some other internal static factory on the Long class

But the benefit to this approach is that we're guaranteed that the actual stringification is exactly the same for big ints and longs (which is what we want conceptually).

test/node/bigint.test.ts
```diff
@@ -263,4 +263,92 @@ describe('BSON BigInt support', function () {
       expect(serializedMap).to.deep.equal(expectedSerialization);
     });
   });
+
+  describe('EJSON.stringify()', function () {
```
baileympearson, line 267: I'm going to leave two general test organization comments here, both of which boil down to "context blocks are great" 😄 

---

It can be really useful to group by behavior using context blocks.  There are many ways to do it (you can group by option, by functionality, ect).  I think in this case, grouping by option makes sense.  I would group any tests specific to canonical ejson and any tests specific to relaxed ejson together.  For example (not a complete refactor, just an example):

```typescript
context('when relaxed=true (canonical mode)', function() {
  it('truncates bigint values when they are outside the range [BSON_INT64_MIN, BSON_INT64_MAX]', function () { ... });
  it('truncates bigint values in the same way as BSON.serialize', function () { ... });
});
```

This

- does not clutter each test description with all the conditions that make up the test
- related to the above, this makes the output of the tests much easier to read
- can help identify code paths you might miss otherwise.  when you group by options / conditions, it's easy to see if you're missing a condition

That's the first change I'd suggest.

---

Related to the above, the "condition" should always come before the "expectation" in test output and in test structure (that's my opinion but one I hold rather strongly).  I think contexts make this easy because they break the condition apart from the expectation.  The example, the following is easier to grok than a single test with the description "produces bigint strings that are equal to the strings generated when using BigInt.toString when the bigint values used are 64 bits wide or less":

```typescript
context('when the bigint values used are 64 bits wide or less', function() {
  it('produces identical values to `BigInt.toString`', () => { ... });
});
```

Grouping by condition (my first point) helps here, but I always use contexts even when I have a single test because it helps with clarity.

baileympearson (changes requested):

src/extended_json.ts
nbbeeken: https://jira.mongodb.org/browse/NODE-4954
Filed this to improve the way we obtain a long from a bigint. Reusing the current logic would mean stringifying the number to calculate the high/low bits only to stringify it again, and undoing that here and confirming total test coverage takes the scope a bit further than necessary for the goal here

baileympearson: Can you clarify?  I don't see why this requires additional test coverage.

This entirely reuses existing Long construction logic (presumably tested) and ejson stringification logic (tested).  The only new behavior we'd need to test is that bigints > 64 bits are truncated properly (straightforward to test).

nbbeeken: True there's implicit coverage through the EJSON tests, good catch. Still the string -> number -> string steps are unnecessary we can avoid those translation steps. long.toString even references div operations being expensive in code comments

W-A-James: I think I agree with Neal here. The existing logic for `Long.fromBigInt` definitely does work, but it is also as Neal said, the code path is overly complicated and repeats work. The code path is `Long.fromBigInt` which has to call `BigInt.toString`, passing that result back to `Long.fromString` and then going ahead and stringifying the resulting `Long`. The time loss from repeating this work might be negligible, but I think in principle it still is preferable to avoid this.

nbbeeken (approved):

baileympearson (approved):
test output is nicely organized 

https://parsley.mongodb.com/evergreen/node_bson_linux_node_tests_v18_patch_3b4b61ed8dfbe7e7f8635e4584e1b17a91c1c4df_63c1bf9032f4171734ce4a31_23_01_13_20_31_14/0/task?bookmarks=0,2871&selectedLine=168

PR #550, opened 2023-01-05
feat(NODE-4927): exports in package.json for react native and document how to polyfill for BSON

nbbeeken:
#### What is changing?

Add react-native exports specifier.
Set it to the commonjs bundle.
Document polyfills.

To support react native.

### DIFF
--- a/README.md
+++ b/README.md
@@ -337,7 +337,7 @@ Finally, import the `BSON` library like so:
 import { BSON, EJSON } from 'bson';
 ```
 
-This will resolve the `node_modules/bson/lib/bson.cjs` bundle per the setting we have in the `"exports"` section of our [package.json](./package.json).
+This will cause React Native to import the `node_modules/bson/lib/bson.cjs` bundle (see the `"react-native"` setting we have in the `"exports"` section of our [package.json](./package.json).)
 
 ### Technical Note about React Native module import
 
--- a/README.md
+++ b/README.md
@@ -339,7 +339,9 @@ import { BSON, EJSON } from 'bson';
 
 This will resolve the `node_modules/bson/lib/bson.cjs` bundle per the setting we have in the `"exports"` section of our [package.json](./package.json).
 
-> *Reasoning:* React Native must use the commonjs syntax version of BSON to avoid the unsupported syntax of a top-level await that is used in the es module build.
+### Technical Note about React Native module import
+
+The `"exports"` definition in our `package.json` will result in BSON's CommonJS bundle being imported in a React Native project instead of the ES module bundle.  Importing the CommonJS bundle is necessary because BSON's ES module bundle of BSON uses top-level await, which is not supported syntax in [React Native's runtime hermes](https://hermesengine.dev/).
 
 ## FAQ
 
--- a/README.md
+++ b/README.md
@@ -311,10 +311,7 @@ try {
 
 ## React Native
 
-BSON requires `TextEncoder`, `TextDecoder`, `atob`, `btoa`, and `crypto.getRandomValues` to be available on the global object in browser-like environments.
-
-Since React Native has a different JS engine with different globals than either browser or Node.js has, polyfills of those APIs are required to get React Native projects working with BSON or EJSON.
-
+BSON requires that `TextEncoder`, `TextDecoder`, `atob`, `btoa`, and `crypto.getRandomValues` are available globally.  These are present in most Javascript runtimes but require polyfilling in React Native.  Polyfills for the missing functionality can be installed with the following command:
 ```sh
 npm install --save react-native-get-random-values text-encoding-polyfill base-64
 ```
--- a/README.md
+++ b/README.md
@@ -309,6 +309,41 @@ try {
 }
 ```
 
+## React Native
+
+BSON requires `TextEncoder`, `TextDecoder`, `atob`, `btoa`, and `crypto.getRandomValues` to be available on the global object in browser-like environments.
+
+Since React Native has a different JS engine with different globals than either browser or Node.js has, polyfills of those APIs are required to get React Native projects working with BSON or EJSON.
+
+```sh
+npm install --save react-native-get-random-values text-encoding-polyfill base-64
+```
+
+The following snippet should be placed at the top of your index.js file for React Native projects using the BSON library.
+
+```typescript
+// Required Polyfills For ReactNative
+import {encode, decode} from 'base-64';
+if (global.btoa == null) {
+  global.btoa = encode;
+}
+if (global.atob == null) {
+  global.atob = decode;
+}
+import 'text-encoding-polyfill';
+import 'react-native-get-random-values';
+```
+
+Import the BSON library like so:
+
+```typescript
+import { BSON, EJSON } from 'bson';
+```
+
+This will resolve the `node_modules/bson/lib/bson.cjs` bundle per the setting we have in the `"exports"` section of our [package.json](./package.json).
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -309,6 +309,41 @@ try {
 }
 ```
 
+## React Native
+
+BSON requires `TextEncoder`, `TextDecoder`, `atob`, `btoa`, and `crypto.getRandomValues` to be available on the global object in browser-like environments.
+
+Since React Native has a different JS engine with different globals than either browser or Node.js has, polyfills of those APIs are required to get React Native projects working with BSON or EJSON.
+
```
baileympearson, line 317:
```suggestion
BSON requires that `TextEncoder`, `TextDecoder`, `atob`, `btoa`, and `crypto.getRandomValues` are present on the `global` object.  These are present in most Javascript runtimes but require polyfilling in React Native.  Polyfills for the missing functionality can be installed with the following command:
```

README.md
```diff
@@ -309,6 +309,41 @@ try {
...
+```sh
+npm install --save react-native-get-random-values text-encoding-polyfill base-64
+```
+
+The following snippet should be placed at the top of your index.js file for React Native projects using the BSON library.
```
baileympearson, line 322:
```suggestion
The following snippet should be placed at the top of the entrypoint for React Native projects using the BSON library.  These lines must be placed for any code that imports `BSON`.
```

README.md
```diff
@@ -309,6 +309,41 @@ try {
...
+  global.btoa = encode;
+}
+if (global.atob == null) {
+  global.atob = decode;
+}
+import 'text-encoding-polyfill';
+import 'react-native-get-random-values';
+```
+
+Import the BSON library like so:
```
baileympearson, line 337:
```suggestion
Finally, import the `BSON` library like so:
```

README.md
```diff
@@ -309,6 +309,41 @@ try {
...
+
+```typescript
+import { BSON, EJSON } from 'bson';
+```
+
+This will resolve the `node_modules/bson/lib/bson.cjs` bundle per the setting we have in the `"exports"` section of our [package.json](./package.json).
+
+> *Reasoning:* React Native must use the commonjs syntax version of BSON to avoid the unsupported syntax of a top-level await that is used in the es module build.
```
baileympearson, line 345: Is this really important information to share with users?

baileympearson (changes requested):

README.md
nbbeeken: Yea, it is technical but I think it is worth mentioning b/c if it turns into an issue it's useful to know that it is our intention to aim the react native code at the commonjs module (a nodejs format) despite the technology being browser based. It works because RN will use the browser based APIs + bundling doesn't make `require` usage a problem.

README.md
```diff
@@ -309,6 +309,38 @@ try {
...
+
+Finally, import the `BSON` library like so:
+
+```typescript
+import { BSON, EJSON } from 'bson';
+```
+
+This will resolve the `node_modules/bson/lib/bson.cjs` bundle per the setting we have in the `"exports"` section of our [package.json](./package.json).
+
+> *Reasoning:* React Native must use the commonjs syntax version of BSON to avoid the unsupported syntax of a top-level await that is used in the es module build.
```
baileympearson, line 342: thoughts on making this a separate section?

```suggestion
### Note

The steps in this document will result in BSON's CommonJS bundle being imported in a React Native project instead of the ESModule bundle.  Importing the CommonJS bundle is necessary because BSON's ESModule bundle of BSON uses top-level async/await, which is not supported syntax in React Native.
```

README.md
baileympearson: see comment below

README.md
nbbeeken: Edited a bit locally, take a look

README.md
```diff
@@ -309,6 +309,40 @@ try {
...
+import 'react-native-get-random-values';
+```
+
+Finally, import the `BSON` library like so:
+
+```typescript
+import { BSON, EJSON } from 'bson';
+```
+
+This will resolve the `node_modules/bson/lib/bson.cjs` bundle per the setting we have in the `"exports"` section of our [package.json](./package.json).
```
baileympearson, line 340:
```suggestion
This will cause React Native to import the `node_modules/bson/lib/bson.cjs` bundle (see the `react native` setting we have in the `"exports"` section of our [package.json](./package.json).)
```

baileympearson (approved):

package.json
```diff
@@ -75,11 +75,11 @@
   },
   "main": "./lib/bson.cjs",
   "module": "./lib/bson.mjs",
-  "browser": "./lib/bson.mjs",
   "exports": {
-    "browser": "./lib/bson.mjs",
     "import": "./lib/bson.mjs",
-    "require": "./lib/bson.cjs"
+    "require": "./lib/bson.cjs",
+    "react-native": "./lib/bson.cjs",
```
kraenhansen, line 81: I'm not 100% sure on this, but I think the `react-native` property should be one level up, looking at [the docs for the Metro config](https://facebook.github.io/metro/docs/configuration/#resolvermainfields) and [the spec that it references](https://github.com/defunctzombie/package-browser-field-spec).

Was this manually tested in a React Native project?

README.md
```diff
@@ -309,6 +309,40 @@ try {
...
+```typescript
+// Required Polyfills For ReactNative
+import {encode, decode} from 'base-64';
+if (global.btoa == null) {
+  global.btoa = encode;
+}
+if (global.atob == null) {
+  global.atob = decode;
+}
+import 'text-encoding-polyfill';
```
kraenhansen, line 330: It would be great if these could be included in the React Native bundle though.

I don't think it's too much to warn users and have them optionally import he `react-native-get-random-values` package for added security, but having these polyfills as strict requirement for React Native users (hence all Realm JS users) is a no-go in my opinion.

kraenhansen (commented):
I have concerns about the need for end-users to provide these polyfills.

README.md
baileympearson: That's a valid concern.  

@nbbeeken would know better than I about the difficultly of bundling these the required dependencies in the react native build. 

As an alternative, could realm provide these polyfills for users instead of BSON?

package.json
nbbeeken: Yes, I made a new project with `react-native-cli` and confirmed that this works with the bundling that comes with that setup, not sure if that is Metro. The `"exports"` property is standardized for [nodejs](https://nodejs.org/dist/latest-v18.x/docs/api/packages.html#exports), [webpack](https://webpack.js.org/guides/package-exports/), and rollup, I suspect it may just not be explicitly mentioned there? 

We can get this into an alpha release and maybe you have projects with more diverse tooling worth testing to see if it gets pulled in correctly, I'll ping you on slack when its available

README.md
nbbeeken: Let's file a ticket for providing a polyfilled bundle for RN and we can fill it out with more detail on how the bundle should behave given that there is a mix of platforms and javascript engines that can be combined here.

Just adding background, one of our goals was to move away from bundling polyfills since they became part of the semantic version of this package despite being outside of our control. We became unable to pull in updates to the polyfills as they would break existing users but prevented other users from puling in the needed updates. Keeping polyfilling outside of a library's responsibility understandably adds friction but allows those that need the specific APIs working in a specific way to have control over that since not every polyfill is perfect for every use case.

kraenhansen: I understand that problem, we have similar issues in the Realm JS package. Ideally the bson package's source code should be runtime independent and therefore not rely on these globals in the first place. Instead it could have these functions "injected" by a platform specific index.ts wrapping the actual library. We're doing similar tricks on the upcoming refactor of the Realm JS SDK.

PR #552, opened 2023-01-05
feat(NODE-4874): support EJSON parse for BigInt from $numberLong

W-A-James:
### Description

Adding support to parse `bigint`s from EJSON

No

Continue BigInt support improvement.


src/extended_json.ts
```diff
@@ -362,13 +370,18 @@ function serializeDocument(doc: any, options: EJSONSerializeOptions) {
  */
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 function parse(text: string, options?: EJSONOptions): any {
+  options = {
+    useBigInt64: options?.useBigInt64 ?? false,
+    relaxed: options?.relaxed ?? true,
+    legacy: options?.legacy ?? false
+  };
```
W-A-James, line 377: Did this to clean up how default options are handled. It is not handled cleanly in a number of places in this file

src/extended_json.ts
```diff
@@ -70,16 +72,22 @@ const keysToCodecs = {
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 function deserializeValue(value: any, options: EJSONOptions = {}) {
   if (typeof value === 'number') {
+    const in32BitRange = value <= BSON_INT32_MAX && value >= BSON_INT32_MIN;
+    const in64BitRange = value <= BSON_INT64_MAX && value >= BSON_INT64_MIN;
```
nbbeeken, line 76: We should start a thread with the team, this auto upgrading of an int64 is an outstanding bug with respect to Longs. its questionable whether we want bigints to do the correct thing or mirror the current behavior. (NODE-4377).

src/extended_json.ts
```diff
@@ -70,16 +72,22 @@ const keysToCodecs = {
...
+    if (options.useBigInt64 && !in32BitRange && in64BitRange) {
+      return BigInt(value);
```
nbbeeken, line 78: We've confirmed that the value is typeof number but not that it is an integer, won't this prevent falling through to the Double case?

src/extended_json.ts
```diff
@@ -362,13 +370,18 @@ function serializeDocument(doc: any, options: EJSONSerializeOptions) {
  */
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 function parse(text: string, options?: EJSONOptions): any {
+  options = {
```
nbbeeken, line 373: LGTM, although I think we've preferred to create a new variable rather than overriding the options argument variable. ex. `const ejsonOptions = ...`

nbbeeken (changes requested):
Starting with code change comments, looking at the tests now

src/long.ts
```diff
@@ -1021,9 +1021,20 @@ export class Long {
...
   }
-  static fromExtendedJSON(doc: { $numberLong: string }, options?: EJSONOptions): number | Long {
-    const result = Long.fromString(doc.$numberLong);
-    return options && options.relaxed ? result.toNumber() : result;
+  static fromExtendedJSON(
+    doc: { $numberLong: string },
+    options?: EJSONOptions
+  ): number | Long | bigint {
+    const longResult = Long.fromString(doc.$numberLong);
+    const defaults = { useBigInt64: false, relaxed: true, legacy: false };
```
nbbeeken, line 1029: legacy doesn't modify the returned value so we don't need to include it here

test/node/bigint.test.ts
```diff
@@ -263,4 +263,113 @@ describe('BSON BigInt support', function () {
...
+      useBigInt64: boolean | undefined;
+      relaxed: boolean | undefined;
+    };
+    type TestTableEntry = {
+      options: ParseOptions;
+      expectedResult: BSON.Document;
+    };
+
+    const useBigInt64Values = [true, false, undefined];
+    const relaxedValues = [true, false, undefined];
```
nbbeeken, line 278: Can you leave a comment here that we don't need to add "legacy" to the matrix of tests because it does not change the output of a Long. And then to make sure things stay that way two small tests over in long.test.ts that calls fromExtendedJSON with and without legacy set and asserts the result is the same regardless

nbbeeken (changes requested):

src/extended_json.ts
nbbeeken: From the specs:

> If the number is an integer, parsers SHOULD interpret it as being of the smallest BSON integer type that can represent the number exactly.

okay so from the specs, the behavior is correct here, sorry that linked ticket is with respect to something else. And I see the examples from the design doc aligns with this so disregard. When the output would've been a Long we're changing it to bigint, but when outputs would've been an int32 it stays as a number/Int32 depending on relaxed.

src/extended_json.ts
W-A-James: Good catch here. Seems like this should have been caught by a test.

src/extended_json.ts
```diff
@@ -70,17 +72,23 @@ const keysToCodecs = {
...
 
     if (Number.isInteger(value) && !Object.is(value, -0)) {
       // interpret as being of the smallest BSON integer type that can represent the number exactly
-      if (value >= BSON_INT32_MIN && value <= BSON_INT32_MAX) {
+      if (in32BitRange) {
         return new Int32(value);
       }
-      if (value >= BSON_INT64_MIN && value <= BSON_INT64_MAX) {
+      if (in64BitRange) {
         // TODO(NODE-4377): EJSON js number handling diverges from BSON
```
nbbeeken, line 88: This TODO was related to the int64BitRange boolean logic, can you move it up to be with that variable

src/extended_json.ts
```diff
@@ -70,17 +72,23 @@ const keysToCodecs = {
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 function deserializeValue(value: any, options: EJSONOptions = {}) {
   if (typeof value === 'number') {
+    const in32BitRange = value <= BSON_INT32_MAX && value >= BSON_INT32_MIN;
+    const in64BitRange = value <= BSON_INT64_MAX && value >= BSON_INT64_MIN;
+
     if (options.relaxed || options.legacy) {
-      return value;
+      return Number.isInteger(value) && options.useBigInt64 ? BigInt(value) : value;
```
nbbeeken, line 79: I don't think relaxed: true is supposed to have any bigint logic, looking at the examples we should keep returning value here

src/long.ts
```diff
@@ -1021,9 +1021,20 @@ export class Long {
...
-    return options && options.relaxed ? result.toNumber() : result;
+  static fromExtendedJSON(
+    doc: { $numberLong: string },
+    options?: EJSONOptions
+  ): number | Long | bigint {
+    const longResult = Long.fromString(doc.$numberLong);
+    const defaults = { useBigInt64: false, relaxed: true };
+    const ejsonOptions = { ...defaults, ...options };
+    if (ejsonOptions.useBigInt64) {
+      return BigInt(doc.$numberLong);
```
nbbeeken, line 1032: Unlike in BSON and the 'number' case for EJSON, there isn't restrictions on value size or length.

Luckily BigInt casting function will throw an error for bad syntax unlike Number (yay!) (can we add a test? `{$numberLong: "blah!"}`)

But we have added support here for syntax that should not work: 
- `{ $numberLong: "0xA" }` should fail
- `{ $numberLong: "0o7" }` should fail
- `{ $numberLong: "0b1" }` should fail

And we've unintentionally allowed strings of any length to be provided here, I'm not sure what would be the best approach here, we can either figure out some way to assert some maximum length. Or we can construct a Long.fromString and use long.toBigInt here, asserting that the fromString logic would prevent an extremely long string from being converted.

nbbeeken (changes requested):

src/long.ts
nbbeeken: From discussion: We want to assert a maximum length on the string, this isn't the same as what Long does but our newer API can be better about preventing malformed input. There should be no leading zeros permitted but a plus or minus sign needs to be supported. We can iterate the string or use regexp to assert this.

Consider making a new BSONError subclass

test/node/bigint.test.ts
```diff
@@ -367,7 +364,7 @@ describe('BSON BigInt support', function () {
       }
     });
 
-    describe('relaxed double input where double is outside of int32 range and useBigInt64 is true', function () {
+    describe('relaxed double input where double is outside of int32 range and useBigInt64 is true', function() {
```
nbbeeken, line 367: We can add a test that checks BigInt usage is all behind gated logic. Take a look at the example where we test the library with and without a crypto global here: https://github.com/mongodb/js-bson/blob/main/test/node/byte_utils.test.ts#L603

We can load the BSON library with BigInt set to undefined, that doesn't cover using trailing n syntax but it's something

nbbeeken (changes requested):

src/long.ts
addaleax: > There should be no leading zeros permitted

They are currently permitted when deserializing to `Long`, as far as I can tell. I would recommend to keep validation consistent between deserializing to `Long` and deserializing to `bigint` (and changing validation for deserializing to `Long` would probably be breaking, but this might be exactly the right time for that if it’s something you want to do).

nbbeeken: If we're all in agreement (@dariakp @baileympearson @durran @W-A-James) I think we can do the validation for both `bigint` and `Long` cases, I was aiming to only impact the bigint case here since it would be opt-in, but useBigInt64 and increasing the validation are divergent concerns so we shouldn't overload the option's behavior.

We certainly don't want to add support for the hex, oct, and bin formats just because useBigInt64 is turned on, so that assertion I think we're all on board with.

`Long.fromString` currently allows a number of things that probably should be errors like 
* `"-----------123"`
* `"00000000123"`
* `"12.23"` - becomes 12
* `'8'.repeat(25)` - too large, can be determined by string length (gets truncated)
* `"9223372036854775808"` - too large int64.max + 1 (gets truncated)

We can decide to apply all, some, or none of the above validations, I think the most valuable and least pedantic is applying a string length check, but they're all great if we want to improve things :)

src/long.ts
```diff
@@ -75,6 +75,21 @@ const INT_CACHE: { [key: number]: Long } = {};
...
+ * @internal
+ * Validate an int64 string.
+ *
+ * Fails on strings longer than 22 characters
+ * Fails on non-decimal strings */
+function validateInt64String(input: string): boolean {
+  if (input.length > MAX_INT64_STRING_LENGTH) {
+    return false;
+  }
+  return /^(\+|-)?(0|[1-9][0-9]*)$/.test(input);
```
W-A-James, line 90: Would appreciate more eyes on this regex just to be safe

src/long.ts
W-A-James: I agree that at the very least, we should add the string length check to inputs of `fromExtendedJSON` in both the `bigInt` and `Long` cases to keep with the canonical EJSON spec and to prevent hanging on strings that would be too long to be a valid 64-bit integer and may be the result of malicious user input.

Canonical EJSON representation of a BSON.Long:

> {"$numberLong": <64-bit signed integer as a string>}

src/long.ts
```diff
@@ -75,6 +75,21 @@ const INT_CACHE: { [key: number]: Long } = {};
 /** A cache of the Long representations of small unsigned integer values. */
 const UINT_CACHE: { [key: number]: Long } = {};
 
+const MAX_INT64_STRING_LENGTH = 22;
```
addaleax, line 78: Is this value right?

```js
> String(1n << 64n).length
20
```

src/long.ts
```diff
@@ -75,6 +75,21 @@ const INT_CACHE: { [key: number]: Long } = {};
...
+
+/**
+ * @internal
+ * Validate an int64 string.
```
addaleax, line 82: If this code is kept as-it and doesn’t actually validate that the string is a 64-bit integer, only that it fits into that length in decimal and is an integer, then the doc comment should probably be clear about that (because I’d expect a method named `validateInt64String` to actually verify that the value is a valid 64-bit integer :))

src/long.ts
W-A-James: Ah, you're right, but it should be 21 since we want to account for a leading minus sign
```javascript
> String(-(1n << 64n)).length
21
```

nbbeeken: correct me if I'm wrong, but I think its still 20 cus those code snippets are the unisigned max values?
```ts
> String(-(2n ** 63n)).length
20
```

W-A-James: Ah, you were right @addaleax, my bad 🙏🏿

test/node/bigint.test.ts
W-A-James: Would it be appropriate to do this in another PR as just removing the `BigInt` field from the `globalThis` object isn't enough to be sure that our BigInt work doesn't break anything in environments that don't support BigInts. We'd have to go through the ES BigInt spec and make sure that we're not using any of the associated functionality and that seems outside the scope of this PR.

src/long.ts
```diff
@@ -1043,9 +1029,14 @@ export class Long {
   ): number | Long | bigint {
     const defaults = { useBigInt64: false, relaxed: true };
     const ejsonOptions = { ...defaults, ...options };
+    const decimalRegEx = /^(\+?0|(\+|-)?[1-9][0-9]*)$/;
```
nbbeeken, line 1032: Let's make this a global const

src/long.ts
```diff
@@ -1043,9 +1029,14 @@ export class Long {
...
+
+    if (doc.$numberLong.length > MAX_INT64_STRING_LENGTH) {
+      throw new BSONError('int64 string is too long');
+    }
 
-    if (!isInt64StrDecimalWithCorrectLength(doc.$numberLong)) {
-      throw new BSONError('Invalid int64 string');
+    if (!decimalRegEx.test(doc.$numberLong)) {
```
nbbeeken, line 1038: Let's add a test that "-0" is a failure

test/node/long.test.ts
```diff
@@ -21,4 +22,53 @@ describe('Long', function () {
...
+
+      expect(longRelaxedLegacy).to.deep.equal(longRelaxedNonLegacy);
+      expect(longCanonicalLegacy).to.deep.equal(longCanonicalNonLegacy);
+    });
+
+    describe('rejects with BSONError', function () {
+      it('hex strings', function () {
+        const ejsonDoc = { $numberLong: '0xffffffff' };
+        expect(() => Long.fromExtendedJSON(ejsonDoc)).to.throw(BSONError);
+      });
```
nbbeeken, line 48: one more case here for binary strings `0b`

test/node/long.test.ts
```diff
@@ -21,4 +22,53 @@ describe('Long', function () {
...
+
+      it('octal strings', function () {
+        const ejsonDoc = { $numberLong: '0o1234567' };
+        expect(() => Long.fromExtendedJSON(ejsonDoc)).to.throw(BSONError);
+      });
+
+      it('strings longer than 22 characters', function () {
+        const ejsonDoc = { $numberLong: '99999999999999999999999' };
+        expect(() => Long.fromExtendedJSON(ejsonDoc)).to.throw(BSONError);
+      });
```
nbbeeken, line 58: And we still want to get to a conclusion on checking if the string is 22 characters but is greater than / less than int64.max and min, respectively, we can add a tests with those input values here regardless

src/long.ts
```diff
@@ -1021,9 +1023,37 @@ export class Long {
...
   }
-  static fromExtendedJSON(doc: { $numberLong: string }, options?: EJSONOptions): number | Long {
-    const result = Long.fromString(doc.$numberLong);
-    return options && options.relaxed ? result.toNumber() : result;
+  static fromExtendedJSON(
+    doc: { $numberLong: string },
+    options?: EJSONOptions
+  ): number | Long | bigint {
+    const defaults = { useBigInt64: false, relaxed: true };
+    const ejsonOptions = { ...defaults, ...options };
```
nbbeeken, line 1031:
```suggestion
    const { useBigInt64 = false, relaxed = true } = { ...options };
```

Something to consider, removed the need for the temp defaults object

src/extended_json.ts
nbbeeken: @W-A-James bump on this

nbbeeken (changes requested):

src/long.ts
nbbeeken: Offline discussion: I proposed we don't use a helper anymore so we can offer more granular error messages

src/long.ts
```diff
@@ -1021,9 +1023,37 @@ export class Long {
...
+    const decimalRegEx = /^(\+?0|(\+|-)?[1-9][0-9]*)$/;
+
+    if (doc.$numberLong.length > MAX_INT64_STRING_LENGTH) {
+      throw new BSONError('int64 string is too long');
+    }
+
+    if (!decimalRegEx.test(doc.$numberLong)) {
+      throw new BSONError('int64 string is not in decimal');
```
nbbeeken, line 1039: Since we know the string is not huge at this point we can include it in the error message

nbbeeken (changes requested):

test/node/bigint.test.ts
nbbeeken: sure, let's file a follow up

src/long.ts
```diff
@@ -1021,9 +1025,35 @@ export class Long {
...
+    options?: EJSONOptions
+  ): number | Long | bigint {
+    const { useBigInt64 = false, relaxed = true } = { ...options };
+
+    if (doc.$numberLong.length > MAX_INT64_STRING_LENGTH) {
+      throw new BSONError('int64 string is too long');
+    }
+
+    if (!DECIMAL_REG_EX.test(doc.$numberLong)) {
+      throw new BSONError('int64 string is not a valid decimal integer');
```
nbbeeken, line 1039:
```suggestion
      throw new BSONError(`int64 string "${doc.$numberLong}" is not a valid decimal integer`);
```

Still think we should put the input in the error message, makes it searchable when someone needs to go looking for the bad input

src/long.ts
```diff
@@ -1021,9 +1025,35 @@ export class Long {
...
+    }
+
+    if (useBigInt64) {
+      const INT64_MAX = BigInt('0x7fffffffffffffff');
+      const INT64_MIN = -BigInt('0x8000000000000000');
+      const bigIntResult = BigInt(doc.$numberLong);
+      if (bigIntResult > INT64_MAX || bigIntResult < INT64_MIN) {
+        throw new BSONError('EJSON numberLong must be in int64 range');
+      }
```
nbbeeken, line 1048: Pending slack thread, will resolve when we come to a conclusion

nbbeeken (changes requested):

src/long.ts
W-A-James: Also going to do this in the `useBigInt64=true` case when the value provided is outside of int64 range.

test/node/bigint.test.ts
```diff
@@ -263,223 +263,225 @@ describe('BSON BigInt support', function () {
...
+    const useBigInt64IsSet = useBigInt64 ?? false;
+    const relaxedIsSet = relaxed ?? true;
+
+    const expectedResult = getExpectedResult(useBigInt64IsSet, relaxedIsSet);
+
+    return [{ options: { useBigInt64, relaxed }, expectedResult }];
+  }
+
+  function generateBehaviourDescription(entry: TestTableEntry, inputString: string): string {
+    // TODO(NODE-4874): When NODE-4873 is merged in, replace this with EJSON.stringify
```
nbbeeken, line 299: This TODO can be addressed now

nbbeeken (changes requested):

src/long.ts
W-A-James: Updating with results of Slack discussion;

Going to change bigint case to truncate rather than throw an error on encountering a bigint outside of int64 range to be consistent with behaviour of `Long.toString`

test/node/bigint.test.ts
W-A-James: Attempted to use `EJSON.stringify` to help construct test output, but it makes the test output harder to understand, so just going to remove the TODO

nbbeeken (approved):

PR #555, opened 2023-01-06
fix(NODE-4887): serializeInto does not check for the presence of a toBSON method for values in Map entries

W-A-James:
### Description

Adding fix to ensure that values of `Map`s run are replaced by the result of calling their `toBSON` method if it exists.

No

Ensure that `BSON.serialize` keeps with spec behaviour

### DIFF
--- a/test/node/to_bson_test.js
+++ b/test/node/to_bson_test.js
@@ -184,7 +184,7 @@ describe('toBSON', function () {
       expect(sizeNestedToBSON).to.equal(33);
     });
 
-    it('uses toBSON on elements of a Map', () => {
+    it('uses toBSON on values contained in a map', () => {
       const map = new Map();
       map.set('a', 100);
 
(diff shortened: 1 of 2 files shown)


test/node/to_bson_test.js
```diff
@@ -183,6 +183,15 @@ describe('toBSON', function () {
       const sizeNestedToBSON = BSON.calculateObjectSize({ a: [0] });
       expect(sizeNestedToBSON).to.equal(33);
     });
+
+    it('uses toBSON on elements of a Map', () => {
```
baileympearson, line 187:
```suggestion
    it('uses toBSON on values contained in a map', () => {
```

baileympearson (changes requested):
Looks good, one small test wording suggestion.  I'm requesting changes for visibility but otherwise LGTM

baileympearson (approved):

PR #556, opened 2023-01-11
test(NODE-4941): clean up double test organization

nbbeeken:
#### What is changing?

- Changes test names to communicate "class" of tests as opposed to literal test values
- Moves tests into context blocks for clarity
- Add a test for passing a string to the double constructor

##### What is the motivation for this change?

Some test cases seem to duplicate coverage when the names referred to a literal value (3 & 1234e7). We now clarify why we're testing each value as a distinct _use_ case of the Double class


test/node/double.test.ts
```diff
@@ -17,6 +17,11 @@ describe('BSON Double Precision', function () {
         // but the constructor at runtime should keep handling it correctly
         expect(new Double(new Number(value)).valueOf()).to.equal(value);
       });
+
+      it('stringified number', () => {
+        expect(new Double('1').valueOf()).to.equal(1);
+        expect(new Double('-0.0').valueOf()).to.equal(-0);
+      });
```
durran, line 24:
```suggestion
      context('when providing a stringified number', () => {
        it('sets the proper value', () => {
          expect(new Double('1').valueOf()).to.equal(1);
          expect(new Double('-0.0').valueOf()).to.equal(-0);
        });
      });
```

test/node/double.test.ts
```diff
@@ -193,6 +184,28 @@ describe('BSON Double Precision', function () {
...
+
+        // The following shows the when the string is an input the EJSON output still preserves the ending "28"
+        expect(new Double('1000000000000000128').toExtendedJSON({ relaxed: false })).to.deep.equal({
+          $numberDouble: '1000000000000000128.0'
+        });
+        // The same is true when the input is a JS number
+        expect(new Double(1000000000000000128).toExtendedJSON({ relaxed: false })).to.deep.equal({
+          $numberDouble: '1000000000000000128.0'
+        });
+      });
```
durran, line 208:
```suggestion
      context('when provided an integer beyond the precision of an 8-byte float', () => {
        // https://262.ecma-international.org/13.0/#sec-number.prototype.tofixed
        // Note: calling toString on this integer returns 1000000000000000100, so toFixed is more precise
        // This test asserts we do not change _current_ behavior, however preserving this value is not
        // something that is possible in BSON, if a future version of this library were to emit
        // "1000000000000000100.0" instead, it would not be incorrect from a BSON/MongoDB/Double precision perspective,
        //  it would just constrain the string output to what is possible with 8 bytes of floating point precision
        const integer = 1000000000000000128;
        const integerString = `${integer}`;

        it('passes equality when the number is beyond the presision', () => {
          expect(new Double(integerString).value).to.equal(1000000000000000128);
          expect(new Double(integerString).value).to.equal(1000000000000000100);
        });

        it('preserves the string value', () => {
          // The following shows the when the string is an input the EJSON output still preserves the ending "28"
          expect(new Double(integerString).toExtendedJSON({ relaxed: false })).to.deep.equal({
            $numberDouble: '1000000000000000128.0'
          });
          // The same is true when the input is a JS number
          expect(new Double(integer).toExtendedJSON({ relaxed: false })).to.deep.equal({
            $numberDouble: '1000000000000000128.0'
          });
        });
      });
```

durran (changes requested):

nbbeeken:
Great job getting the suggestion formatting right 😅 no prettier warnings from that commit

durran (dismissed):

durran:
@nbbeeken Can you update the description when you get a chance? Thanks!

nbbeeken:
@durran Done!

baileympearson (changes requested):
> Moves tests into context blocks for clarity

This bullet from the PR description is not addressed in this PR, let's wrap the assertions in the loop on line 155 in a context

test/node/double.test.ts
```diff
@@ -40,159 +47,174 @@ describe('BSON Double Precision', function () {
...
-            );
-          }
-        });
+      context('returns a stringified value that', () => {
+        for (const test of tests) {
+          const input = test.input;
+          const output = test.output;
+          const title = test.title;
+          it(title, () => {
+            const inputAsDouble = new Double(input);
```
baileympearson, line 161: This isn't exactly what I was looking for, I don't think I read the tests thoroughly enough on my first pass.  

The context should break apart the test classes, not group all test classes.  
The tests should not contain a "title" property, instead they should describe the class of test cases they're covering (i.e., a context) and that should be listed in the context blocks.  

Here's a before (with the current test structure) and after (for the first four test cases).  I think the latter is much easier to understand

Screen Shot 2023-01-13 at 11 31 43 AM
Screen Shot 2023-01-13 at 11 35 50 AM

baileympearson: Screen Shot 2023-01-13 at 11 38 46 AM
here's my quick code changes for reference

nbbeeken: "Stringifies to EJSON correctly" what's correctly mean in that context?

baileympearson: that was an example

if the title there is important, that could also be included in the tests (not my preference, why would we include one title but not the other) but another title could be generated

`stringifies 0 to { $numberDouble: 0.0 }` for example

nbbeeken: Sure, updated

baileympearson (approved):

durran (approved):

PR #558, opened 2023-01-30
test(NODE-4976): ensure all use of BigInt is gated by the useBigInt64 flag

W-A-James:
### Description

We would like js-bson to be usable in environments which do not support BigInt and so it is important to ensure that any use of BigInt features in the library should be gated by the `useBigInt64` flag.

#### What is changing?

- add new evergreen task to run test suite without access to BigInt APIs
- update tests in `test/node/bigint.test.ts` to be skipped when BigInt is not available
- updated `package.json` and `package-lock.json` to install eslint plugin 
- updated `.eslintrc.json` to enable eslint plugin to fail on the usage of BigInt literals and functions so that builds fail when BigInt features are used without explicitly adding an `eslint-disable` comment above the line using these features

No 


baileympearson (changes requested):
forgot to publish my changes yesterday

etc/eslint/no-bigint-usage/test/no_calls_to_BigInt_methods.test.js
```diff
@@ -0,0 +1,31 @@
...
+      errors: 1
+    },
+    {
+      name: 'Rejects calls BigInt static methods',
+      code: `
+      let x = BigInt.asIntN(64, 10n);
+      x = BigInt.asUintN(64, 10n);
+      x = BigInt.toString(10n);
+      x = BigInt.valueOf(10n);
+      x = BigInt.whatever();
```
baileympearson, line 26:
```suggestion
      x = BigInt.toLocaleString();
```

We may as well test all 5 static bigint methods if we're going to test 4.

baileympearson (changes requested):
Mostly minor comments, nice job.  CI looks good too

etc/eslint/no-bigint-usage/test/no_calls_to_BigInt_methods.test.js
W-A-James: Testing for this case would be somewhat different as `BigInt.toLocaleString` [is an instance method ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/toLocaleString) and we'd have to know beforehand that the value the method is being called is a bigint. 

This should be possible, but would require a refactor of the plugin to make use of ts-parser, which I could look more into.

package.json
```diff
@@ -86,10 +87,11 @@
   },
   "scripts": {
     "pretest": "npm run build",
-    "test": "npm run check:node && npm run check:web",
+    "test": "npm run check:node && npm run check:web && npm run check:web-no-bigint",
```
W-A-James, line 90: Only testing for web without bigint here as we expect that Node will always support BigInt. Open to changing this if anyone has other thoughts. (@nbbeeken, @addaleax)

W-A-James:
Summarizing a Slack conversation with @baileympearson:

We do not need a custom lint rule to prevent the usage of the `BigInt` constructor or its static methods as eslint has a `no-restricted-globals` rule which disallows any reference to globals provided to it in `.eslintrc.json`. This would prevent something like 
```javascript
const obj = BigInt;  // Unexpected use of 'BigInt' 	no-restricted-globals
obj.asUint64(64, obj(10));
```
by reporting an error on the assignment.

It is also not necessary to add linting rules on the instance methods of BigInt as running out test suite with BigInt and associated APIs set to null will flag any instances where we are using BigInt functionality that would negatively affect users in restricted environments. In addition, since we prevent calls to the BigInt constructor and use of bigint literals, we should not encounter the use of instance methods outside of APIs that expect to take bigints as inputs (for example serializing a user-provided object that has a bigint field) and so should not pose a problem.

For sake of completeness, the point about instance methods also applies to the use of `DataView`'s `(set|get)BigInt` methods. In the case of the former, it is only used in cases where we explicitly handle bigint input and for the latter, our restricted environment testing removes this method from the `DataView` class and so any unexpected use of it will be caught in CI.

package.json
baileympearson: The codepaths are largely the same (we mostly take a different path for `Buffer` related things).  I think this is sufficient.  I'm mainly concerned with the core serialization / deserialization / EJSON.stringification / EJSON.parsing, which is mostly the same across Node and web.

src/extended_json.ts
```diff
@@ -93,6 +93,7 @@ function deserializeValue(value: any, options: EJSONOptions = {}) {
       }
       if (in64BitRange) {
         if (options.useBigInt64) {
+          // eslint-disable-next-line no-restricted-globals
```
baileympearson, line 96: One comment, one question

- Let's add a comment explaining why we're disabling the rule here.  With the custom bigint plugin, it was obvious, but now that we're relying on `no-restricted-globals`, it's no longer obvious why the linter is disabled for these lines.
- Related to the above - can we configure `no-restricted-globals` to always require an additional message / description of why we're ignoring the rule?  It would be nice if we could enforce that when lint rules are skipped, the justification is explained.

baileympearson (changes requested):
small comment

src/extended_json.ts
W-A-James: From what I'm reading so far, the only way to do this would be to write a custom rule that checks contents of comments for the `eslint-disable` and `eslint-disable-next-line` directives and report when they don't contain what usually is an optional description.

Relevant documentation: https://eslint.org/docs/latest/use/configure/rules#comment-descriptions

baileympearson (approved):

PR #562, opened 2023-02-13
fix(NODE-5056): EJSON.parse date handling when useBigInt64=true

W-A-James:
### Description

`EJSON.parse` fails to correctly deserialize cEJSON strings which contain date fields.

#### What is changing?

- `EJSON.parse` now correctly deserializes cEJSON dates
- Added new `BSONError` subclass, `BSONRuntimeError` to cover internal failures

No

- [X] Ran `npm run lint` script

### DIFF
--- a/src/error.ts
+++ b/src/error.ts
@@ -58,3 +58,14 @@ export class BSONVersionError extends BSONError {
     );
   }
 }
+
+/** @public */
+export class BSONRuntimeError extends BSONError {
+  get name(): 'BSONRuntimeError' {
+    return 'BSONRuntimeError';
+  }
+
+  constructor(message: string) {
+    super(message);
+  }
+}
--- a/src/error.ts
+++ b/src/error.ts
@@ -6,7 +6,7 @@ import { BSON_MAJOR_VERSION } from './constants';
  *
  * `BSONError` objects are thrown when BSON ecounters an error.
  *
- * This is the parent calss for all the other errors thrown in this library.
+ * This is the parent class for all the other errors thrown by this library.
  */
 export class BSONError extends Error {
   /**
--- a/src/error.ts
+++ b/src/error.ts
@@ -4,7 +4,9 @@ import { BSON_MAJOR_VERSION } from './constants';
  * @public
  * @category Error
  *
- * `BSONError` objects are thrown when runtime errors occur.
+ * `BSONError` objects are thrown when BSON ecounters an error.
+ *
+ * This is the parent calss for all the other errors thrown in this library.
  */
 export class BSONError extends Error {
   /**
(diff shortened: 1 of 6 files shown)


src/extended_json.ts
```diff
@@ -125,10 +125,14 @@ function deserializeValue(value: any, options: EJSONOptions = {}) {
     if (options.legacy) {
       if (typeof d === 'number') date.setTime(d);
       else if (typeof d === 'string') date.setTime(Date.parse(d));
+      else if (typeof d === 'bigint') date.setTime(Number(d));
+      else throw new BSONRuntimeError('Unrecognized type for EJSON date');
```
nbbeeken, line 129:
```suggestion
      else throw new BSONRuntimeError(`Unrecognized type for EJSON date: ${typeof d}`);
```

src/error.ts
```diff
@@ -58,3 +58,14 @@ export class BSONVersionError extends BSONError {
     );
   }
 }
+
+/** @public */
```
nbbeeken, line 62:
```suggestion
/**
 * An error generated when BSON functions encounter an unexpected input
 * or reaches an unexpected/invalid internal state
 *
 * @public
 * @category Error
 */
```

And can we add ` * @category Error` to the other errors while we're here, a little docs++

nbbeeken (commented):
sorry jumping on this early just noticed some things

W-A-James:
Should we change the tsdoc comment for the `BSONError` class? At the moment it reads:

> `BSONError` objects are throw when runtime errors occur.

Having a subclass by the name of `BSONRuntimeError` seems confusing if that's what we have for the description of its parent class even if we specify that `BSONRuntimeError` specifically occurs when we have invalid internal state.

nbbeeken:
Sounds like a good improvement although there's still plenty of usage of it, someday we'll have descriptive errors as needed. We can make it more generic and omit the "runtime" bit. "thrown when BSON encounters an error" and then maybe an additional note about how it is the parent of all other errors we intentionally throw.

src/error.ts
```diff
@@ -2,7 +2,11 @@ import { BSON_MAJOR_VERSION } from './constants';
 
 /**
  * @public
- * `BSONError` objects are thrown when runtime errors occur.
+ * @category Error
+ *
+ * `BSONError` objects are thrown when BSON ecounters an error.
+ *
+ * This is the parent calss for all the other errors thrown in this library.
```
nbbeeken, line 9:
```suggestion
 * This is the parent class for all the other errors thrown by this library.
```

nbbeeken (changes requested):

nbbeeken (approved):

baileympearson (approved):

PR #569, opened 2023-03-31
feat(NODE-4855): add hex and base64 ctor methods to Binary and ObjectId

nbbeeken:

### DIFF
--- a/src/binary.ts
+++ b/src/binary.ts
@@ -258,6 +258,36 @@ export class Binary extends BSONValue {
     );
   }
 
+  /** Creates an Binary instance from a hex digit string */
+  static createFromHexString(hex: string, subType?: number): Binary {
+    const bytes = ByteUtils.fromHex(hex);
+    return new Binary(bytes, subType);
+  }
+
+  /** Creates an Binary instance from a base64 string */
+  static createFromBase64(base64: string, subType?: number): Binary {
+    const bytes = ByteUtils.fromBase64(base64);
+    return new Binary(bytes, subType);
+  }
+
+  /**
+   * Creates a new view over sequence and constructs a new Binary instance.
+   *
+   * @param sequence - A sequence of binary data
+   * @param subType - BSON Binary subtype, defaults to zero
+   * @param byteOffset - An offset to start the view from
+   * @param byteLength - A length to set the view to
+   */
+  static createFromBytes(
+    sequence: Uint8Array | ArrayBuffer | SharedArrayBuffer,
+    subType = 0,
+    byteOffset = 0,
+    byteLength = sequence.byteLength
+  ): Binary {
+    const bytes = ByteUtils.toLocalBufferType(sequence).subarray(byteOffset, byteLength);
+    return new Binary(bytes, subType);
+  }
+
   /** @internal */
   static fromExtendedJSON(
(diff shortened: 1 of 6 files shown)


src/binary.ts
```diff
@@ -292,7 +302,13 @@ export class Binary extends BSONValue {
   }
 
   inspect(): string {
-    return `new Binary(Buffer.from("${ByteUtils.toHex(this.buffer)}", "hex"), ${this.sub_type})`;
+    if (this.position === 0) {
```
nbbeeken, line 305: We had a bug here where the default `new Binary()` printed:

```ts
Binary(Buffer.from("00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000", "hex"), 0)
```

Which isn't the same as what would end up in BSON, b/c position determines how many bytes are serialized. The above constructs a Binary where position is set to 256, instead of set to zero.

src/objectid.ts
```diff
@@ -264,16 +264,22 @@ export class ObjectId extends BSONValue {
    * @param hexString - create a ObjectId from a passed in 24 character hexstring.
    */
   static createFromHexString(hexString: string): ObjectId {
-    // Throw an error if it's not a valid setup
-    if (typeof hexString === 'undefined' || (hexString != null && hexString.length !== 24)) {
-      throw new BSONError(
-        'Argument passed in must be a single String of 12 bytes or a string of 24 hex characters'
```
nbbeeken, line 270: Looks like a copy pasta mistake, the if stmt asserts for 24 characters, so a 12 byte string wouldn't work here.

test/node/binary.test.ts
```diff
@@ -0,0 +1,107 @@
+import { expect } from 'chai';
+import * as vm from 'node:vm';
+import { Binary, BSON } from '../register-bson';
+
+describe('class Binary', () => {
+  context('constructor()', () => {
+    it('creates an 256 byte Binary with subtype 0 by default', () => {
```
nbbeeken, line 7: Just testing the existing behavior here.

src/binary.ts
```diff
@@ -292,7 +302,13 @@ export class Binary extends BSONValue {
...
+      return this.sub_type !== Binary.BSON_BINARY_SUBTYPE_DEFAULT
+        ? `new Binary(undefined, ${this.sub_type})`
+        : 'new Binary()';
```
addaleax, line 308: I’m a bit surprised we’re printing this – is there a meaningful difference between an “empty” `Binary` object vs. a zero-length `Binary` object? I’d have expected `new Binary()` to result in something that inspects as `Binary.createFromBase64(""[, subtype])`

addaleax (approved):

src/binary.ts
nbbeeken: Good catch, yea as long as the position is set to zero then it doesn't matter if we've constructed the Binary with the default space or not. Better to be consistent :)

baileympearson (approved):

W-A-James (approved):

PR #572, opened 2023-04-20
fix(NODE-4960): UUID validation too strict

nbbeeken:
#### What is changing?

Our UUID class generates UUID v4 instances but it does not need to contain only v4. We lift the validation on the version byte entirely here to support all uuid versions and "empty" uuid.


src/binary.ts
```diff
@@ -485,6 +471,26 @@ export class UUID extends Binary {
     return new UUID(ByteUtils.fromBase64(base64));
   }
 
+  /** @internal */
+  static uuidBytesFromString(representation: string) {
```
baileympearson, line 475: The naming is redundant here since it's a static `UUID` method
```suggestion
  static bytesFromString(representation: string) {
```

src/binary.ts
```diff
@@ -311,42 +310,39 @@ export class Binary extends BSONValue {
...
 const UUID_BYTE_LENGTH = 16;
+const UUID_WITHOUT_DASHES = /^[0-F]{32}$/i;
+const UUID_WITH_DASHES = /^[0-F]{8}-[0-F]{4}-[0-F]{4}-[0-F]{4}-[0-F]{12}$/i;
 
 /**
  * A class representation of the BSON UUID type.
  * @public
  */
 export class UUID extends Binary {
-  static cacheHexString: boolean;
```
baileympearson, line 321: Can this safely be removed?  It's a public property

src/binary.ts
nbbeeken: SGTM

test/node/tools/utils.js
```diff
@@ -128,6 +128,14 @@ const bufferFromHexArray = array => {
 
 exports.bufferFromHexArray = bufferFromHexArray;
 
+function int32ToHex(int32) {
```
baileympearson, line 131: the naming here is a bit misleading, especially without TS.  Since we have a dedicated `Int32` class, I expected this method to take an instance of Int32.  Can we clarify somehow that this is a number? 

One option:

```typescript
function numberToHex(size: 'i32' | 'i64', n) { 
...
}
```

test/node/uuid.test.ts
```diff
@@ -62,51 +51,23 @@ describe('UUID', () => {
...
     const org = new UUID();
     const copy = new UUID(org);
     expect(org.id).to.not.equal(copy.id);
     expect(org.id).to.deep.equal(copy.id);
   });
 
-  /**
-   * @ignore
-   */
-  it('should throw if passed invalid 36-char uuid hex string', () => {
```
baileympearson, line 78: instead of deleting this test - do we instead want to modify it (/add a new test) that demonstrates that we do not enforce the UUID type?

test/node/uuid.test.ts
```diff
@@ -62,51 +51,23 @@ describe('UUID', () => {
...
-   */
   it('should correctly check if a buffer isValid', () => {
     const validBuffer = Buffer.from(UPPERCASE_VALUES_ONLY_UUID_STRING, 'hex');
-    const invalidBuffer1 = Buffer.from('aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa', 'hex');
-    const invalidBuffer2 = Buffer.alloc(16);
-
-    expect(validBuffer.length).to.equal(invalidBuffer1.length);
-    expect(validBuffer.length).to.equal(invalidBuffer2.length);
-    expect(UUID.isValid(invalidBuffer1)).to.be.false;
-    expect(UUID.isValid(invalidBuffer2)).to.be.false;
```
baileympearson, line 103: we're missing a lot of coverage here (we're now only testing the success case).  Should we add tests for the expected failure cases?  

I'm asking because you specifically removed the only two error tests we had.

baileympearson (changes requested):

test/node/tools/utils.js
nbbeeken: I added documentation here, the method can accept an Int32 already b/c of type coercion

src/binary.ts
nbbeeken: Good catch, added it back and filed follow ups, we can deprecate it in the subtask I made and remove in v6

test/node/uuid.test.ts
nbbeeken: Added isValid coverage, fixed the instanceof usage

src/binary.ts
```diff
@@ -311,42 +310,40 @@ export class Binary extends BSONValue {
 export type UUIDExtended = {
   $uuid: string;
 };
+
 const UUID_BYTE_LENGTH = 16;
+const UUID_WITHOUT_DASHES = /^[0-F]{32}$/i;
+const UUID_WITH_DASHES = /^[0-F]{8}-[0-F]{4}-[0-F]{4}-[0-F]{4}-[0-F]{12}$/i;
```
addaleax, line 316: Keep in mind that `[0-F]` and `[0-9A-F]` are different ranges :)

src/binary.ts
```diff
@@ -311,42 +310,40 @@ export class Binary extends BSONValue {
...
    */
   constructor(input?: string | Uint8Array | UUID) {
     let bytes: Uint8Array;
-    let hexStr;
     if (input == null) {
       bytes = UUID.generate();
     } else if (input instanceof UUID) {
       bytes = ByteUtils.toLocalBufferType(new Uint8Array(input.buffer));
-      hexStr = input.__id;
     } else if (ArrayBuffer.isView(input) && input.byteLength === UUID_BYTE_LENGTH) {
```
addaleax, line 337: I brought it up in the ticket but this can still lead to validation issues:

```js
const { Binary, serialize, deserialize } = require('bson');

const bin = Binary.createFromHexString('aaaaaa', 4);
const s = serialize({ bin });
const d = deserialize(s); // throws
```

and of course this can also happen when reading data from a real server. (My suggestion would be to skip this [branch here](https://github.com/mongodb/js-bson/blob/91b84ed72c1f1f5f882d993c2fec9a4aea9740a9/src/parser/deserializer.ts#L407-L408) for invalid UUIDs but I don’t have strong feelings about that.)

If that’s not in scope for this ticket, I’d make sure to create another one.

addaleax (changes requested):

src/binary.ts
nbbeeken: Good catch, I was only thinking about 16 byte sequences we were too strictly validating. I think we should make this change, the spirit of the ticket is to handle values generated outside this library more gracefully.

addaleax (approved):

src/parser/deserializer.ts
```diff
@@ -433,7 +433,8 @@ function deserializeObject(
         if (promoteBuffers && promoteValues) {
           value = _buffer;
         } else if (subType === constants.BSON_BINARY_SUBTYPE_UUID_NEW) {
-          value = new Binary(buffer.slice(index, index + binarySize), subType).toUUID();
+          const binary = new Binary(buffer.slice(index, index + binarySize), subType);
+          value = UUID.isValid(binary) ? binary.toUUID() : binary;
         } else {
           value = new Binary(buffer.slice(index, index + binarySize), subType);
         }
```
baileympearson, line 440:
```suggestion
          value = new Binary(buffer.slice(index, index + binarySize), subType);
          if (subType === constants.BSON_BINARY_SUBTYPE_UUID_NEW && UUID.isValid(value)) {
                      value = value.toUUID();
          }
```
can we align the logic between deserialize and deserializeObject?

nbbeeken: good cleanup, fixed

baileympearson (approved):

PR #578, opened 2023-05-31
feat(NODE-4938): improve react native bundle experience

durran:
### Description

Creates a new CJS bundle to be used with React Native.

#### What is changing?
- Vendors the text-encoding and base64 modules.
- Removes modifying the global object from the vendored modules.
- Creates a new bson.rn.cjs artefact that imports the vendored functions for `TextEncoder`, `TextDecoder`, `btoa` and `atob`.

None

NODE-4938


vendor/text-encoding/index.js
```diff
@@ -0,0 +1,9 @@
+// This is free and unencumbered software released into the public domain.
+// See LICENSE.md for more information.
+
+var encoding = require("./lib/encoding.js");
```
baileympearson, line 4: Where will `encoding.js` be loaded from?  Should that be included in the bundle?

durran: It's loaded from the path as the vendor directory is part of the published module. They are not part of the bson.rn.cjs artefact, hence it adds the following lines:

```
const { TextEncoder, TextDecoder } = require('../vendor/text-encoding');
const { encode: btoa, decode: atob } = require('../vendor/base64');
```

baileympearson: But we need to include an `encoding.js` file somewhere, right?  sorry for these dumb questions:  

Where is this file?  Is it generated?  I don't see it in our existing bson source, or generated `lib`, or in this PR, and neither library was added as a dependency so it won't come from there.

also, failing tests: https://spruce.mongodb.com/task/node_bson_linux_no_bigint_web_tests_patch_259547d9e926c87f61f664a68c2a3a9172f53d2c_647790670305b946d4e34eee_23_05_31_18_22_32/logs?execution=0

baileympearson: I see it's included in the release test - it might be an accident that it didn't make it into the PR?

durran: Yeah I seemed to have missed it in the PR.

baileympearson (dismissed):

nbbeeken (dismissed):

README.md
```diff
@@ -188,23 +188,17 @@ try {
 
 ## React Native
 
-BSON requires that `TextEncoder`, `TextDecoder`, `atob`, `btoa`, and `crypto.getRandomValues` are available globally.  These are present in most Javascript runtimes but require polyfilling in React Native.  Polyfills for the missing functionality can be installed with the following command:
+BSON vendors the required polyfills for `TextEncoder`, `TextDecoder`, `atob`, `btoa` when using React Native. One
```
kraenhansen, line 191: I suspect the term "vendors" might not be clear enough for publicly facing documentation. Perhaps it would make sense to include an explanation on what that means for the user.

```suggestion
BSON vendors the required polyfills for `TextEncoder`, `TextDecoder`, `atob`, `btoa` imported from React Native and therefore doesn't expect users to polyfill these. One
```

README.md
```diff
@@ -188,23 +188,17 @@ try {
...
+additional polyfill, `crypto.getRandomValues` is required and can be installed with the following command:
```
kraenhansen, line 192: > is required

Isn't this optional?
```suggestion
additional polyfill, `crypto.getRandomValues` is recommended and can be installed with the following command:
```

kraenhansen (approved):

etc/rollup/rollup-plugin-require-vendor/require_vendor.mjs
```diff
@@ -0,0 +1,30 @@
+import MagicString from 'magic-string';
+
+const REQUIRE_TEXT_ENCODING =
```
kraenhansen, line 3: This name seam to suggest only one of the polyfills are provided here.

kraenhansen:
BTW: I've built the PR and tested it in a React Native app 👍

README.md
durran: Updated.

README.md
durran: Updated.

durran:
> An alternative I don't know if you've considered: If you had a separate input for the RN bundle (which would setup the polyfills and re-export `src/index.ts`), you could most likely avoid the code injection / `RequireVendor` all together. It would require that the runtime independent code stored the polyfilled APIs on an object that the wrapping input file would then inject its functions into before importing `src/index.ts`. We're [already using this pattern in Realm JS](https://github.com/realm/realm-js/blob/ca790bedfb5d6d1ecd1f557ee951c0fa5175b142/packages/realm/src/platform/react-native/device-info.ts#L61-L80).
> 
> If you had a separate input file for the RN bundle, it could make sense (not necessarily for this PR) to move [this check and warning](https://github.com/mongodb/js-bson/blob/main/src/utils/web_byte_utils.ts#L58-L63) into the RN bundle to avoid relying on [checks of the runtime](https://github.com/mongodb/js-bson/blob/main/src/utils/web_byte_utils.ts#L31-L34) (as that tends to be brittle in my experience).
> 
> BTW: I've built the PR and tested ObjectId and UUID in a React Native app 👍

I think that may be something we can discuss and potentially ticket in the future.

nbbeeken (approved):

baileympearson (approved):

PR #579, opened 2023-05-31
chore(NODE-5324): update to typescript 5

durran:
### Description

Updates the driver to use Typescript 5.

#### What is changing?
- Updates typescript to ^5.0.4
- Updates api-extractor to ^7.35.1 (to use typescript 5)
- All other typescript dependencies were already further along than the minimum.
- Removes the `importsNotUsedAsValues` config and replaces with eslint rule.
- Updates all the files to adhere to new eslint rule.
- Fixes the number comparison between `number` and `Int32`.

None

NODE-5324

### DIFF
--- a/src/timestamp.ts
+++ b/src/timestamp.ts
@@ -61,24 +61,26 @@ export class Timestamp extends LongWithoutOverridesClass {
       if (typeof low.i !== 'number' && (typeof low.i !== 'object' || low.i._bsontype !== 'Int32')) {
         throw new BSONError('Timestamp constructed from { t, i } must provide i as a number');
       }
-      if (low.t.valueOf() < 0) {
+      const t = Number(low.t);
+      const i = Number(low.i);
+      if (t < 0 || Number.isNaN(t)) {
         throw new BSONError('Timestamp constructed from { t, i } must provide a positive t');
       }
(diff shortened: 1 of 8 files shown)


src/timestamp.ts
```diff
@@ -61,18 +61,18 @@ export class Timestamp extends LongWithoutOverridesClass {
       if (typeof low.i !== 'number' && (typeof low.i !== 'object' || low.i._bsontype !== 'Int32')) {
         throw new BSONError('Timestamp constructed from { t, i } must provide i as a number');
       }
-      if (low.t < 0) {
+      if (low.t.valueOf() < 0) {
```
nbbeeken, line 64: I do wonder about explicitly calling `.valueOf()`. It is correct, but should we instead respect the type coercion built into JS and use a `Number()` cast? There's a ticket to add [Symbol.toPrimitive](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toPrimitive) to our BSON types, we would likely want that to be automatically respected here, is what I'm thinking. 

We would have to confirm that `t` and `i` are not `NaN`

nbbeeken (commented):
lgtm, question about valueOf

nbbeeken (approved):

PR #581, opened 2023-06-06
rename init-nvm.sh

baileympearson:


.evergreen/install-dependencies.sh
```diff
@@ -90,20 +90,18 @@ else
...
-hash -r
-
-# Set npm -g prefix to our local artifacts directory
-cat <<EOT > .npmrc
-prefix=$NODE_ARTIFACTS_PATH/npm_global
-EOT
+if [[ -z "${npm_global_prefix}" ]]; then
+  echo "npm_prefix_config not set"
+  exit 1
+fi
```
baileympearson, line 96: I'm going to port this to the driver.  

I realized that without checking that the correct env variable is set, we could easily silently go back to installing `npm` globally.

nbbeeken: We also rely on NODE_ARTIFACTS_PATH in this script, should we error check both below the sourcing of init-node-env?

baileympearson: Done.

nbbeeken (changes requested):

nbbeeken (approved):

PR #585, opened 2023-06-22
fix(NODE-5363): defer byte slicing to utf8 decoding API in nodejs

W-A-James:
##### `ByteUtils`
- Updated signature of `ByteUtils.toUTF8`
	- `toUTF8: (buffer: Uint8Array) => string` -> `toUTF8: (buffer: Uint8Array, start: number, end: number) => string`
- Updated node implementation of `ByteUtils.toUTF8` to take `start` and `end` parameters and pass them into `Buffer.toString` directly
- Updated node implementation of `ByteUtils.toUTF8` to take `start` and `end` parameters and pass them into `Uint8Array.slice`
- Updated node implementation of `ByteUtils.toUTF8` to not make a call to `toLocalBufferType`

##### `deserializer`
- Removed unnecessary call to `Buffer.subarray` which was one source of performance degradation

##### Benchmarks
- Added int32 and Long deserialization benchmarks
##### Performance

This PR is meant to address a performance regression that was introduced in v4.6.0 during the introduction of our `ByteUtils` functions.

The benchmark used to assess performance is the `deserialize a large batch of documents each with an array of many strings` benchmark in `etc/benchmarks/main.js`. The benchmark was run 100 times on v4.6, v5.3(latest as at time of writing) and with the changes made in this PR.

The chart below shows the percent increase in runtime (lower is better) over v4.6. It clearly shows the difference in performance between v5.3 and v4.6 with the `local` column displaying marked improvement over v5.3. Note that there are still performance gains to be made before we are once again on par with v4.6.

NODE-5458-performance

No

Increasing bson and driver performance


src/utils/node_byte_utils.ts
```diff
@@ -125,8 +125,8 @@ export const nodeJsByteUtils = {
     return Buffer.from(text, 'utf8');
   },
 
-  toUTF8(buffer: Uint8Array): string {
-    return nodeJsByteUtils.toLocalBufferType(buffer).toString('utf8');
+  toUTF8(buffer: Uint8Array, start?: number, end?: number): string {
+    return (buffer as NodeJsBuffer).toString('utf8', start, end);
```
W-A-James, line 129: I don't like the cast here, but I'm not sure how to go about changing this in a way that wouldn't require a large refactoring of the entire `ByteUtils` interface.

Running the tests, it's apparent that wherever `ByteUtils.toUTF8` is called that the buffer passed in is already a `NodeJsBuffer`, so it begs the question: where else we should be able to get rid of calls to `toLocalBufferType` and how much more speedup we can get as a result?

billouboq:
Hi @W-A-James,

Is there any way we can help you on those performance changes ? 
Like is there any other things to look for ?

src/utils/node_byte_utils.ts
addaleax: Yeah, I don’t like the cast either, it’s making me a bit that this would just silently give the wrong result if `buffer` is a plain `Uint8Array` (like the method signature asks for!).

> that wouldn't require a large refactoring of the entire `ByteUtils` interface

Honestly, I think this might be the way to go, even if it’s inconvenient.

```js
export type ByteUtils<LBT extends Uint8Array> = {
  toLocalBufferType(buffer: Uint8Array | ArrayBufferView | ArrayBuffer): LBT;
  fromUTF8: (text: string) => LBT;
  toUTF8: (buffer: LBT) => string;
  /* ... */
};

export const webByteUtils: ByteUtils<Uint8Array> = {
  toLocalBufferType(buffer: Uint8Array | ArrayBufferView | ArrayBuffer): Uint8Array;
  fromUTF8: (text: string) => Uint8Array;
  toUTF8: (buffer: Uint8Array) => string;
  /* ... */
};

export const nodeJsByteUtils: ByteUtils<Buffer> = {
  toLocalBufferType(buffer: Uint8Array | ArrayBufferView | ArrayBuffer): Buffer;
  fromUTF8: (text: string) => Buffer;
  toUTF8: (buffer: Buffer) => string;
  /* ... */
};
```

durran: I'm also concerned about the cast here and think it's probably better to go with something like what Anna has recommended here just to ensure we've got proper typing in place. It's an internal class so we should feel free to update as needed.

durran (changes requested):

W-A-James:
> Is there any way we can help you on those performance changes ?
> Like is there any other things to look for ?

Hi @billouboq, thanks for your interest in this work! The work in this PR specifically is focused on string deserialization, but we are interested in investigating more about bson performance in general. We would be interested in investigating:

  1. string serialization performance
  2. ObjectId serialization and deserialization performance
  3. Int32 serialization and deserialization performance
  4. Double serialization and deserialization performance
  5. Long serialization and deserialization performance
  5. Nested document serialization and deserialization performance

  Any investigation into any one of the areas listed above would be very useful. If you identify any new areas that could be improved and have suggestions on how to do so, filing a ticket on the [NODE project](https://jira.mongodb.org/projects/NODE) with your findings would be the best way to inform us about your findings and allow us to take action on them.

  As for the benchmarks we currently have, they live in `etc/benchmarks/main.mjs`. We currently have benchmarks for
  - string deserialization (see the 'deserialized a large batch of documents each with an array of many
    strings' benchmark)
  - Double deserialization (see the 'Many Doubles deserialization' benchmark)

  This pr adds similar benchmarks for Int32 and Long deserialization.
  The workflow (which itself could use improvement) for running these benchmarks is as follows:

  1. Set the `skip` flag to false on the benchmarks you are interested in running
  2. Set the iteration count variable in etc/benchmarks/main.mjs to an appropriate value (we default
     to 100)
  3. Check etc/benchmarks/bson_versions.json to ensure that the bson versions you are interested in
     comparing are present. If only interested in testing a local copy, then you can remove the
     entries from the array.
  4. Run ./etc/benchmarks/install_bson_versions.sh
  5. Run `node etc/benchmarks/main.mjs | tee <results file>` to put the statistics collected into a file. Note that this will also emit cpu profiles for each of the version-test pairs specified
  6. Use etc/benchmarks/convert_to_csv.sh to convert the results to a csv

src/utils/byte_utils.ts
```diff
@@ -26,7 +26,7 @@ export type ByteUtils = {
   /** Create a Uint8Array containing utf8 code units from a string */
   fromUTF8: (text: string) => Uint8Array;
   /** Create a string from utf8 code units */
-  toUTF8: (buffer: Uint8Array) => string;
+  toUTF8: (buffer: Uint8Array, start?: number, end?: number) => string;
```
nbbeeken, line 29: Can we make start and end required? I think there's only a few usages that would need updating.

nbbeeken (changes requested):

src/utils/node_byte_utils.ts
W-A-James: I discussed this with @nbbeeken and we came to the conclusion that a refactor on the scale that would be necessary to get this to work would be out of scope for this ticket.

I think it would be better to leave the call to `toLocalBufferType` in `toUTF8` and address that in a later PR. For now, the change that this PR will centre around is the removal of unnecessary `Buffer.subarray` calls at `ByteUtils.toUTF8` callsites. I will update the PR description and ticket description and post updated performance results with the new changes.

src/binary.ts
```diff
@@ -223,8 +223,8 @@ export class Binary extends BSONValue {
   toString(encoding?: 'hex' | 'base64' | 'utf8' | 'utf-8'): string {
     if (encoding === 'hex') return ByteUtils.toHex(this.buffer);
     if (encoding === 'base64') return ByteUtils.toBase64(this.buffer);
-    if (encoding === 'utf8' || encoding === 'utf-8') return ByteUtils.toUTF8(this.buffer);
-    return ByteUtils.toUTF8(this.buffer);
+    if (encoding === 'utf8' || encoding === 'utf-8') return ByteUtils.toUTF8(this.buffer, 0, this.buffer.length);
+    return ByteUtils.toUTF8(this.buffer, 0, this.buffer.length);
```
nbbeeken:
```suggestion
    if (encoding === 'utf8' || encoding === 'utf-8') return ByteUtils.toUTF8(this.buffer, 0, this.buffer.byteLength);
    return ByteUtils.toUTF8(this.buffer, 0, this.buffer.byteLength);
```

nit: length and byteLength are equivalent when working with Uint8Arrays but I appreciate the "in code" reminder that this is byte-wise length we're slicing on.

nbbeeken (changes requested):

durran (approved):

nbbeeken (approved):

PR #593, opened 2023-07-10
ci(NODE-5403): add string deserialization benchmark to ci

W-A-James:
##### New Benchmark infrastructure
- Added benchmark infrastructure under `test/bench` with suites living under `test/bench/suites`

##### Migrate String deserialization benchmark
- String deserialization benchmark added in `test/bench/string_deserialization.ts`
	- Checks with utf8 validation disabled and enabled

##### New CI tasks
- Added new ci tasks for running benchmarks using web and node versions of bson
- Tasks run on rhel 9.0 perf distribution

No

NODE-5403


.evergreen/config.yml
```diff
@@ -200,6 +208,30 @@ tasks:
         vars:
           TS_VERSION: "next"
           TRY_COMPILING_LIBRARY: "false"
+  - name: run-benchmarks-node
+    commands:
+      - func: fetch source
+        vars:
+          NODE_LTS_VERSION: 16
```
W-A-James, line 215: Need to pin this to 16.20.1 specifically

package.json
```diff
@@ -103,6 +103,7 @@
     "check:tsd": "npm run build:dts && tsd",
     "check:web": "WEB=true mocha test/node",
     "check:web-no-bigint": "WEB=true NO_BIGINT=true mocha test/node",
+    "check:bench":"npx ts-node test/bench/index.ts",
```
W-A-James, line 106: Should we compile it to javascript and run that directly or is this fine?

test/bench/index.ts
```diff
@@ -0,0 +1,33 @@
...
+console.log(
+  [
+    `\n- cpu: ${platform.name}`,
+    `- cores: ${platform.cores}`,
+    `- os: ${process.platform}`,
+    `- ram: ${platform.ram}`
+  ].join('\n')
+);
+
+for (const suite of [getStringDeserializationSuite(BSON)]) {
```
W-A-James, line 21: This is meant to collect all the suites we may end up adding to the `suites` directory

.evergreen/config.yml
W-A-James: Pinned

package.json
nbbeeken: We should compile to JS, the overhead ts-node introduces is unknown and I don't think reflects real typical usage. 
Maybe a shell script or &&-ed commands to compile / clean files?

test/bench/suites/string_deserialization.ts
```diff
@@ -18,7 +18,11 @@ export function getStringDeserializationSuite(BSON): Suite {
       data: DOC,
       fn: serializedDoc =>
         BSON.deserialize(serializedDoc, { validation: { utf8: utf8Validation } }),
-      iterations: 1000
+      iterations: 1000,
```
nbbeeken, line 21: I think 1000 is quite small. I'm not sure what is a good value to raise it to, probably north of 100,000.

test/bench/index.ts
nbbeeken: It will be important for us to isolate these micro bench marks in individual nodejs processes, since this is the only one this is fine, can you leave a comment about that? Like "before adding more benches, either comment out the existing one or setup child process logic".

.evergreen/config.yml
```diff
@@ -93,7 +93,15 @@ functions:
...
         args:
           - .evergreen/run-bundling-test.sh
-
+  run benchmarks:
+    - command: shell.exec
+      type: test
+      params:
+        working_dir: src
+        script: |
+          ${PREPARE_SHELL}
```
nbbeeken, line 102: I'm not seeing a PREPARE_SHELL expansion, I think this is copy pasta from the driver. Can we use subprocess.exec like "run bundling" does?

.evergreen/config.yml
```diff
@@ -200,6 +208,30 @@ tasks:
         vars:
           TS_VERSION: "next"
           TRY_COMPILING_LIBRARY: "false"
+  - name: run-benchmarks-node
+    commands:
+      - func: fetch source
+        vars:
+          NODE_LTS_VERSION: v16.20.1
```
nbbeeken, line 215: Our perf nodejs version is v18.16.0 in the driver, should we use the same here?

.evergreen/run-benchmarks.sh
```diff
@@ -0,0 +1,20 @@
...
+
+source "${PROJECT_DIRECTORY}/.evergreen/init-node-and-npm-env.sh"
+
+case $1 in
+  "node")
+    export WEB=false
+    npm run check:bench
+    ;;
+  "web")
+    export WEB=true
```
nbbeeken, line 11: Per our slack convo we're going to skip testing the WEB setting for now. 

But for structuring this script, could we just let the evg config set WEB to true/false? just so there isn't multple ways to achieve the same thing

test/bench/util.ts
```diff
@@ -0,0 +1,29 @@
...
+  benchmarkName: string,
+  metrics: {
+    name: string;
+    results: number[];
+  }[],
+  args?: Record<string, string>
+): PerfSendData {
+  return {
+    info: {
+      test_name: benchmarkName.replace(' ', '_'),
```
nbbeeken, line 20: You want to replace all spaces right?

Docs on `replace()`:
> A string pattern will only be replaced once. To perform a global search and replace, use a regular expression with the g flag, or use replaceAll() instead.

test/bench/task.ts
```diff
@@ -0,0 +1,62 @@
...
+    // Warmup
+    for (let i = 0; i < this.iterations; i++) {
+      this.fn(this.data);
+    }
+    const results: number[] = [];
+
+    for (let i = 0; i < this.iterations; i++) {
+      const start = performance.now();
+      this.fn(this.data);
+      const end = performance.now();
```
nbbeeken, line 46:
```suggestion
      const { data } = this
      const start = performance.now();
      this.fn(data);
      const end = performance.now();
```

One less step that is part of the measured code path. Probably want to pull data out at the top of run so the warm up is invoked the same way too

nbbeeken (changes requested):

.evergreen/config.yml
W-A-James: Not sure why I thought we were using 16 for the perf distribution. Thanks for catching that.

test/bench/suites/string_deserialization.ts
W-A-James: Raising it to 100,000 makes evergreen time out. Do we want to extend the timeout duration or reduce the number of iterations?

test/bench/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "devDependencies": {
+    "@types/node": "^18.16.3",
+    "typescript": "^5.0.4",
+    "eslint": "^8.39.0",
+    "prettier": "^2.8.8",
+    "eslint-config-prettier": "^8.8.0",
+    "eslint-plugin-prettier": "^4.2.1"
+  },
+  "dependencies": {
+    "bson": "file:../.."
```
nbbeeken, line 23: from team chat, let's try and reuse the set up in this repo for compiling and linting

test/bench/src/suites/string_deserialization.ts
```diff
@@ -0,0 +1,31 @@
...
+export function getStringDeserializationSuite(): Suite {
+  const DOC: Uint8Array = BSON.serialize({
+    nextBatch: Array.from({ length: 100 }, () => {
+      return {
+        _id: new BSON.ObjectId(),
+        arrayField: Array.from({ length: 20 }, (_, i) => `5e99f3f5d3ab06936d36000${i}`)
+      };
+    })
+  });
+  const suite = new Suite('string seserialization');
```
nbbeeken, line 13:
```suggestion
  const suite = new Suite('string serialization');
```

nbbeeken (changes requested):

test/bench/tsconfig.json
```diff
@@ -0,0 +1,27 @@
+{
```
nbbeeken, line 1: We can delete this now, since we're running `tsc` with no args its using the root config.

.evergreen/config.yml
```diff
@@ -56,7 +55,6 @@ functions:
       params:
         working_dir: src
         script: |
-          ${PREPARE_SHELL}
           echo "TEST_TARGET=${TEST_TARGET}"
```
nbbeeken, line 58: you can either leave prepare_shell here, or convert to subprocess.exec, the lint is failing b/c PROJECT_DIRECTORY is missing, which would be added to the env by PREPARE_SHELL or we can use `add_expansions_to_env`

nbbeeken (changes requested):

test/bench/tsconfig.json
W-A-James: We run currently run tsc using this config. We'd have to add the 'node' types into the root config if we want to be able to have it use console and other node APIs

nbbeeken (approved):

test/bench/tsconfig.json
nbbeeken: good call, okay sounds good to leave as is

test/bench/suites/string_deserialization.ts
nbbeeken: Let's follow up when we have more benchmarks, I don't think we have an accurate data point on where JS will turn this code into machine instructions to decide here.

PR #601, opened 2023-07-26
feat(NODE-4770)!: remove 12 length string support from ObjectId constructor

nbbeeken:
#### What is changing?

- Removing the ability to pass a string that reports a length of 12 from creating an ObjectId
- Updating the .equals method to correspond

The usage was never reported by the API prior and worked inconsistently, 12 length strings interpreted as utf8 /  iso88591 interchangeably. I think this release note serves as enough documentation. 

See notes, and doc section.

### Strings of length 12 can no longer make an ObjectId

> (From [String.length](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/length)): [The String length] property returns the number of code units in the string. JavaScript uses [UTF-16](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String#utf-16_characters_unicode_code_points_and_grapheme_clusters) encoding, where each Unicode character may be encoded as one or two code units, so it's possible for the value returned by length to not match the actual number of Unicode characters in the string.

The `ObjectId` constructor erroneously interpreted a string with length of 12 as UTF8 bytes that could be converted to an ObjectId. This is unexpected for at least two reasons. The first is that a legacy approach (pre- [Uint8Arrays](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array)) to handling binary data was to pass around "[binary strings](https://developer.mozilla.org/en-US/docs/Web/API/btoa)", where each character represents a single byte, this is not the same as interpreting a sting as UTF8, which has restrictions on how each byte can be formatted. The second is that a string of length 12 does not result in 12 bytes of data when converted to utf8 (ex. `'🐶🐶🐶🐶🐶🐶'.length === 12`, but as UTF8 bytes this is a 24-byte sequence).

Despite the bugginess of the behavior discussed above, the right string in the right context does create the proper byte sequence, so we are considering this a breaking change and removing it in this major release.

### DIFF
--- a/test/node/object_id.test.ts
+++ b/test/node/object_id.test.ts
@@ -116,13 +116,8 @@ describe('ObjectId', function () {
     const objectValidString24Hex = {
       id: 'aaaaaaaaaaaaaaaaaaaaaaaa'
     };
-    const objectValidString12Bytes = {
-      id: 'abcdefghijkl'
-    };
     const buf24Hex = Buffer.from('aaaaaaaaaaaaaaaaaaaaaaaa', 'hex');
-    const buf12Bytes = Buffer.from('abcdefghijkl');
     expect(new ObjectId(objectValidString24Hex).id).to.deep.equal(buf24Hex);
-    expect(new ObjectId(objectValidString12Bytes).id).to.deep.equal(buf12Bytes);
   });
 
   it('should correctly create ObjectId from object with valid string id and toHexString method', function () {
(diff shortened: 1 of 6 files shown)


test/node/object_id.test.ts
```diff
@@ -336,22 +331,9 @@ describe('ObjectId', function () {
...
-      'Argument passed in must be a string of 12 bytes'
-    );
-    expect(() => new ObjectId(length12Not12Bytes).id).to.throw(
-      BSONError,
-      'Argument passed in must be a string of 12 bytes'
-    );
-  });
-
-  it('should have isValid be true for 12-char length and 12-byte length string', function () {
+  it('should have isValid be false for 12-char length and 12-byte length string', function () {
```
W-A-James, line 334: We should break out the `isValid` test strings with 2-byte unicode code points separate from this test

W-A-James (changes requested):

W-A-James:
Also, if you could put the dependency changes in the ticket AC, that would be great.

test/node/object_id.test.ts
nbbeeken: I organized the isValid tests per input type and covered relevant cases

W-A-James (approved):

PR #605, opened 2023-08-02
feat(NODE-5504): bump bson major version

durran:
### Description

Bumps `BSON_MAJOR_VERSION` constant to 6.

#### What is changing?

Bumps `BSON_MAJOR_VERSION` constant to 6 and adds a test.

None

NODE-5504

#### `BSON_MAJOR_VERSION` bumped to 6

Only BSON objects that have this major version can be serialized with this version of the library. Mismatched objects will throw a `BSONVersionError` when attempting to serialize.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -1,5 +1,5 @@
 /** @internal */
-export const BSON_MAJOR_VERSION = 5 as const;
+export const BSON_MAJOR_VERSION = 6 as const;
 
 /** @internal */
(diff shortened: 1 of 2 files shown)


W-A-James (changes requested):
You're getting failures in CI because we have tests for what seems to be all of our BSON types that check for the `@@mdb.bson.version` flag that are failing to match the new value.

W-A-James (approved):

PR #606, opened 2023-08-03
feat(NODE-2624)!: remove Long methods from Timestamp class

W-A-James:
##### `Timestamp`
- `Timestamp` no longer subclasses from `Long` and instead is now a direct subclass of `BSONValue`
- Deleted `LongWithoutOverrides` and `LongWithoutOverridesClass`
- Updated `MAX_VALUE` to be instance of `Timestamp`
- Expose `i` (increment) and `t` (timestamp) readonly fields directly on the `Timestamp` class
- Added `toLong` method

##### `Long`
- Updated signatures of all `Long` methods to no longer accept a `Timestamp` argument

##### tests
- Added unit tests t
NODE-2624

### `Timestamp` now no longer inherits from `Long`

`Timestamp` no longer supports `Long`'s comparison methods (`gt`, `lt`, `lte`, ...). The `high` and `low` fields have been replaced by `t` and `i` respectively to represent the timestamp and increment fields on the Timestamp object (See [bson spec here](https://bsonspec.org/spec.html)).

If your application relied on the inherited `Long` methods, we have added a `toLong()` method which converts a `Timestamp` to a `Long` and all of the previously available methods can be accessed in this way.

```ts
const ts0 = new Timestamp({t: 100, i: 1});
const ts1 = new Timestamp({t: 101, i: 1)};

// Old behaviour; no longer works
ts1.gt(ts0); // true

// New behaviour
ts1.toLong().greaterThan(ts0.toLong()); // true
```

### `Long` comparison and arithmetic methods no longer accept a `Timestamp` argument

Since `Timestamp` is no longer a subclass of `Long`, it is no longer compatible with
- `Long.add`
- `Long.subtract`
- `Long.and`
- `Long.compare`
- `Long.multiply`
- `Long.divide`
- `Long.notEquals`
- `Long.equals`
- `Long.greaterThan`
- `Long.greaterThanOrEqual`
- `Long.lessThan`
- `Long.lessThanOrEqual`
- `Long.modulo`
- `Long.rem`

And their associated aliases

### DIFF
--- a/src/extended_json.ts
+++ b/src/extended_json.ts
@@ -324,7 +324,7 @@ const BSON_TYPE_MAPPINGS = {
   ObjectId: (o: ObjectId) => new ObjectId(o),
   BSONRegExp: (o: BSONRegExp) => new BSONRegExp(o.pattern, o.options),
   BSONSymbol: (o: BSONSymbol) => new BSONSymbol(o.value),
-  Timestamp: (o: Timestamp) => Timestamp.fromBits(o.low, o.high)
+  Timestamp: (o: Timestamp) => Timestamp.fromBits(o.i, o.t)
 } as const;
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
--- a/src/extended_json.ts
+++ b/src/extended_json.ts
@@ -324,7 +324,7 @@ const BSON_TYPE_MAPPINGS = {
   ObjectId: (o: ObjectId) => new ObjectId(o),
   BSONRegExp: (o: BSONRegExp) => new BSONRegExp(o.pattern, o.options),
   BSONSymbol: (o: BSONSymbol) => new BSONSymbol(o.value),
-  Timestamp: (o: Timestamp) => Timestamp.fromBits(o.i, o.t)
+  Timestamp: (o: Timestamp) => new Timestamp(o)
 } as const;
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
--- a/src/parser/serializer.ts
+++ b/src/parser/serializer.ts
@@ -1,5 +1,5 @@
 import { Binary } from '../binary';
-import type { BSONSymbol, DBRef, Document, MaxKey } from '../bson';
+import type { BSONSymbol, DBRef, Document, MaxKey, Timestamp } from '../bson';
 import type { Code } from '../code';
 import * as constants from '../constants';
 import type { DBRefLike } from '../db_ref';
@@ -345,8 +345,7 @@ function serializeDecimal128(buffer: Uint8Array, key: string, value: Decimal128,
 
 function serializeLong(buffer: Uint8Array, key: string, value: Long, index: number) {
   // Write the type
-  buffer[index++] =
-    value._bsontype === 'Long' ? constants.BSON_DATA_LONG : constants.BSON_DATA_TIMESTAMP;
+  buffer[index++] = constants.BSON_DATA_LONG;
   // Number of written bytes
   const numberOfWrittenBytes = ByteUtils.encodeUTF8Into(buffer, key, index);
   // Encode the name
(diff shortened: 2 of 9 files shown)


src/extended_json.ts
```diff
@@ -324,7 +324,7 @@ const BSON_TYPE_MAPPINGS = {
   ObjectId: (o: ObjectId) => new ObjectId(o),
   BSONRegExp: (o: BSONRegExp) => new BSONRegExp(o.pattern, o.options),
   BSONSymbol: (o: BSONSymbol) => new BSONSymbol(o.value),
-  Timestamp: (o: Timestamp) => Timestamp.fromBits(o.low, o.high)
+  Timestamp: (o: Timestamp) => Timestamp.fromBits(o.i, o.t)
```
W-A-James, line 327: Change this to just pass into the constructor

src/timestamp.ts
```diff
@@ -107,11 +108,11 @@ export class Timestamp extends LongWithoutOverridesClass {
   /**
    * Returns a Timestamp for the given high and low bits. Each is assumed to use 32 bits.
    *
-   * @param lowBits - the low 32-bits.
-   * @param highBits - the high 32-bits.
+   * @param i - the low 32-bits.
+   * @param t - the high 32-bits.
```
W-A-James, line 112: Update the param descriptions

test/node/timestamp.test.ts
```diff
@@ -22,9 +22,9 @@ describe('Timestamp', () => {
       // We do expect it to work so that round tripping the Int32 instance inside a Timestamp works
       new BSON.Timestamp({ t: new BSON.Int32(0x7fff_ffff), i: new BSON.Int32(0x7fff_ffff) })
     ];
-
     for (const timestamp of table) {
-      expect(timestamp).to.have.property('unsigned', true);
+      expect(timestamp.t).to.be.gte(0);
+      expect(timestamp.i).to.be.gte(0);
```
W-A-James, line 27: Is this a useful check?

nbbeeken: Should we add one more test to the table that passed in `-1` for `t`/`i`?

nbbeeken (changes requested):

test/node/timestamp.test.ts
W-A-James: We have tests that check that it throws when passing in a negative number for either `t` or `i`

src/parser/serializer.ts
```diff
@@ -838,8 +863,10 @@ export function serializeInto(
         index = serializeObjectId(buffer, key, value, index);
       } else if (type === 'object' && value._bsontype === 'Decimal128') {
         index = serializeDecimal128(buffer, key, value, index);
-      } else if (value._bsontype === 'Long' || value._bsontype === 'Timestamp') {
+      } else if (value._bsontype === 'Long') {
         index = serializeLong(buffer, key, value, index);
+      } else if (value._bsontype === 'Timestamp') {
+        index = serializeTimestamp(buffer, key, value, index);
```
nbbeeken, line 869: I don't feel strongly:
```ts
// if Timestamp
index = serializeLong(buffer, key, value.toLong(), index);
```

test/node/timestamp.test.ts
```diff
@@ -153,4 +156,42 @@ describe('Timestamp', () => {
...
+      const ts = Timestamp.fromBits(0, 1);
+      expect(ts.i).to.equal(0);
+      expect(ts.t).to.equal(1);
+    });
+  });
+
+  describe('static fromNumber()', function () {
+    describe('accepts', function () {});
+    describe('rejects', function () {});
+  });
```
W-A-James, line 184: We don't have any tests that check the behaviour of `Long.fromNumber` (which this function calls), should we add any tests for it here or in long.test.ts?

src/timestamp.ts
```diff
@@ -1,20 +1,10 @@
+import { BSONValue } from './bson_value';
 import { BSONError } from './error';
 import type { Int32 } from './int_32';
 import { Long } from './long';
 
 /** @public */
 export type TimestampOverrides = '_bsontype' | 'toExtendedJSON' | 'fromExtendedJSON' | 'inspect';
```
nbbeeken, line 7: delete this

src/timestamp.ts
```diff
@@ -28,12 +18,20 @@ export interface TimestampExtended {
  * @public
  * @category BSONType
  */
-export class Timestamp extends LongWithoutOverridesClass {
+export class Timestamp extends BSONValue {
```
addaleax, line 21: There is a lot of functionality going away without obvious replacements. Should `ts1.equals(ts2)` keep working or should users manually re-implement it? `ts1.greaterThan(ts2)`? `ts1.isZero()`? Users who access the two components of a Timestamp probably use `ts.high, ts.low` or `ts.getHighBits(), ts.getLowBits()` or something similar, should that keep working?

Is it really intended to break user code this strongly? I can absolutely that it does not make sense to have e.g. a `ts1.multiply(ts2)` API, but there’s a lot of things on `Long` that really do make sense for `Timestamp` objects as well.

src/timestamp.ts
W-A-James: Was this meant to refer to the `TimeStampOverrides` type?

W-A-James:
Closing PR as motivation and user impact of not inheriting from Long is not entirely clear

PR #607, opened 2023-08-04
fix(NODE-5452): Allow undefined or null params in ObjectId.equals

vanstinator:
#### What is changing?

I've added `null` and `undefined` as new types to the `otherId` type union on `ObjectId.equals`.

I don't believe so. Functionally nothing is changing about what the function is doing. We're just making the types more permissive to match the function body.

We've been migrating away from a custom `idEquals` function in a large codebase towards `ObjectId.equals`. Our custom function allowed for `undefined` ids and it looks like `.equals` allows for them as well when reading the function body, but the Typescript definition doesn't include `undefined` or `null` in the type union. This blocks the ability to pass a potentially undefined `ObjectId` to `.equals` and leads to awkward ternaries or other hacks that shouldn't be necessary from the caller.

### DIFF
--- a/src/objectid.ts
+++ b/src/objectid.ts
@@ -194,7 +194,7 @@ export class ObjectId extends BSONValue {
    *
    * @param otherId - ObjectId instance to compare against.
    */
-  equals(otherId: string | ObjectId | ObjectIdLike): boolean {
+  equals(otherId: string | ObjectId | ObjectIdLike | undefined | null): boolean {
     if (otherId === undefined || otherId === null) {
       return false;
     }


duncanbeevers:
:+1: for a types-only change.

I could argue all day about the implementation, but this change makes the types match the implementation. 💯

src/objectid.ts
```diff
@@ -194,7 +194,7 @@ export class ObjectId extends BSONValue {
    *
    * @param otherId - ObjectId instance to compare against.
    */
-  equals(otherId: string | ObjectId | ObjectIdLike): boolean {
+  equals(otherId: string | ObjectId | ObjectIdLike | undefined | null): boolean {
     if (otherId === undefined || otherId === null) {
```
durran, line 198: I think we could maybe simplify this now to:

```suggestion
    if (otherId == null) {
```

durran (changes requested):

src/objectid.ts
avaly: This will only change the runtime implementation of the code. The types change of the parameter in this function still should be updated to accept `null` and `undefined` values.

durran: Yes, I was just thinking of also simplifying the runtime implementation as well along with the type change while we are here. But it's just a nit, I'm happy to merge if no one feels strongly about it.

durran (approved):

PR #611, opened 2023-08-20
perf(NODE-5557): move DataView and Set allocation used for double parsing and utf8 validation to nested path

billouboq:
### Description

This is a small performance improvement that has 0 impact on the code behaviour.

Things can be improved a little bit by changing the UTF-8 validation option, but this is more work, will see if I can test it later

#### What is changing?

Remove uneeded Set and dataview creation.

Also added a new benchmark with a more "natural" object to serialize

No documentation changes needed

Increasing deserialization performance

### Deserialization performance increased

If BSON data does not contain Doubles and UTF8 validation is disabled the deserializer is careful to not allocate data structures needed to support that functionality. This has shown to greatly increase (2x-1.3x) the performance of the deserializer.

Thank you @billouboq for this contribution!

  - Example: `feat([NODE-1234](https://jira.mongodb.org/browse/NODE-1234))!: rewriting everything in coffeescript`


etc/benchmarks/main.mjs
```diff
@@ -158,15 +158,41 @@ await runner({
 });
 
 await runner({
-  skip: true,
-  name: 'deserialize a large batch of documents each with an array of many Int64s',
+  skip: false,
+  name: 'deserialize a large batch of documents with multiple types',
```
nbbeeken, line 162: Instead of replacing the test that was here can we start a new one with targeting this muli-type object?

package.json
```diff
@@ -61,7 +61,8 @@
     "tsd": "^0.28.1",
     "typescript": "^5.0.4",
     "typescript-cached-transpile": "0.0.6",
-    "uuid": "^9.0.0"
+    "uuid": "^9.0.0",
+    "v8-profiler-next": "^1.9.0"
```
nbbeeken, line 65: This package was intentionally removed from our package.json here: https://github.com/mongodb/js-bson/commit/409c592524a6eec6a89eb96dee4b51e4f8adf5ae#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L64

It currently fails to install on Node.js 20 (on macos arm at least) so we don't want it to be present by default.

src/parser/deserializer.ts
```diff
@@ -284,9 +285,11 @@ function deserializeObject(
         (buffer[index++] << 16) |
         (buffer[index++] << 24);
     } else if (elementType === constants.BSON_DATA_NUMBER && promoteValues === false) {
+      const dataview = new DataView(buffer.buffer, buffer.byteOffset, buffer.byteLength);
```
nbbeeken, line 288: The thinking with declaring the Dataview at the top of the loop is that numbers are a very common data type so paying the allocation cost once and sharing it for the rest of the document's processing out weighs nesting it here and potentially allocating and garbage collecting it for every number processed. Are you able to gain any insight into that concern? Either through benching or possibly the chrome dev tools may show information related to GC

nbbeeken (changes requested):
Hey @billouboq thanks so much for contributing! I have a couple questions for you and some cleanup to get this closer to ready

Can you share some performance numbers you're observing after this change?

etc/benchmarks/main.mjs
billouboq: Yep will clean everything if the perf changes are validated

package.json
billouboq: Will remove it ! 👍

src/parser/deserializer.ts
```diff
@@ -240,7 +241,7 @@ function deserializeObject(
 
     // shouldValidateKey is true if the key should be validated, false otherwise
     let shouldValidateKey = true;
-    if (globalUTFValidation || utf8KeysSet.has(name)) {
+    if (globalUTFValidation || (utf8KeysSet && utf8KeysSet.has(name))) {
```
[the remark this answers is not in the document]
billouboq: Will change it 👍

src/parser/deserializer.ts
billouboq: When I have time I will create a Benchmark based on our most used mongodb document and see how much improvement it makes on it

billouboq:
@nbbeeken I just updated the code and add a new benchmark that is really close to my real world usage.

Made some benchmark tests and here are the mean durations :

No changes : 548.269 - 555
Remove Set creations : 543 - 545
Remove dataview creation : 499 - 502
Both changes : 484 - 486

billouboq:
@W-A-James Hello, any chance this one get's merged ?

W-A-James:
Hi @billouboq, apologies for the long turnaround time on this PR. We are currently in the middle of reworking how we do performance testing for js-bson (see NODE-3654). As a result, we're putting a pause on re-reviewing and potentially merging this and other potential performance improvements until that work is done. At that point, with more robust benchmarking infrastructure in place, we will reassess this PR and come to a decision.

Thanks for all your work on this repo.

billouboq:
Hello @W-A-James, do you news about this ?

It seems people are having performance issues with the new mongodb driver :
https://github.com/Automattic/mongoose/issues/13456#issuecomment-1869559205

billouboq:
Hello, I would like to know what's holding us from using that optimisation ? 

Is there something I can do to make it pass ?

billouboq:
Thanks for the update

> Do you think your document represents a case unique enough to be separate from the [bestbuy](https://github.com/mongodb/js-bson/blob/main/test/bench/documents/bestbuy_medium.json) test case? If so we can include this document as its own test.

I created a test that was representing something close to what I could use in a production environment, so with lot's of different types of values. We can remove it, it was more a test to check how much improvements I could have in our own environment.

> I am thinking we can split the difference here by storing the dataview on a let variable when we create it so that it can be shared after the first double creates it.

That might be a good idea indeed !

Don't hesitate to tell me what I should change or you can close that PR and make a new PR based on my code, up to you

nbbeeken:
Awesome thanks for sticking with this after so much time :) Could you rebase this and move the dataview variable back up to the top as a let, and the double code path can use `??=` to only create it once? 

I will handle running the benchmarks on my end and report back the results, TIA!

billouboq:
I just have small possible improvements to test first based on the same kind of changes from that PR. I will rebase after those testings

billouboq:
@nbbeeken I rebased and push only the needed changes 👍 

Also, I do not have the time to dig more, but I feel like we could improve date deserialisation performance by skipping creating a "new Long" and just create a small function that take lowBits and hightBits and do only this part without the context "this" :

```ts
  // taken from Long class file
  toNumber(): number {
    if (this.unsigned) return (this.high >>> 0) * TWO_PWR_32_DBL + (this.low >>> 0);
    return this.high * TWO_PWR_32_DBL + (this.low >>> 0);
  }
```

That way we avoid creating a new Long object just for the sake of converting it directly.

But it should be benchmarked to make sure it's worse it

nbbeeken (approved):
Thanks @billouboq for the help! LGTM

```ts
// bytes: { _id: new bson.Int32(2) }
bson.deserialize(bytes, {validation: {utf8:false}})
```

```txt
do_nothing x 851,290,397 ops/sec ±0.05% (196 runs sampled)
bson_deserialize_current x 3,191,767 ops/sec ±0.20% (193 runs sampled)
bson_deserialize_nested_dv x 6,522,438 ops/sec ±0.15% (194 runs sampled)
```

I ran our granular (per bson type) benchmarks and I see numbers corroborating this, for non-doubles we get way better performance, looks like DataView is not too great. I will file follow up tickets for your suggestion about Date's and I bet we can get Doubles to be even better if we use the readDoubleLE API that Node.js offers (when running in Node.js)

Filed:
- [NODE-5935](https://jira.mongodb.org/browse/NODE-5935)
- [NODE-5934](https://jira.mongodb.org/browse/NODE-5934)

aditi-khare-mongoDB (approved):
LGTM, thank you @billouboq!

PR #613, opened 2023-08-21
fix(NODE-5546): decimal 128 fromString performs inexact rounding

W-A-James:
#### What is changing?
- Corrected `Decimal128.fromString` algorithm to match that in [mongo-c-driver](https://github.com/mongodb/mongo-c-driver/blob/8b323563920fc6e38ef34f175d0a7b97607c5de8/src/libbson/src/bson/bson-decimal128.c#L477-L779)
- Unskipped `Decimal128` spec tests
- Update `Decimal128` spec tests

- No

NODE-5546

### `Decimal128` constructor now throws when detecting loss of precision

Prior to this release, `Decimal128` would round numbers with more than 34 significant digits and lose precision. Now, on detecting loss of precision, `Decimal128`'s constructor and `Decimal128.fromString` will throw a `BSONError`. This behaviour should have been the default as the `Decimal128` class was always intended to be high-precision floating point value. As such, silently rounding is undesirable behaviour.


src/decimal128.ts
```diff
@@ -357,85 +360,57 @@ export class Decimal128 extends BSONValue {
...
+
+        invalidErr(representation, 'exponent underflow');
       }
 
       if (nDigitsStored < nDigits) {
+        if (
+          representation[nDigits - 1 + Number(sawSign) + Number(sawRadix)] !== '0' &&
+          significantDigits !== 0
+        ) {
+          invalidErr(representation, 'inexact rounding not allowed');
```
nbbeeken, line 377:
```suggestion
          invalidErr(representation, 'inexact rounding');
```

same for the others

src/decimal128.ts
```diff
@@ -160,6 +160,7 @@ export class Decimal128 extends BSONValue {
   static fromString(representation: string): Decimal128 {
     // Parse state tracking
     let isNegative = false;
+    let sawSign = false;
```
nbbeeken, line 163: 🎵 I saw the sign
And it opened up my eyes, I saw the sign
Life is demanding without understanding 🎵

nbbeeken (changes requested):

src/decimal128.ts
```diff
@@ -331,7 +335,7 @@ export class Decimal128 extends BSONValue {
     // to represent user input
 
     // Overflow prevention
-    if (exponent <= radixPosition && radixPosition - exponent > 1 << 14) {
+    if (exponent <= radixPosition && radixPosition > exponent + (1 << 14)) {
```
W-A-James, line 338: @W-A-James Double check that these changes are necessary

src/decimal128.ts
```diff
@@ -342,10 +346,9 @@ export class Decimal128 extends BSONValue {
       // Shift exponent to significand and decrease
       lastDigit = lastDigit + 1;
 
-      if (lastDigit - firstDigit > MAX_DIGITS) {
+      if (lastDigit - firstDigit >= MAX_DIGITS) {
```
nbbeeken, line 349: Check this with a test case

src/decimal128.ts
```diff
@@ -357,85 +360,57 @@ export class Decimal128 extends BSONValue {
...
       // documented with ndigits_read, so inc the position of the first nonzero
       // digit and the position that digits are read to.
       if (sawRadix) {
         firstNonZero = firstNonZero + 1;
-        endOfString = endOfString + 1;
       }
-      // if negative, we need to increment again to account for - sign at start.
-      if (isNegative) {
+      // if saw sign, we need to increment again to account for - or + sign at start.
+      if (sawSign) {
```
nbbeeken, line 406: saw sign test

nbbeeken (changes requested):

src/decimal128.ts
W-A-James: @nbbeeken Taking another look, this was just reorganizing the inequality for readibility's sake, so I'd prefer to keep this change

src/decimal128.ts
W-A-James: Looking at it, `firstDigit` is always 0; it doesn't change at all throughout the entire function, so I'm going to go ahead and remove all references to it.

With regards to changing to the gte sign, the cases that trigger this branch are clamped zeros with large positive increments (see spec tests in decimal128-1.json), but any number with a sufficiently large exponent can enter this block.

If we consider a number like `1000000000000000000000000000000000e6112`, which has 1 significant digit, 34 digits before the exponent and an exponent that's above max exponent, then what happens when we have a gt sign here is that we end up with the following result:

```
> Decimal128.fromString('1000000000000000000000000000000000e6112')
new Decimal128("0E+6111")
```

since we skip the iteration when `lastDigit` reaches `MAX_DIGITS` and proceed without error despite the fact that we should have overflowed. This leads us to get to our encoding and produce the incorrect result we see above.

With the gte sign in place, we get the following result, which is what we expect

```
> Decimal128.fromString('1000000000000000000000000000000000e6112')
Uncaught:
BSONError: "1000000000000000000000000000000000e6112" is not a valid Decimal128 string - overflow
    at invalidErr (/home/wajames/js-bson/lib/bson.cjs:1402:11)
    at Decimal128.fromString (/home/wajames/js-bson/lib/bson.cjs:1540:17)
```

it seems like we should have spec test that test corner cases like this.

nbbeeken (approved):

PR #615, opened 2023-08-24
fix(NODE-5578): improve objectId deserialisation by skipping buffer allocation

billouboq:
### Description

Improve ObjectId deserialisation by around 20% by skipping one buffer allocation.
Also add a benchmark that deserialise lot's of ObjectIds

### DIFF
--- a/etc/benchmarks/main.mjs
+++ b/etc/benchmarks/main.mjs
@@ -25,6 +25,26 @@ await runner({
     bson.lib.deserialize(documents[i], { validation: { utf8: false } });
   }
 });
+
+await runner({
+  skip: false,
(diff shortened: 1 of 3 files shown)


W-A-James:
Hi @billouboq, thanks so much for this PR. Taking a look in the debugger, with this change, the `ObjectId` will not own the `Uint8Array` that is passed in. Because of this the buffer could be changed later on and mutate the value of the `ObjectId`, potentially leading to data corruption. So while this does improve performance, it breaks an underlying assumption we try to maintain with our BSON types.

billouboq:
Yeah no problem @W-A-James , I can understand. Thanks for the heads up 👍

PR #560, opened 2023-02-07
bug(NODE-5047): fix rounding error

hconn-riparian:
### Description

input: `37.499999999999999196428571428571375`
output (w/o fix): `48.50000000000000020753968253968248`
output (w/ fix): `37.49999999999999919642857142857138`

#### What is changing?

It needs to break the loop when it finds the first non-9 digit during rounding

### Double check the following

- [x] Ran `npm run lint` script
- [x] Self-review completed using the [steps outlined here](https://github.com/mongodb/node-mongodb-native/blob/HEAD/CONTRIBUTING.md#reviewer-guidelines)
- [x] PR title follows the [correct format](https://www.conventionalcommits.org/en/v1.0.0/): `type(NODE-xxxx)[!]: description`
  - Example: `feat(NODE-1234)!: rewriting everything in coffeescript`
- [x] Changes are covered by tests
- [x] New TODOs have a related JIRA ticket

PR #617, opened 2023-08-30
feat(NODE-5594): add Decimal128.fromStringWithRounding() static method

W-A-James:
#### What is changing?

- Adds `Decimal128.fromStringWithRounding` static method
- Convert `test/node/decimal128_tests.js` to typescript and add tests for `fromStringWithRounding`

- Yes, DOCSP ticket will be filed

NODE-5594

### Adds new `Decimal128.fromStringWithRounding` static method

Prior to the 6.0.0 release, `Decimal128.fromString` performed rounding behaviour by default that rounded numbers with >34 significant digits incorrectly. In addition to this bug, the intention of the `Decimal128` type and specification is to provide an interface to a high-precision floating point value, so rounding by default also went against this intention, which is why that behaviour was removed from the method in the 6.0.0 release.

We understand that some users may have come to rely on the rounding behaviour, so that has been corrected here and exposed via a new static method, `Decimal128.fromStringWithRounding`, which should now allow these users to migrate to the v6 of the driver.

Thank you to @hconn-riparian for reporting a related rounding bug and fix in #560 which has been included in this feature.
```
// 5.x
> let d = Decimal128.fromString('127341286781293491234791234667890123')
new Decimal128("1.273412867812934912347912346678901E+35")

// 6.x
> let d = Decimal128.fromString('127341286781293491234791234667890123')
Uncaught:
BSONError: "127341286781293491234791234667890123" is not a valid Decimal128 string - inexact rounding
    at invalidErr (./js-bson/lib/bson.cjs:1402:11)
    at Decimal128.fromStringInternal (./js-bson/lib/bson.cjs:1633:25)
    at Decimal128.fromString (./js-bson/lib/bson.cjs:1424:27)

> d = Decimal128.fromStringWithRounding('127341286781293491234791234667890123')
new Decimal128("1.273412867812934912347912346678901E+35")
```

### DIFF
--- a/src/decimal128.ts
+++ b/src/decimal128.ts
@@ -158,6 +158,20 @@ export class Decimal128 extends BSONValue {
    * @param representation - a numeric string representation.
    */
   static fromString(representation: string): Decimal128 {
+    return Decimal128.fromStringInternal(representation, { allowRounding: false });
+  }
+
+  /**
+   * Create a Decimal128 instance from a string representation, allowing for rounding to 34
+   * significant digits
+   *
+   * @param representation - a numeric string representation.
+   */
+  static fromStringWithRounding(representation: string): Decimal128 {
+    return Decimal128.fromStringInternal(representation, { allowRounding: true });
+  }
+
+  private static fromStringInternal(representation: string, options: { allowRounding: boolean }) {
     // Parse state tracking
     let isNegative = false;
     let sawSign = false;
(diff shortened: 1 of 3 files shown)


test/node/decimal128.test.ts
```diff
@@ -70,7 +72,7 @@ describe('Decimal128', function () {
   });
 
   it('fromString NaN input', function (done) {
-    var result = Decimal128.fromString('NaN');
+    let result = Decimal128.fromString('NaN');
```
durran, line 75: While we are here, instead of reassigning the `result` for `NaN` and `+NaN`, can we break these out into separate `it` blocks for each? So one `it` for `NaN` input and one `it` for `+NaN` input and then `result` can be a `const`. Same with the next block for `Infinity` and `+Infinity`.

test/node/decimal128.test.ts
```diff
@@ -107,8 +109,8 @@ describe('Decimal128', function () {
 
   it('fromString simple', function (done) {
     // Create decimal from string value 1
-    var result = Decimal128.fromString('1');
-    var bytes = Buffer.from(
+    let result = Decimal128.fromString('1');
```
durran, line 112: Same as above comments... Instead of reassigning `result` can we break out into multiple `it` blocks with their own assertions? (Just making comments here since there's a test refactor my opinion is we might as well clean it up further.)

test/node/decimal128.test.ts
```diff
@@ -210,8 +212,8 @@ describe('Decimal128', function () {
 
   it('fromString scientific format', function (done) {
     // Create decimal from string value 10e0
-    var result = Decimal128.fromString('10e0');
-    var bytes = Buffer.from(
+    let result = Decimal128.fromString('10e0');
```
durran, line 215: Same here for splitting out multiple `it` blocks as well as all the subsequent tests.

test/node/decimal128.test.ts
```diff
@@ -1232,5 +1235,108 @@ describe('Decimal128', function () {
         });
       }
     );
+
+    describe('fromStringWithRounding', function () {
```
durran, line 1239:
```suggestion
    describe('.fromStringWithRounding', function () {
```

durran (changes requested):

src/decimal128.ts
```diff
@@ -158,6 +158,32 @@ export class Decimal128 extends BSONValue {
    * @param representation - a numeric string representation.
    */
   static fromString(representation: string): Decimal128 {
+    return Decimal128.fromStringInternal(representation, { allowRounding: false });
```
baileympearson, line 161: since we don't really have a convention this is optional - but usually to mark something internal we'd either use a free function that isn't exported or use an `_` prefixed method instead: `Decimal128._fromString(..)`.  I prefer the former since it visually distinguishes public versus private methods and members but is cleaner than using "internal".

W-A-James:
[Link to CI run with pinned npm version](https://spruce.mongodb.com/version/64f209b0d6d80a17be52544d/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

durran (approved):

PR #621, opened 2023-09-08
feat(NODE-5579): add Decimal128.fromStringWithRounding() static method

W-A-James:
#### What is changing?

- Adds `Decimal128.fromStringWithRounding` static method
- Convert `test/node/decimal128_tests.js` to typescript and add tests for `fromStringWithRounding`

- Yes, DOCSP ticket will be filed

NODE-5594

### Adds new `Decimal128.fromStringWithRounding` static method

Prior to the merging of #620, `Decimal128.fromString` performed rounding behaviour by default that rounded numbers with >34 significant digits incorrectly. In addition to this bug, the intention of the `Decimal128` type and specification is to provide an interface to a high-precision floating point value, so rounding by default also went against this intention, which is why that behaviour was removed from the method.

We understand that some users may have come to rely on the rounding behaviour, so that has been corrected here and exposed via a new static method, `Decimal128.fromStringWithRounding`, which should now allow these users to migrate to the v6 of the driver.

Thank you to @hconn-riparian for reporting a related rounding bug and fix in #560 which has been included in this feature.
```
// pre v5.5
> let d = Decimal128.fromString('127341286781293491234791234667890123')
new Decimal128("1.273412867812934912347912346678901E+35")

// >= v5.5
> let d = Decimal128.fromString('127341286781293491234791234667890123')
Uncaught:
BSONError: "127341286781293491234791234667890123" is not a valid Decimal128 string - inexact rounding
    at invalidErr (./js-bson/lib/bson.cjs:1402:11)
    at Decimal128.fromStringInternal (./js-bson/lib/bson.cjs:1633:25)
    at Decimal128.fromString (./js-bson/lib/bson.cjs:1424:27)

> d = Decimal128.fromStringWithRounding('127341286781293491234791234667890123')
new Decimal128("1.273412867812934912347912346678901E+35")
```


W-A-James:
Blocking on #620

nbbeeken:
Should this be "feat"?

nbbeeken (changes requested):
As expected some conflicts from the "fix" pr need resolving

nbbeeken (approved):

PR #626, opened 2023-09-12
fix: throw BSONVersionError according to its message

ggondim:
#### What is changing?

- A comparison of BSON_MAJOR_VERSION

- No

- The [message thrown by `BSONVersionError`](https://github.com/mongodb/js-bson/blob/main/src/error.ts#L64) is that the BSON type should be the current major **or later**.
- The [symbol comparison at extended_json.js](https://github.com/mongodb/js-bson/blob/main/src/extended_json.ts#L361) should be an operator `<` instead a strict `!==`  to be in accordance with what the message stands.

### DIFF
--- a/src/extended_json.ts
+++ b/src/extended_json.ts
@@ -358,7 +358,7 @@ function serializeDocument(doc: any, options: EJSONSerializeOptions) {
     doc != null &&
     typeof doc === 'object' &&
     typeof doc._bsontype === 'string' &&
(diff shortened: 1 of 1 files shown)


nbbeeken:
Hi @ggondim, thanks for bringing this to our attention! Our intention with the version check is that only BSON types from the same major version are permitted. Understandably, the error message is misleading implying that "6.0.0 or later" includes later major versions. We will update the message to clarify our intention.

PR #634, opened 2023-09-27
chore(NODE-5547): fix BSON 4.x CI

nbbeeken:
#### What is changing?

- Remove nvm and backport node download script from other repos
  - backports init-node-and-npm-env.sh in place of init-nvm
  - backports install-dependencies script.
    - Remove `--continue-at -` flag, older node versions unable to be downloaded with this set
    - exempt node 6 from upgrading its npm version, npm 3 is unable to install a newer npm version
- bump `@types/node` to fix a typescript@next compilation failure

No

Fix BSON 4.x setup failures to make it usable for testing

- [x] Changes are tests

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -19,64 +19,71 @@ functions:
     - command: git.get_project
       params:
         directory: src
-    - command: shell.exec
-      params:
-        working_dir: src
-        script: |
-          # Get the current unique version of this checkout
-          if [ "${is_patch}" = "true" ]; then
-             CURRENT_VERSION=$(git describe)-patch-${version_id}
-          else
-             CURRENT_VERSION=latest
-          fi
-          export PROJECT_DIRECTORY="$(pwd)"
-
-          # get the latest version of node for given major version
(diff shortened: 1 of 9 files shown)


durran (approved):
TS old checks unrelated to this change.

nbbeeken:
It is actually related to this change, the `@types/node` update broke the old TS, but it is unrelated to our package, so we can continue with this, or I can revert that which will make TS-next fail. Pick your poison 😭

durran:
> It is actually related to this change, the `@types/node` update broke the old TS, but it is unrelated to our package, so we can continue with this, or I can revert that which will make TS-next fail. Pick your poison 😭

@nbbeeken I'd just say revert that one dependency update so everything's green.

nbbeeken:
> I can revert that which will make TS-next fail

Everything won't be green if I revert, either way we need a follow up to fix TS-old or TS-next. I figure we stick with TS-next passing because I think generally folks are on newer TS versions 4.0.2 is quite old now.

PR #635, opened 2023-10-06
feat(Node 5040): add color to BSON inspect

aditi-khare-mongoDB:
### Description
Utilizing Node.js's custom inspect functionality `[Symbol.for('nodejs.util.inspect.custom')]`, we are able to write custom functions to print out user input to BSON types. This allows users to visualize their BSON objects in color, as well as set options to their inspect function. 

#### What is changing?
Node.js colors the various types of javascript values, through the inspect function, and we can now harness that for BSON types. In addition, we support all user-defined inspect option inputs, for all types except RegExp.

This functionality is not available in the browser, in which case the Node Driver defaults to the default JS inspect. 

Yes. Users need to be provided with information about what is printed with the new inspect functionality and how to modify the options.

To help users have better control over their visualization of BSON objects, especially in the case of color visualization of values of different types.

### Node.js BSON now supports inspect options, specifically visualizing in color
| First Header  | Second Header | Second Header |
| ------------- | ------------- | ------------- | ------------- |
| Content Cell  | Content Cell  | Content Cell  | Content Cell  |
| Content Cell  | Content Cell  | Content Cell  | Content Cell  |
| Content Cell  | Content Cell  | Content Cell  | Content Cell  |
| Content Cell  | Content Cell  | Content Cell  | Content Cell  |
| Content Cell  | Content Cell  | Content Cell  | Content Cell  |
| Content Cell  | Content Cell  | Content Cell  | Content Cell  |
| Content Cell  | Content Cell  | Content Cell  | Content Cell  |
| Content Cell  | Content Cell  | Content Cell  | Content Cell  |

### DIFF
--- a/src/binary.ts
+++ b/src/binary.ts
@@ -1,4 +1,4 @@
-import { isAnyArrayBuffer, isUint8Array } from './parser/utils';
+import { getStylizeFunction, isAnyArrayBuffer, isUint8Array } from './parser/utils';
 import type { EJSONOptions } from './extended_json';
 import { BSONError } from './error';
 import { BSON_BINARY_SUBTYPE_UUID_NEW } from './constants';
@@ -264,13 +264,15 @@ export class Binary extends BSONValue {
   }
 
   /** @internal */
-  [Symbol.for('nodejs.util.inspect.custom')](): string {
-    return this.inspect();
+  [Symbol.for('nodejs.util.inspect.custom')](depth?: number, options?: unknown): string {  
+    return this.inspect(depth, options);
   }
 
-  inspect(): string {
+  inspect(depth?: number, options?: unknown): string {
+    const stylize = getStylizeFunction(options);
     const base64 = ByteUtils.toBase64(this.buffer.subarray(0, this.position));
-    return `Binary.createFromBase64("${base64}", ${this.sub_type})`;
+    const base64Arg = stylize(`"${base64}"`, 'string');
+    return `Binary.createFromBase64(${base64Arg}, ${this.sub_type})`;
   }
 }
 
@@ -465,11 +467,12 @@ export class UUID extends Binary {
    * @returns return the 36 character hex string representation.
    * @internal
    */
-  [Symbol.for('nodejs.util.inspect.custom')](): string {
-    return this.inspect();
+  [Symbol.for('nodejs.util.inspect.custom')](depth?: number, options?: unknown): string {
+    return this.inspect(depth, options);
   }
 
-  inspect(): string {
-    return `new UUID("${this.toHexString()}")`;
+  inspect(depth?: number, options?: unknown): string {
+    const stylize = getStylizeFunction(options);
+    return `new UUID(${stylize(`"${this.toHexString()}"`, 'string')})`;
   }
 }
--- a/src/binary.ts
+++ b/src/binary.ts
@@ -1,4 +1,9 @@
-import { getStylizeFunction, isAnyArrayBuffer, isUint8Array } from './parser/utils';
+import {
+  type InspectParameterFn,
+  getBasicInspectParameterFn,
+  isAnyArrayBuffer,
+  isUint8Array
+} from './parser/utils';
 import type { EJSONOptions } from './extended_json';
 import { BSONError } from './error';
 import { BSON_BINARY_SUBTYPE_UUID_NEW } from './constants';
@@ -264,15 +269,20 @@ export class Binary extends BSONValue {
   }
 
   /** @internal */
-  [Symbol.for('nodejs.util.inspect.custom')](depth?: number, options?: unknown): string {  
-    return this.inspect(depth, options);
+  [Symbol.for('nodejs.util.inspect.custom')](
+    depth?: number,
+    options?: unknown,
+    inspect?: InspectParameterFn
+  ): string {
+    return this.inspect(depth, options, inspect);
   }
 
-  inspect(depth?: number, options?: unknown): string {
-    const stylize = getStylizeFunction(options);
+  inspect(depth?: number, options?: unknown, inspect?: InspectParameterFn): string {
+    inspect ??= getBasicInspectParameterFn();
     const base64 = ByteUtils.toBase64(this.buffer.subarray(0, this.position));
-    const base64Arg = stylize(`"${base64}"`, 'string');
-    return `Binary.createFromBase64(${base64Arg}, ${this.sub_type})`;
+    const base64Arg = inspect(base64, options);
+    const subTypeArg = inspect(this.sub_type, options);
+    return `Binary.createFromBase64(${base64Arg}, ${subTypeArg})`;
   }
 }
 
@@ -467,12 +477,16 @@ export class UUID extends Binary {
    * @returns return the 36 character hex string representation.
    * @internal
    */
-  [Symbol.for('nodejs.util.inspect.custom')](depth?: number, options?: unknown): string {
-    return this.inspect(depth, options);
+  [Symbol.for('nodejs.util.inspect.custom')](
+    depth?: number,
+    options?: unknown,
+    inspect?: InspectParameterFn
+  ): string {
+    return this.inspect(depth, options, inspect);
   }
 
-  inspect(depth?: number, options?: unknown): string {
-    const stylize = getStylizeFunction(options);
-    return `new UUID(${stylize(`"${this.toHexString()}"`, 'string')})`;
+  inspect(depth?: number, options?: unknown, inspect?: InspectParameterFn): string {
+    inspect ??= getBasicInspectParameterFn();
+    return `new UUID(${inspect(this.toHexString(), options)})`;
   }
 }
--- a/src/db_ref.ts
+++ b/src/db_ref.ts
@@ -1,3 +1,4 @@
+import { getStylizeFunction, InspectParameterFn } from './parser/utils';
 import type { Document } from './bson';
 import { BSONValue } from './bson_value';
 import type { EJSONOptions } from './extended_json';
@@ -112,16 +113,28 @@ export class DBRef extends BSONValue {
   }
 
   /** @internal */
-  [Symbol.for('nodejs.util.inspect.custom')](): string {
-    return this.inspect();
+  [Symbol.for('nodejs.util.inspect.custom')](
+    depth?: number,
+    options?: unknown,
+    inspect?: InspectParameterFn
+  ): string {
+    return this.inspect(depth, options, inspect);
   }
 
-  inspect(): string {
-    // NOTE: if OID is an ObjectId class it will just print the oid string.
-    const oid =
-      this.oid === undefined || this.oid.toString === undefined ? this.oid : this.oid.toString();
-    return `new DBRef("${this.namespace}", new ObjectId("${String(oid)}")${
-      this.db ? `, "${this.db}"` : ''
-    })`;
+  inspect(depth?: number, options?: unknown, inspect?: InspectParameterFn): string {
+    const stylize = getStylizeFunction(options);
+    inspect ??= v => JSON.stringify(v);
+
+    const args = [stylize(`"${this.namespace}"`, 'string'), this.oid.inspect(depth, options)];
+
+    if (this.db) {
+      args.push(stylize(`"${this.db}"`, 'string'));
+    }
+
+    if (Object.keys(this.fields).length > 0) {
+      args.push(inspect(this.fields, options));
+    }
+
+    return `new DBRef(${args.join(', ')})`;
   }
 }
--- a/src/db_ref.ts
+++ b/src/db_ref.ts
@@ -1,8 +1,8 @@
-import { getStylizeFunction, InspectParameterFn } from './parser/utils';
 import type { Document } from './bson';
 import { BSONValue } from './bson_value';
 import type { EJSONOptions } from './extended_json';
 import type { ObjectId } from './objectid';
+import { type InspectParameterFn, getBasicInspectParameterFn } from './parser/utils';
 
 /** @public */
 export interface DBRefLike {
@@ -122,13 +122,12 @@ export class DBRef extends BSONValue {
   }
 
   inspect(depth?: number, options?: unknown, inspect?: InspectParameterFn): string {
-    const stylize = getStylizeFunction(options);
-    inspect ??= v => JSON.stringify(v);
+    inspect ??= getBasicInspectParameterFn();
 
-    const args = [stylize(`"${this.namespace}"`, 'string'), this.oid.inspect(depth, options)];
+    const args = [inspect(this.namespace, options), inspect(this.oid, options)];
 
     if (this.db) {
-      args.push(stylize(`"${this.db}"`, 'string'));
+      args.push(inspect(this.db, options));
     }
 
     if (Object.keys(this.fields).length > 0) {
--- a/src/double.ts
+++ b/src/double.ts
@@ -1,5 +1,6 @@
 import { BSONValue } from './bson_value';
 import type { EJSONOptions } from './extended_json';
+import { getStylizeFunction } from './parser/utils';
 
 /** @public */
 export interface DoubleExtended {
@@ -72,12 +73,13 @@ export class Double extends BSONValue {
   }
 
   /** @internal */
-  [Symbol.for('nodejs.util.inspect.custom')](): string {
-    return this.inspect();
+  [Symbol.for('nodejs.util.inspect.custom')](depth?: number, options?: unknown): string {
+    return this.inspect(depth, options);
   }
 
-  inspect(): string {
+  inspect(depth?: number, options?: unknown): string {
     const eJSON = this.toExtendedJSON() as DoubleExtended;
-    return `new Double(${eJSON.$numberDouble})`;
+    const stylize = getStylizeFunction(options);
+    return `new Double(${stylize(eJSON.$numberDouble, 'number')})`;
   }
 }
--- a/src/double.ts
+++ b/src/double.ts
@@ -1,6 +1,6 @@
 import { BSONValue } from './bson_value';
 import type { EJSONOptions } from './extended_json';
-import { getStylizeFunction } from './parser/utils';
+import { type InspectParameterFn, getBasicInspectParameterFn } from './parser/utils';
 
 /** @public */
 export interface DoubleExtended {
@@ -73,13 +73,16 @@ export class Double extends BSONValue {
   }
 
   /** @internal */
-  [Symbol.for('nodejs.util.inspect.custom')](depth?: number, options?: unknown): string {
-    return this.inspect(depth, options);
+  [Symbol.for('nodejs.util.inspect.custom')](
+    depth?: number,
+    options?: unknown,
+    inspect?: InspectParameterFn
+  ): string {
+    return this.inspect(depth, options, inspect);
   }
 
-  inspect(depth?: number, options?: unknown): string {
-    const eJSON = this.toExtendedJSON() as DoubleExtended;
-    const stylize = getStylizeFunction(options);
-    return `new Double(${stylize(eJSON.$numberDouble, 'number')})`;
+  inspect(depth?: number, options?: unknown, inspect?: InspectParameterFn): string {
+    inspect ??= getBasicInspectParameterFn();
+    return `new Double(${inspect(this.valueOf(), options)})`;
   }
 }
--- a/src/parser/utils.ts
+++ b/src/parser/utils.ts
@@ -27,3 +27,23 @@ export function isMap(d: unknown): d is Map<unknown, unknown> {
 export function isDate(d: unknown): d is Date {
   return Object.prototype.toString.call(d) === '[object Date]';
 }
+
+/** @internal */
+export type StylizeFunction = (x: unknown, style: string) => string;
+/** @internal */
+export type InspectParameterFn = (x: unknown, options: unknown) => string;
+export function getStylizeFunction(options?: unknown): StylizeFunction {
+  const stylizeExists = 
+    options != null && 
+    typeof options === 'object' && 
+    'stylize' in options && 
+    typeof options.stylize === 'function';
+  
+  if (stylizeExists) {
+    return options.stylize as 
+    (x: unknown, style: string) => string; 
+  } else {
+    // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+    return v => `${v}`;
+  } 
+}
--- a/src/parser/utils.ts
+++ b/src/parser/utils.ts
@@ -32,18 +32,20 @@ export function isDate(d: unknown): d is Date {
 export type StylizeFunction = (x: unknown, style: string) => string;
 /** @internal */
 export type InspectParameterFn = (x: unknown, options: unknown) => string;
+export function getBasicInspectParameterFn(): InspectParameterFn {
+  // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+  return v => `${v}`;
+}
 export function getStylizeFunction(options?: unknown): StylizeFunction {
-  const stylizeExists = 
-    options != null && 
-    typeof options === 'object' && 
-    'stylize' in options && 
+  const stylizeExists =
+    options != null &&
+    typeof options === 'object' &&
+    'stylize' in options &&
     typeof options.stylize === 'function';
-  
+
   if (stylizeExists) {
-    return options.stylize as 
-    (x: unknown, style: string) => string; 
+    return options.stylize as (x: unknown, style: string) => string;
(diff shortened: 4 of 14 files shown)


src/binary.ts
```diff
@@ -264,13 +269,20 @@ export class Binary extends BSONValue {
...
   /** @internal */
-  [Symbol.for('nodejs.util.inspect.custom')](): string {
-    return this.inspect();
+  [Symbol.for('nodejs.util.inspect.custom')](
+    depth?: number,
+    options?: unknown,
+    inspect?: InspectParameterFn
+  ): string {
+    return this.inspect(depth, options, inspect);
   }
```
addaleax, line 278: Does it make sense to define this function (or switch them and then define `inspect()`) on `BSONValue`? `inspect()` is already declared as an abstract method there right now, and that would get rid of quite a bit of repetition.

src/parser/utils.ts
```diff
@@ -27,3 +27,24 @@ export function isMap(d: unknown): d is Map<unknown, unknown> {
 export function isDate(d: unknown): d is Date {
   return Object.prototype.toString.call(d) === '[object Date]';
 }
+
+/** @internal */
+export type StylizeFunction = (x: unknown, style: string) => string;
```
addaleax, line 32:
```suggestion
export type StylizeFunction = (x: string, style: string) => string;
```

since stylizing takes an existing string already?

src/parser/utils.ts
```diff
@@ -27,3 +27,24 @@ export function isMap(d: unknown): d is Map<unknown, unknown> {
...
+export type InspectParameterFn = (x: unknown, options: unknown) => string;
+export function getBasicInspectParameterFn(): InspectParameterFn {
+  // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+  return v => `${v}`;
+}
```
addaleax, line 37: This doesn’t feel like the most impactful question so please feel free to just ignore and resolve, but why the extra level of indirection? This could be

```suggestion
// eslint-disable-next-line @typescript-eslint/restrict-template-expressions
export function basicInspectParameterFn: InspectParameterFn = v => `${v}`;
```

right?

src/parser/utils.ts
```diff
@@ -27,3 +27,24 @@ export function isMap(d: unknown): d is Map<unknown, unknown> {
...
+export function getStylizeFunction(options?: unknown): StylizeFunction {
+  const stylizeExists =
+    options != null &&
+    typeof options === 'object' &&
+    'stylize' in options &&
+    typeof options.stylize === 'function';
+
+  if (stylizeExists) {
+    return options.stylize as (x: unknown, style: string) => string;
```
addaleax, line 46: again, super tiny nit, feel free to ignore, but since you already have the type defined:

```suggestion
    return options.stylize as StylizeFunction;
```

src/db_ref.ts
```diff
@@ -112,16 +113,27 @@ export class DBRef extends BSONValue {
...
+    options?: unknown,
+    inspect?: InspectParameterFn
+  ): string {
+    return this.inspect(depth, options, inspect);
   }
 
-  inspect(): string {
-    // NOTE: if OID is an ObjectId class it will just print the oid string.
-    const oid =
-      this.oid === undefined || this.oid.toString === undefined ? this.oid : this.oid.toString();
```
addaleax, line 122: :tada: 

This should enable us to get rid of our custom inspection in mongosh for DBRefs :slightly_smiling_face:

src/binary.ts
aditi-khare-mongoDB: could you clarify what you mean by switching them?

addaleax: Mh, there are two methods here, `inspect()` and `[Symbol.for('nodejs.util.inspect.custom')]()`, where one just directly calls the other, and it doesn’t really matter which one calls which, so if one of them gets moved to the superclass, that could also be `[Symbol.for('nodejs.util.inspect.custom')]()` (i.e. keep things as is) or `inspect()` (i.e. switch around which method calls the other one).

W-A-James (changes requested):
Looks good, but I agree with Anna's comments.

src/db_ref.ts
aditi-khare-mongoDB: Great!

src/parser/utils.ts
aditi-khare-mongoDB: Currently, getStylizeFunction takes in only a string; however, that is not the only potential case it can be used in the future, e.g. as for non-string types that require custom stylization.

src/parser/utils.ts
aditi-khare-mongoDB: Sounds good!

src/parser/utils.ts
aditi-khare-mongoDB: Yes, I forget functions are first class objects sometimes!

src/binary.ts
aditi-khare-mongoDB: Sounds good, I removed  custom `[Symbol.for('nodejs.util.inspect.custom')]()` logic for all except regExp since it has a unique signature.

src/db_ref.ts
```diff
@@ -111,17 +112,26 @@ export class DBRef extends BSONValue {
     return new DBRef(doc.$ref, doc.$id, doc.$db, copy);
   }
 
-  /** @internal */
-  [Symbol.for('nodejs.util.inspect.custom')](): string {
-    return this.inspect();
-  }
+  inspect(depth?: number, options?: unknown, inspect?: InspectParameterFn): string {
+    const addQuotes = inspect ? false : true;
```
baileympearson, line 116:
```suggestion
    const addQuotes = !inspect;
```

src/db_ref.ts
```diff
@@ -111,17 +112,26 @@ export class DBRef extends BSONValue {
...
+      : inspect(this.oid, options);
+    const args = [namespaceArg, objectIDArg];
+
+    if (this.db) {
+      args.push(addQuotes ? `'${inspect(this.db, options)}'` : inspect(this.db, options));
+    }
+
+    if (Object.keys(this.fields).length > 0) {
+      args.push(inspect(this.fields, options));
+    }
```
baileympearson, line 134:
```suggestion
    const args = [
     inspect(this.namespace, options),
     inspect(this.oid, options),
     
     // this is stylistic - but a commonly used technique to conditionally add items to an array using the `...` operator
     ...(this.db ? [inspect(this.db, options)] : []),
     ...(Object.keys(this.fields).length > 0 ? inspect(this.fields, options) : [])
    ].map(arg => addQuotes ? `'${arg}'` : arg);
    
    return `new DBRef(${args.join(', ')})`;
```

optional.  specifically, I like the idea of building up the args array and then conditionally wrapping all the args in quotes instead of using a ternary operator for each argument individually.  how we build up the array is another optional suggestion - I like using the `...` operator with the ternary operator but that's personal preference.

src/db_ref.ts
baileympearson: this applies to all occurrences of this logic in the PR

baileympearson (changes requested):

src/binary.ts
addaleax: It doesn’t have a unique signature, though, right? It doesn’t make use of the optional third argument but that’s still fine.

src/bson_value.ts
```diff
@@ -10,8 +11,16 @@ export abstract class BSONValue {
     return BSON_MAJOR_VERSION;
   }
 
+  [Symbol.for('nodejs.util.inspect.custom')](
```
W-A-James, line 14: Let's mark this as internal

src/regexp.ts
```diff
@@ -103,12 +104,15 @@ export class BSONRegExp extends BSONValue {
     throw new BSONError(`Unexpected BSONRegExp EJSON object form: ${JSON.stringify(doc)}`);
   }
 
-  /** @internal */
-  [Symbol.for('nodejs.util.inspect.custom')](): string {
-    return this.inspect();
+   /** @internal */
+   [Symbol.for('nodejs.util.inspect.custom')](depth?: number, options?: unknown): string {
```
W-A-James, line 108: lint failure here

W-A-James (changes requested):

src/bson_value.ts
addaleax: … even though it is just implementing a standardized API of the platform? What are you trying to hide from users here?

W-A-James: Good point here. Please disregard.

src/parser/utils.ts
aditi-khare-mongoDB: Oh I see, I didn't realize these were the Node.js `stylize` function's argument expectations. I'll be sure to have 2 string arguments for my `stylizeFunction` type. I wasn't able to find the documentation on the `stylize` function on the internet could you link me to it?

src/db_ref.ts
aditi-khare-mongoDB: I'll be sure to change this for my next commit!

src/parser/utils.ts
addaleax: Yeah, `stylize` is definitely underdocumented. It is only mentioned in [the docs for custom inspect() functions in Node.js](https://nodejs.org/api/util.html#custom-inspection-functions-on-objects), but the TS types [do have the right signature](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/cbdc741ac84952a5202b1e822f87ec0017ebd6db/types/node/util.d.ts#L109) and then there’s also always the [actual Node.js source code](https://github.com/nodejs/node/blob/78a15702dd1bc03505ee130b0ad999ee292b6581/lib/internal/util/inspect.js#L560-L572) which doesn’t explicitly say anything but seems to heavily imply to me that the first argument should always be a string (calling it `str` and conditionally either returning that original value or a string that includes it which can only lead to a consistent return type if `str` is always a string to begin with).

src/regexp.ts
aditi-khare-mongoDB: I'll remove this excess function

src/db_ref.ts
aditi-khare-mongoDB: Sounds good, I committed these changes and added one small tweak to args[1].

test/node/bson_test.js
```diff
@@ -1841,20 +1841,63 @@ describe('BSON', function () {
...
+    context('Code', function () {
+      it('when non-nested fields', function () {
+        const code = new Code('this.a > i', { i: 1 });
+        expect(inspect(code)).to.equal(`new Code('this.a > i', { i: 1 })`);
+      });
+      it('when non-nested fields', function () {
+        const code = new Code('this.a > i', { a: 1, b: { nest: 'mine' } });
+        expect(inspect(code)).to.equal(`new Code('this.a > i', { a: 1, b: { nest: 'mine' } })`);
+      });
+      it('when multiline code', function () {
```
W-A-James, line 1853:
```suggestion
      it('when it contains multiline code', function () {
```

test/node/bson_test.js
```diff
@@ -1841,20 +1841,63 @@ describe('BSON', function () {
...
      * @ignore
      */
-    it('DBRef', function () {
-      const oid = new ObjectId('deadbeefdeadbeefdeadbeef');
-      const dbref = new DBRef('namespace', oid, 'integration_tests_');
-      expect(inspect(dbref)).to.equal(
-        `new DBRef('namespace', new ObjectId('deadbeefdeadbeefdeadbeef'), 'integration_tests_')`
-      );
+    context('DBRef', function () {
+      it('when non-nested fields', function () {
```
W-A-James, line 1884:
```suggestion
      it('when it does not contain nested fields', function () {
```

test/node/bson_test.js
```diff
@@ -1841,20 +1841,63 @@ describe('BSON', function () {
...
+        const oid = new ObjectId('deadbeefdeadbeefdeadbeef');
+        const dbref = new DBRef('namespace', oid, 'integration_tests_');
+        expect(inspect(dbref)).to.equal(
+          `new DBRef('namespace', new ObjectId('deadbeefdeadbeefdeadbeef'), 'integration_tests_')`
+        );
+      });
+      it('when nested fields', function () {
```
W-A-James, line 1891:
```suggestion
      it('when it contains nested fields', function () {
```

test/colors.mjs
```diff
@@ -0,0 +1,58 @@
+'use strict';
```
W-A-James, line 1: Quick question, was this used to generate the screenshots?

Adding a comment to the top of this file to explain what it's for would be useful.
I would also move this to `etc/colors.mjs`. @baileympearson does that sound right?

W-A-James (changes requested):

src/symbol.ts
```diff
@@ -52,7 +49,20 @@ export class BSONSymbol extends BSONValue {
   }
 
   /** @internal */
-  [Symbol.for('nodejs.util.inspect.custom')](): string {
-    return this.inspect();
+  [Symbol.for('nodejs.util.inspect.custom')](
```
baileympearson, line 52: Just confirming - were we planning on removing this custom implementation?

test/colors.mjs
aditi-khare-mongoDB: This is what was used to generate the screenshots, I can add a comment to the top of this, and also move it to etc/colors.mjs once we hear back from Bailey

src/bson_value.ts
```diff
@@ -10,8 +11,16 @@ export abstract class BSONValue {
...
+    depth?: number,
+    options?: unknown,
+    inspect?: InspectParameterFn
+  ): string {
+    return this.inspect(depth, options, inspect);
+  }
+
   /** @public */
-  public abstract inspect(): string;
+  public abstract inspect(depth?: number, options?: unknown, inspect?: InspectParameterFn): string;
```
baileympearson, line 23: Would it be possible to provide sensible defaults here instead of making all the arguments optional?

```suggestion
  public abstract inspect(depth: number = 3, options: unknown = { stylize: basicInspectParameterFn }, inspect: InspectParameterFn = basicInspectParameterFn): string;
```

Then all the implementations in each BSON type don't need to worry about setting defaults

src/parser/utils.ts
```diff
@@ -27,3 +27,23 @@ export function isMap(d: unknown): d is Map<unknown, unknown> {
...
   return Object.prototype.toString.call(d) === '[object Date]';
 }
+
+/** @internal */
+export type StylizeFunction = (x: string, style: string) => string;
+export type InspectParameterFn = (x: unknown, options: unknown) => string;
+// eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+export const basicInspectParameterFn: InspectParameterFn = v => `${v}`;
+
+export function getStylizeFunction(options?: unknown): StylizeFunction {
```
baileympearson, line 37: This is only used in regexp, right?  Maybe we could move this function out of the utils and into that file.

src/parser/utils.ts
```diff
@@ -27,3 +27,23 @@ export function isMap(d: unknown): d is Map<unknown, unknown> {
 export function isDate(d: unknown): d is Date {
   return Object.prototype.toString.call(d) === '[object Date]';
 }
+
+/** @internal */
+export type StylizeFunction = (x: string, style: string) => string;
+export type InspectParameterFn = (x: unknown, options: unknown) => string;
+// eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+export const basicInspectParameterFn: InspectParameterFn = v => `${v}`;
```
baileympearson, line 35:
```suggestion
export const defaultInspect: InspectParameterFn = v => `${v}`;
```
We're using this as a default, so maybe using "default" in the name provides more information than "basic"

test/colors.mjs
baileympearson: If we decide to keep this file, `etc` seems appropriate.  But if this was just used to generate screen shots for the release, I'm not sure it provides much value and we can remove it.

baileympearson (changes requested):

test/colors.mjs
aditi-khare-mongoDB: I'll go ahead and remove the file.

src/parser/utils.ts
aditi-khare-mongoDB: sounds good! I'll change it.

src/parser/utils.ts
aditi-khare-mongoDB: Sounds good, I'll move it into the regexp function. My original reasoning was that we may want to stylize other types in special cases in the future, so it's best to have the stylize function available everywhere.

addaleax: Yeah, I’d agree with @aditi-khare-mongoDB and also slightly lean towards keeping it in the utils file, I know it’s not a great thing to have a “utils” file/directory/package/etc. where lots of super random stuff ends up, but the function is really not very regexp-specific either :slightly_smiling_face:

test/node/bson_test.js
```diff
@@ -1841,20 +1841,63 @@ describe('BSON', function () {
     /**
      * @ignore
      */
-    it('Code', function () {
-      const code = new Code('this.a > i', { i: 1 });
-      expect(inspect(code)).to.equal(`new Code('this.a > i', { i: 1 })`);
+    context('Code', function () {
+      it('when non-nested fields', function () {
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Sounds good, I made this change in my latest commit!

src/symbol.ts
aditi-khare-mongoDB: Yes, I for some reason didn't catch this comment in my first pass, but I'll make the change.

src/bson_value.ts
aditi-khare-mongoDB: It would be nice if we were able to make some sensible defaults here! However, to my knowledge, since this is an abstract method we can't assign values to arguments without the `"A parameter initializer is only allowed in a function or constructor implementation."` error popping up.

W-A-James (approved):

baileympearson (dismissed):

src/double.ts
```diff
@@ -71,13 +72,8 @@ export class Double extends BSONValue {
...
-  [Symbol.for('nodejs.util.inspect.custom')](): string {
-    return this.inspect();
-  }
-
-  inspect(): string {
-    const eJSON = this.toExtendedJSON() as DoubleExtended;
-    return `new Double(${eJSON.$numberDouble})`;
+  inspect(depth?: number, options?: unknown, inspect?: InspectParameterFn): string {
+    inspect ??= defaultInspect;
+    return `new Double(${inspect(this.valueOf(), options)})`;
```
nbbeeken, line 77:
```suggestion
    return `new Double(${inspect(this.value, options)})`;
```

This is what we have in Int32, and I think there's no need for the indirection of invoking the valueOf() function.

src/regexp.ts
```diff
@@ -103,12 +104,27 @@ export class BSONRegExp extends BSONValue {
     throw new BSONError(`Unexpected BSONRegExp EJSON object form: ${JSON.stringify(doc)}`);
   }
 
-  /** @internal */
-  [Symbol.for('nodejs.util.inspect.custom')](): string {
-    return this.inspect();
+  inspect(depth?: number, options?: unknown): string {
+    const stylize = getStylizeFunction(options);
+    const pattern = stylize(`'${this.pattern}'`, 'regexp');
+    const flags = stylize(`'${this.options}'`, 'regexp');
```
nbbeeken, line 110: Because of the way we build this string we will emit a broken string if `pattern` contains a single quote `'`, can we use inspect to get a properly quoted string and then stylize it?

src/timestamp.ts
```diff
@@ -141,12 +142,10 @@ export class Timestamp extends LongWithoutOverridesClass {
...
-  [Symbol.for('nodejs.util.inspect.custom')](): string {
-    return this.inspect();
-  }
-
-  inspect(): string {
-    return `new Timestamp({ t: ${this.getHighBits()}, i: ${this.getLowBits()} })`;
+  inspect(depth?: number, options?: unknown, inspect?: InspectParameterFn): string {
+    inspect ??= defaultInspect;
+    const t = inspect(this.getHighBits(), options);
+    const i = inspect(this.getLowBits(), options);
```
nbbeeken, line 148: Let's change this to align with the way `toExtendedJSON` gets the high and low bits. Specifically, it is more aligned with what timestamp is to view each component as an unsigned value, which is what the shift 0 does.

```js
> new BSON.Timestamp(BSON.Timestamp.MAX_VALUE)
new Timestamp({ t: -1, i: -1 })
// should be
new Timestamp({ t: 4294967295, i: 4294967295 })
```

test/node/binary.test.ts
```diff
@@ -81,27 +81,27 @@ describe('class Binary', () => {
...
-    it('when value is empty returns "Binary.createFromBase64("", 0)"', () => {
-      expect(new Binary(new Uint8Array(0)).inspect()).to.equal('Binary.createFromBase64("", 0)');
+    it(`when value is empty returns "Binary.createFromBase64('', 0)"`, () => {
+      expect(new Binary(new Uint8Array(0)).inspect()).to.equal(`Binary.createFromBase64('', 0)`);
     });
 
-    it('when value is default with a subtype returns "Binary.createFromBase64("", 35)"', () => {
-      expect(new Binary(null, 0x23).inspect()).to.equal('Binary.createFromBase64("", 35)');
+    it(`when value is default with a subtype returns "Binary.createFromBase64('', 35)"`, () => {
+      expect(new Binary(undefined, 0x23).inspect()).to.equal(`Binary.createFromBase64('', 35)`);
```
nbbeeken, line 93: Why this change?

src/regexp.ts
```diff
@@ -103,12 +104,27 @@ export class BSONRegExp extends BSONValue {
...
+    typeof options === 'object' &&
+    'stylize' in options &&
+    typeof options.stylize === 'function';
 
-  inspect(): string {
-    return `new BSONRegExp(${JSON.stringify(this.pattern)}, ${JSON.stringify(this.options)})`;
+  if (stylizeExists) {
+    return options.stylize as StylizeFunction;
+  } else {
+    return defaultInspect;
```
nbbeeken, line 128: super nit, generally if we early return inside an if then we don't write an `else` block

src/bson_value.ts
```diff
@@ -10,8 +11,16 @@ export abstract class BSONValue {
     return BSON_MAJOR_VERSION;
   }
 
+  [Symbol.for('nodejs.util.inspect.custom')](
+    depth?: number,
+    options?: unknown,
+    inspect?: InspectParameterFn
+  ): string {
+    return this.inspect(depth, options, inspect);
```
nbbeeken, line 19: Let's chat on slack about this because I think I may be overlooking some complexity, could this function be responsible for defaulting the inspect function when it is not provided, and could the defaultInspect function include the logic for "addQuotes"?

src/parser/utils.ts
```diff
@@ -27,3 +27,7 @@ export function isMap(d: unknown): d is Map<unknown, unknown> {
 export function isDate(d: unknown): d is Date {
   return Object.prototype.toString.call(d) === '[object Date]';
 }
+
+export type InspectParameterFn = (x: unknown, options: unknown) => string;
+// eslint-disable-next-line @typescript-eslint/restrict-template-expressions
+export const defaultInspect: InspectParameterFn = v => `${v}`;
```
nbbeeken, line 33: This default inspect is not good enough to handle types that do not stringify well, notably, objects will become `"[object Object]"`. We know `JSON.stringify` has limitations but I think we should strive to get a string that is representative of the data.

nbbeeken (changes requested):

src/double.ts
aditi-khare-mongoDB: Sounds good, I'll change this for my next commit

src/bson_value.ts
nbbeeken: From IRL: We can continue to rely on JSON.stringify as we had prior it will handle the quoting and escaping for us as long as the original type is passed in.

Ideally, we can handle ES Map and bigint in a replacer function without too much complexity

test/node/binary.test.ts
aditi-khare-mongoDB: Honestly, no idea, I'll change it back

nbbeeken (approved):

test/colors.mjs
W-A-James, line 1: Any reason we brought this back?

W-A-James (changes requested):
LGTM, but not sure why we brought back test/colors.mjs

test/colors.mjs
aditi-khare-mongoDB: Oh no reason, I'll delete it, I was just using it for personal testing purposes.

W-A-James (approved):

nbbeeken (approved):

PR #636, opened 2023-11-02
test(NODE-5692): Add new benchmarks using bson-bench

W-A-James:
##### New benchmarks

* Adds new granular benchmarks to `test/bench/granular`
	* Note that these new benchmarks directly implement the AC for deserialization and serialization tests laid out in the ticket
* Adds bson spec benchmarks to `test/bench/spec`
	* Note that these tests perform the migration of bson benchmark testing from the driver repository to the js-bson repository and will require the removal of the bsonBench tests from the driver repository in a follow-up PR
* Adds test documents to `test/bench/documents`
	* These tests have all been generated using the new file added to `test/bench/etc/generate_documents.ts` and understanding these documents is better done by looking at the logic in this file
* Adds new utility scripts to `test/bench/etc`
	* `generate_documents.ts` - Regenerates documents in `test/bench/documents`
	* `run_granular_benchmarks.js` - runs all compiled benchmarks in `test/bench/lib/granular` and consolidates results into `resultsCollected.json` in perf.send format
	* `convertToCSV.js` - Reads in a data in perf.send format from stdin and writes it to stdout converted to a csv more easily ingestible by google sheets, or other data-analysis tools

##### CI changes
* Adds new CI task for granular benchmarks
	* Changed `check:bench` script in package.json to `check:granular-bench`
		* this script builds the benchmark typescript files in `test/bench/granular` and runs `run_granular_benchmarks.js`
	* Added `run granular benchmarks` function to config.yml
	* Added `.evergreen/run-granular-benchmarks.sh`
	* Added `run-granular-benchmarks-node-18` task in place of `run-benchmarks-node-18` task
	* Runs `run-granular-benchmarks-node-18` task on perf distribution
* Adds new CI task for spec benchmarks
	* Added `check:spec-bench` script in package.json
	* Added `.evergreen/run-spec-benchmarks.sh`
	* Added `run spec benchmarks` function to config.yml
	* Added `run-spec-benchmarks-node-18` task
	* Runs `run-granular-benchmarks-node-18` task on perf distribution
	
##### Removals
* Removes old benchmarking infrastructure from `etc/benchmarks`
* Removes old benchmarking infrastructure from `test/bench/src`

No

NODE-5692


test/bench/etc/generate_documents.ts
```diff
@@ -0,0 +1,191 @@
...
+
+// string
+generateSimpleTests('string', () => randomBytes(30).toString('base64'));
+
+// binary
+generateSimpleTests('binary', () => new bson.Binary(randomBytes(30)));
+for (const { name, size } of [
+  { name: 'small', size: 1024 },
+  { name: 'medium', size: 1024 ** 2 }
+]) {
```
W-A-James, line 86: Note the absence of a large binary document; EJSON stringify fails to stringify documents with Binary instances of size 1024 ** 3 bytes

test/bench/etc/run_granular_benchmarks.js
```diff
@@ -0,0 +1,93 @@
...
+const fs = require('fs/promises');
+const path = require('path');
+const { Task } = require('dbx-js-tools/packages/bson-bench');
+
+const BENCHMARK_REGEX = /(.*)\.bench\.js$/;
+const BENCHMARK_PATH = path.resolve(`${__dirname}/../lib/granular`);
+const DOCUMENT_ROOT = path.resolve(`${__dirname}/../documents`);
+const LIBRARY_PATH = path.resolve(`${__dirname}/../../../.`);
+(async () => {
+  // HACK : run one dummy task with the local bson to ensure it's available for subsequent suites
```
nbbeeken, line 17: If this is a hack what is the non-hack way to accomplish this?

test/bench/etc/run_granular_benchmarks.js
```diff
@@ -0,0 +1,93 @@
...
+      const child = cp.fork(`${BENCHMARK_PATH}/${dirent}`);
+      if (child.stdout) child.stdout.pipe(process.stdout);
+      if (child.stderr) child.stdout.pipe(process.stderr);
+
+      await new Promise((resolve, reject) =>
+        child.once('exit', code => {
+          if (code !== 0) return reject();
+          return resolve();
+        })
+      );
```
nbbeeken, line 42: Could we use `import {once} from 'events'`? just trying to simplify manual promise construction.

test/bench/etc/run_granular_benchmarks.js
```diff
@@ -0,0 +1,93 @@
...
+      if (set.has(key)) throw new Error(`Found a duplicate testName:Option pair: ${key}`);
+      set.add(key);
+    }
+  }
+
+  console.log('No duplcate testName:Option pairs found. Now merging files...');
+
+  // Iterate over all result files and merge into one file
+  const collectedResults = await fs.open('resultsCollected.json', 'w+');
+  await collectedResults.write('[\n');
```
nbbeeken, line 74: What is the motivation for writing this per line vs gathering all the content and using `writeFile`?

test/bench/etc/run_granular_benchmarks.js
```diff
@@ -0,0 +1,93 @@
...
+
+  if (resultPaths.length === 0) throw new Error('Benchmarks did not run successfully');
+
+  // Ensure that there are no duplicate test-name/options pairs as this will prevent us from
+  // uploading with perf.send
+  const set = new Set();
+  for (const resultFile of resultPaths) {
+    const results = require(resultFile);
+    for (const entry of results) {
+      const name = entry.info.test_name;
```
nbbeeken, line 61: I'm not precisely following where it comes from but I think this is where the string like:

"timestamp_array_1000._deserialize_bson:/data/mci/20eab6d75d6f8a3717b1f015dd7a7648/src"

is coming from, don't we need this name to be consistent so that the trend chart can build on the history? I imagine that commit hash is going to change, right?

test/bench/granular/boolean.bench.ts
```diff
@@ -0,0 +1,49 @@
...
+  // deserialize
+  for (const documentPath of testDocs) {
+    for (const promoteValues of BOOL) {
+      suite.task({
+        documentPath,
+        library: LIBRARY_SPEC,
+        iterations: ITERATIONS,
+        warmup: WARMUP,
+        operation: 'deserialize',
+        options: { ...OPTIONS['deserialize'], promoteValues }
```
nbbeeken, line 31:
```suggestion
        options: { ...OPTIONS.deserialize, promoteValues }
```

I see this in a number of places, but I won't comment on more, any reason for the change in syntax?

test/bench/etc/generate_documents.ts
```diff
@@ -0,0 +1,191 @@
...
+type numericConstructor =
+  | typeof bson.Double
+  | typeof bson.Long
+  | typeof bson.Int32
+  | typeof bson.Decimal128;
+type numericType = bson.Double | bson.Long | bson.Int32 | bson.Decimal128;
+
+const DOCUMENT_PATH = resolve(`${__dirname}/../documents`);
+
+function generateSimpleTests(typeName: string, newFn: (...arg: any) => any) {
```
nbbeeken, line 21: TODO: Asking the team about just generating in CI

test/bench/etc/generate_documents.ts
```diff
@@ -0,0 +1,191 @@
...
+generateSimpleTests(
+  'code-with-scope',
+  () =>
+    new bson.Code(
+      function () {
+        // @ts-expect-error nested
+        console.log(left);
+        // @ts-expect-error nested
+        console.log(right);
+      },
```
nbbeeken, line 122: Can we use the same function from the Code test, so that the test only changes one variable between them

test/bench/etc/generate_documents.ts
```diff
@@ -0,0 +1,191 @@
...
+      {
+        left: 100,
+        right: 10_000
+      }
```
nbbeeken, line 126: Is this a bit too simple? Maybe we can put Best Buy medium in here?

test/bench/etc/run_granular_benchmarks.js
```diff
@@ -0,0 +1,93 @@
...
+    .run()
+    .catch(() => null);
+
+  // Run all benchmark files
+  const lib = await fs.readdir(BENCHMARK_PATH);
+  for await (const dirent of lib) {
+    if (BENCHMARK_REGEX.test(dirent)) {
+      const child = cp.fork(`${BENCHMARK_PATH}/${dirent}`);
+      if (child.stdout) child.stdout.pipe(process.stdout);
+      if (child.stderr) child.stdout.pipe(process.stderr);
```
nbbeeken, line 35: isn't there a `stdio: 'inherit'` option for this?

.evergreen/run-granular-benchmarks.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+source "${PROJECT_DIRECTORY}/.evergreen/init-node-and-npm-env.sh"
+WARMUP=$WARMUP
```
nbbeeken, line 4: xtrace? after the npm env, just for debugging the env settings

nbbeeken (changes requested):

test/bench/etc/run_granular_benchmarks.js
W-A-James: The best way to go about fixing this would be to expose the `Package` type from the bson-bench library repository and run `Package.install`.

test/bench/etc/run_granular_benchmarks.js
W-A-James: That's a really good catch. I didn't notice that. The test name joins the test file name, the operation and the library used. We can set the test configurations to install from the main branch and that should have us covered as that won't change. That should give us a test name like `timestamp_array_1000._deserialize_bson#main`.

test/bench/etc/run_granular_benchmarks.js
W-A-James: Since it's possible to do generate the csv line by line, I preferred doing it that way instead of having two complete representations of the file in memory at once.

test/bench/etc/run_granular_benchmarks.js
nbbeeken: Can we allude to that in the comment or file a ticket to expose that?

test/bench/etc/generate_documents.ts
nbbeeken: Is this something we should fix?

nbbeeken (changes requested):
Small follow up Qs, LGTM

test/bench/etc/generate_documents.ts
W-A-James: The issue here is that when converting `Binary` instances to EJSON, we convert them to a base64 string. We end up running up against Node's string character limit, so I don't think this is something we can fix easily here.

```
Error: Cannot create a string longer than 0x1fffffe8 characters
    at Object.slice (node:buffer:695:37)
    at Buffer.toString (node:buffer:865:14)
    at Object.toBase64 (/home/wajames/js-bson/src/utils/node_byte_utils.ts:103:54)
    at Binary.toExtendedJSON (/home/wajames/js-bson/src/binary.ts:201:36)
    at serializeDocument (/home/wajames/js-bson/src/extended_json.ts:392:19)
    at serializeValue (/home/wajames/js-bson/src/extended_json.ts:298:58)
    at serializeDocument (/home/wajames/js-bson/src/extended_json.ts:341:23)
    at serializeValue (/home/wajames/js-bson/src/extended_json.ts:298:58)
    at Object.stringify (/home/wajames/js-bson/src/extended_json.ts:475:15)
    at Object.<anonymous> (/home/wajames/js-bson/test/bench/etc/generate_documents.ts:92:16) {
  code: 'ERR_STRING_TOO_LONG'
```

test/bench/etc/run_granular_benchmarks.js
W-A-James: Will file a ticket for this and reference in the comment

nbbeeken (approved):

PR #637, opened 2023-11-14
fix(NODE-3034): ObjectId.isValid() returns false if a number is passed

lifeisfoo:
#### What is changing?

- Change `ObjectId.isValid()` to fail when a number is passed as an argument

No.

`ObjectId.isValid(1)` returns `true` but this is wrong, since _numbers_ aren't valid ObjectIds.
The improper behaviour has been added with [this commit](https://github.com/mongodb/js-bson/commit/cd132d9b8498d1a10f93a011223758f340f3a466) on May 20, 2015.
Since then the `ObjectId.isValid()` function has been refactored a couple of times but the behaviour is still the same.
The last major refactor occurred on [Dec 2, 2021](https://github.com/mongodb/js-bson/commit/187d1c40f7222fd63dbc55d1535669ff0bfcbce2#diff-18c447fb56b40e1238fd00fc5473e2994d33a1f4618c2bdb71a3af1759c599abL293) where all the checks have been replaced with a try-catched call to the ObjectId constructor, but the programmer was misled by the existing (wrong) support for `number`.

`ObjectId.isValid()` tests with numbers were added with [this commit](https://github.com/mongodb/js-bson/commit/409c592524a6eec6a89eb96dee4b51e4f8adf5ae#diff-dd8a26a1854c387c9ca9fa21c8edbb62d38d81e51f96e4a76335fd023b1024baR346) on Aug 1, 2023 during a fix to the ObjectId constructor (perhaps to improve coverage).

This bug is tracked by [NODE-3034](https://jira.mongodb.org/browse/NODE-3034).

### DIFF
--- a/src/objectid.ts
+++ b/src/objectid.ts
@@ -269,8 +269,10 @@ export class ObjectId extends BSONValue {
    * Checks if a value can be used to create a valid bson ObjectId
    * @param id - any JS value
    */
-  static isValid(id: string | number | ObjectId | ObjectIdLike | Uint8Array): boolean {
+  static isValid(id: string | ObjectId | ObjectIdLike | Uint8Array): boolean {
     if (id == null) return false;
+    // ObjectId constructor accepts a number (seconds) but a number is not a valid ObjectId
+    if (typeof id === 'number') return false;
 
     try {
       new ObjectId(id);
--- a/test/node/bson_test.js
+++ b/test/node/bson_test.js
@@ -1696,7 +1696,6 @@ describe('BSON', function () {
     expect(false).to.equal(ObjectId.isValid({ length: 12 }));
     expect(false).to.equal(ObjectId.isValid([]));
     expect(false).to.equal(ObjectId.isValid(true));
-    expect(true).to.equal(ObjectId.isValid(0));
     expect(false).to.equal(ObjectId.isValid('invalid'));
     expect(false).to.equal(ObjectId.isValid('zzzzzzzzzzzzzzzzzzzzzzzz'));
     expect(true).to.equal(ObjectId.isValid('000000000000000000000000'));
--- a/test/node/object_id.test.ts
+++ b/test/node/object_id.test.ts
@@ -344,12 +344,12 @@ describe('ObjectId', function () {
     });
 
     context('when called with a number', () => {
-      it('returns true for any number (0)', () => expect(ObjectId.isValid(0)).to.be.true);
+      it('returns false for any number (0)', () => expect(ObjectId.isValid(0)).to.be.false);
 
-      it('returns true for any number (MIN_SAFE_INTEGER)', () =>
-        expect(ObjectId.isValid(Number.MIN_SAFE_INTEGER)).to.be.true);
+      it('returns false for any number (MIN_SAFE_INTEGER)', () =>
+        expect(ObjectId.isValid(Number.MIN_SAFE_INTEGER)).to.be.false);
 
-      it('returns true for any number (NaN)', () => expect(ObjectId.isValid(NaN)).to.be.true);
+      it('returns false for any number (NaN)', () => expect(ObjectId.isValid(NaN)).to.be.false);
     });
 
     context('when called with a ObjectId or ObjectIdLike', () => {


aditi-khare-mongoDB:
Hi @lifeisfoo, thank you for your submission! The team will look into it within the coming week.

nbbeeken:
Hi @lifeisfoo,

The `isValid` method "Checks if a value can be used to create a valid bson ObjectId" so a number returning true is accurate for the intent of the function. Since the ObjectId constructor validates its input and can throw, the isValid method provides a way to pre-check unknown inputs before attempting construction. 

Could you share what your intended use case is for this method? And how does it relate to returning true for number that does not apply to `Uint8Array` or `number[]`?

lifeisfoo:
Hi @nbbeeken,
I was so focused on the code that I overlooked the comment above the function.

I will try to explain it in the best way possibile, so it will be a bit long.
The ObjectId is a [12 bytes sequence](https://github.com/mongodb/specifications/blob/master/source/objectid.rst) and it can be represented as:
- an ObjectId (type) with its internal Uint8Array
- an ObjectIdLike (type) with its internal Uint8Array OR 
- a 24 char string (2 char per byte to represent)
- an Uint8Array (you can map a single byte into an int value => an array of 12 int is an ObjectId)

When you pass one of such values to the `ObjectId.isValid()` function or to the ObjectId constructor the result will be the "same": a successful validation in the first case and a new ObjectId in the second.
The ObjectId created, with the constructor, from one of such values is a 1-1 copy of the ObjectId represented by the value used.
So the results of the validation function and the constructor are consistent.

But when you validate a number, or use it as a value for the constructor it's not the same story: a number cannot be a valid ObjectId.
In that case the constructor will use it as the value for the 4-bytes timestamp section of the ObjectId.

      4 byte timestamp    5 byte process unique   3 byte counter
    |<----------------->|<---------------------->|<------------>|
    [----|----|----|----|----|----|----|----|----|----|----|----]
    0                   4                   8                   12

Indeed, the constructor code calls the `generate` function

      this[kId] = ObjectId.generate(typeof workingId === 'number' ? workingId : undefined);

That uses it as expected, just to populate the first 4 bytes of the object.
```
  /**
   * Generate a 12 byte id buffer used in ObjectId's
   *
   */
  static generate(time?: number): Uint8Array {
    if ('number' !== typeof time) {
      time = Math.floor(Date.now() / 1000);
    }
...
```

> Numbers can't be ObjectIds but just an argument to generate them.

When you see a static function named `ObjectId.isValid()`, it can have only one meaning, the original one: "Checks if a value is a valid bson ObjectId".

This function is necessary when I receive an `ObjectId` from somewhere (e.g. an API, a message, at the end of a computation, ...) and I want to check if it's a valid OID before building it or using it in a query.

[This commit of Aug 1 2023 changed](https://github.com/mongodb/js-bson/commit/409c592524a6eec6a89eb96dee4b51e4f8adf5ae#diff-18c447fb56b40e1238fd00fc5473e2994d33a1f4618c2bdb71a3af1759c599abL284 ) the documentation of this method from the original one.

Other BSON libraries have similar functions to check if a value is a valid ObjectId:
- [ruby](https://github.com/mongodb/bson-ruby/blob/master/lib/bson/object_id.rb#L336)
- [go](https://github.com/mongodb/mongo-go-driver/blob/v1/bson/primitive/objectid.go#L97)
- [python](https://github.com/mongodb/mongo-python-driver/blob/master/bson/objectid.py#L141)
- [java](https://github.com/mongodb/mongo-java-driver/blob/master/bson/src/main/org/bson/types/ObjectId.java#L114)
As you can see they always check if the value represents a complete ObjectId, so there are no numbers able to pass the validation.

The `createFromHexString()` function cannot be a replacement for `ObjectId.isValid()` because it accepts strings only and it checks only their length.

If you really want a static function to validate the constructor why don't call it `ObjectId.isValidForConstructor()`?
But in that case, how could I validate an existing ObjectId or a Uint8Array?

nbbeeken:
Thank you for the elaborate detail @lifeisfoo! That captured a lot of context, I really appreciate all the links.

We are not going to go forward with the change proposed here, instead, we plan to remove the ObjectId constructor's support for numbers in the next major version, which will also make isValid no longer return true for that input type. I have updated [NODE-3034](https://jira.mongodb.org/browse/NODE-3034) and the follow-up [NODE-5763](https://jira.mongodb.org/browse/NODE-5763) with more details.

lifeisfoo:
Less code, less bugs :slightly_smiling_face: .

PR #640, opened 2024-01-25
feat(NODE-3034): deprecate number as an input to ObjectId constructor

alenakhineika:
### Description
Deprecate `number` as an input to `ObjectId` constructor.

#### What is changing?
Create separate overloads for each input type of the ObjectID constructor and deprecate one that accepts a numeric input value.

None

[NODE-3034](https://jira.mongodb.org/browse/NODE-3034)

### A number as an input to the `ObjectId` constructor is deprecated
Instead, use `static createFromTime()` to set a numeric value for the new ObjectId.

### DIFF
--- a/src/objectid.ts
+++ b/src/objectid.ts
@@ -44,9 +44,46 @@ export class ObjectId extends BSONValue {
   private __id?: string;
 
   /**
-   * Create an ObjectId type
+   * Create ObjectId from a number.
    *
-   * @param inputId - Can be a 24 character hex string, 12 byte binary Buffer, or a number.
+   * @param inputId - A number.
+   * @deprecated Instead, use `static createFromTime()` to set a numeric value for the new ObjectId.
+   */
+  constructor(inputId: number);
+  /**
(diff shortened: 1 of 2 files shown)


src/objectid.ts
```diff
@@ -44,9 +44,46 @@ export class ObjectId extends BSONValue {
...
+   * Implementation overload.
+   *
+   * @param inputId - All input types that are used in the constructor implementation.
+   */
+  constructor(inputId?: string | number | ObjectId | ObjectIdLike | Uint8Array);
+  /**
+   * Create a new ObjectId.
+   *
+   * @param inputId - An input value to create a new ObjectId from.
    */
```
nbbeeken, line 87:
```
  /**
   * Generate a new ObjectId.
   */
  constructor();
```

I think breaking out the empty ctor would be a nice addition as well, then we can mention that's how you generate a new oid.

nbbeeken (changes requested):

nbbeeken (approved):

PR #642, opened 2024-01-29
feat(NODE-5861): optimize parsing basic latin strings

nbbeeken:
#### What is changing?

- BSON's UTF8 parsing logic will attempt to use `String.fromCharCode` if bytes are under 128
- Move UTF8 validation after attempting the above, if bytes are below 128 there is nothing to validate

- No

This has shown to have much better performance for strings that fit in the basic latin range. Quite often this is BSON keys.

### BSON short basic latin string parsing performance improved!

The BSON library's string decoding logic now attempts to optimize for basic latin (ASCII) characters. This will apply to both BSON keys and BSON values that are or contain strings. If strings are less than 6 bytes we observed approximately ~100% increase in speed while around 15 bytes the performance was about ~30% better. For any non-basic latin bytes or at 20 bytes or greater the BSON library will continue to use Node.js' Buffer.toString API.

The intent is to generally target the deserialization of BSON keys which are often short and only use basic latin, <i>Et tu, _id?</i>

### DIFF
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -13,7 +13,7 @@ const tsConfig = {
   module: 'esnext',
   moduleResolution: 'node',
   removeComments: true,
-  lib: ['es2021'],
+  lib: ['es2021', 'ES2022.Error'],
   importHelpers: false,
   noEmitHelpers: false,
   noEmitOnError: true,
--- a/src/utils/latin.ts
+++ b/src/utils/latin.ts
@@ -1,3 +1,5 @@
+/* eslint-disable prefer-spread */
+
 /**
  * This function is an optimization for small basic latin strings.
  * @internal
@@ -57,7 +59,6 @@ export function tryLatin(uint8array: Uint8Array, start: number, end: number): st
     }
 
     if (basicLatin) {
-      // eslint-disable-next-line prefer-spread
       return String.fromCharCode.apply(String, latinBytes);
     }
   }
--- a/src/utils/latin.ts
+++ b/src/utils/latin.ts
@@ -1,6 +1,6 @@
 /**
  * This function is an optimization for small basic latin strings.
- *
+ * @internal
  * @remarks
  * ### Important characteristics:
  * - If the uint8array or distance between start and end is 0 this function returns an empty string
--- a/src/utils/latin.ts
+++ b/src/utils/latin.ts
@@ -23,6 +23,10 @@ export function tryLatin(uint8array: Uint8Array, start: number, end: number): st
     return '';
   }
 
+  if (stringByteLength > 20) {
+    return null;
+  }
+
   if (stringByteLength === 1 && uint8array[start] < 128) {
     return String.fromCharCode(uint8array[start]);
   }
@@ -44,22 +48,14 @@ export function tryLatin(uint8array: Uint8Array, start: number, end: number): st
     );
   }
 
-  if (stringByteLength <= 20) {
-    let basicLatin = true;
-    const latinBytes = [];
-    for (let i = start; i < end; i++) {
-      const byte = uint8array[i];
-      if (byte > 127) {
-        basicLatin = false;
-        break;
-      }
-      latinBytes.push(byte);
+  const latinBytes = [];
+  for (let i = start; i < end; i++) {
+    const byte = uint8array[i];
+    if (byte > 127) {
+      return null;
     }
-
-    if (basicLatin) {
-      return String.fromCharCode(...latinBytes);
-    }
+    latinBytes.push(byte);
   }
 
-  return null;
+  return String.fromCharCode(...latinBytes);
 }
--- a/src/utils/latin.ts
+++ b/src/utils/latin.ts
@@ -1,5 +1,3 @@
-/* eslint-disable prefer-spread */
-
 /**
  * This function is an optimization for small basic latin strings.
  * @internal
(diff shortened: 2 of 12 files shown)


src/utils/web_byte_utils.ts
```diff
@@ -172,8 +172,42 @@ export const webByteUtils = {
...
+        if (byte > 127) {
+          basicLatin = false;
+          break;
+        }
+        latinBytes.push(byte);
+      }
+
+      if (basicLatin) {
+        // eslint-disable-next-line prefer-spread
+        return String.fromCharCode.apply(String, latinBytes);
```
addaleax, line 199: I'd be pretty curious to see how this compares against
```suggestion
        return String.fromCharCode.apply(String, uint8array);
```

or

```suggestion
        return String.fromCharCode.apply(String, Array.from(uint8array));
```

as creating an Array and then repeatedly `.push()`ing into it can also be quite a bit slower than using a pre-generated or fixed-length array

(I can also try myself at some point but you probably have the benchmarking setup working already)

nbbeeken: I attached benchmarks to the ticket, the take away is that pushing to an array is faster than pre-allocating it or just assigning to the next index of an empty array (`toUTF8_assign_array`). I would guess this has something to do with creating an array with empty slots vs no slots and filling it with a consistent type (always numbers). 

Just guessing though, it is revealed that attaching strings via concat is super fast for small cases, so I've hard-coded 1, 2, and 3 length strings.

src/utils/latin.ts
```diff
@@ -0,0 +1,66 @@
...
+    return '';
+  }
+
+  const stringByteLength = end - start;
+  if (stringByteLength === 0) {
+    return '';
+  }
+
+  if (stringByteLength === 1 && uint8array[start] < 128) {
+    return String.fromCharCode(uint8array[start]);
```
addaleax, line 27: Kind of raises the question, would adding a

```js
const basicLatinChars = [...Array(128).keys()].map(i => String.fromCharCode(i));
```

at the top of the file and then using 

```suggestion
    return basicLatinChars[uint8array[start]];
```

in place of `String.fromCharCode()` improve things a bit more even?

addaleax (approved):
🚀

src/utils/latin.ts
```diff
@@ -0,0 +1,67 @@
...
+      const byte = uint8array[i];
+      if (byte > 127) {
+        basicLatin = false;
+        break;
+      }
+      latinBytes.push(byte);
+    }
+
+    if (basicLatin) {
+      return String.fromCharCode.apply(String, latinBytes);
```
durran, line 62: Just a question, why are we calling `apply` here and not just `fromCharCode` directly?

rollup.config.mjs
```diff
@@ -13,7 +13,7 @@ const tsConfig = {
   module: 'esnext',
   moduleResolution: 'node',
   removeComments: true,
-  lib: ['es2021'],
+  lib: ['es2021', 'ES2022.Error'],
```
durran, line 16: This seems unrelated to the ticket, just curious as to the reasoning for this change.

src/utils/latin.ts
nbbeeken: Good question, I realize I hadn't actually tested, I assumed that spreading would be worse for performance than apply, figured that spreading might make a copy of the array. I tested and found the opposite, looks like spreading is ever so slightly better, I don't think it makes much difference in performance but it is certainly nicer on the eyes.

```txt
stringLen: 7
toUTF8_try_latin x 13,747,586 ops/sec ±1.06% (92 runs sampled)
toUTF8_try_latin_spread x 14,469,017 ops/sec ±0.29% (93 runs sampled)

stringLen: 9
toUTF8_try_latin x 12,610,412 ops/sec ±0.45% (93 runs sampled)
toUTF8_try_latin_spread x 13,128,721 ops/sec ±0.25% (96 runs sampled)

stringLen: 16
toUTF8_try_latin x 9,209,619 ops/sec ±0.18% (95 runs sampled)
toUTF8_try_latin_spread x 9,485,863 ops/sec ±0.20% (96 runs sampled)
```

rollup.config.mjs
nbbeeken: As a part of this work we were able to move UTF validation below the latin optimization (because below below 127 cannot be in an invalid format). On the webByteUtils side, validation is done by TextDecoder, but we have an API expectation that we throw a BSONError, but I figured it would be nice to maintain the original error from the TextDecoder. So I added support for the cause property on BSONError, TS doesn't allow us to pass a second argument to the Error constructor without adding this argument to the lib setting. 

This doesn't have downstream effects, we only support versions of Node.js that support the cause argument, but if this was run on old versions, it would just be ignored.

durran (dismissed):

src/utils/latin.ts
```diff
@@ -0,0 +1,65 @@
...
+      String.fromCharCode(uint8array[start + 2])
+    );
+  }
+
+  if (stringByteLength <= 20) {
+    let basicLatin = true;
+    const latinBytes = [];
+    for (let i = start; i < end; i++) {
+      const byte = uint8array[i];
+      if (byte > 127) {
```
baileympearson, line 52: just a suggestion: why not simply return if a byte is >127, which would remove the need for `basicLatin`?

baileympearson (commented):
just one small suggestion, otherwise looks good

src/utils/latin.ts
nbbeeken: SGTM! Previously this logic was inlined in toUTF8, so break+bool was used to fall through to Buffer.toString, but good call we can remove some indentation here and exit earlier on more than 20 bytes.

baileympearson (approved):

durran (approved):

PR #643, opened 2024-01-31
fix: use Symbol.for('id') in ObjectId to fix errors when mixing cjs and mjs

dot-i:
### Description

Fix interoperability of `bson` between ESM and CommonJS.

#### What is changing?

`Symbol('id')` is replaced with `Symbol.for('id')` in `objectid.ts`. This avoids the issue of two different symbols being used for ESM and CommonJS, causing `ObjectId.equals()` to fail with an `undefined` error.

No.

We are switching our modules to be ESM and ran in to this issue because we imported `bson` directly from our module (ESM), but also sometimes through a CommonJS import which in turn also imported `bson` -- but as CommonJS module.

This caused `ObjectId.equals()` to fail since there were two separate `'id'` symbols in use, causing an `undefined` error.


src/objectid.ts
```diff
@@ -21,7 +21,7 @@ export interface ObjectIdExtended {
   $oid: string;
 }
 
-const kId = Symbol('id');
+const kId = Symbol.for('id');
```
addaleax, line 24: `Symbol.for` uses a global namespace for values, so you probably want to make the key here more specific than just `'id'`

dot-i: Good point, any preference according to naming conventions?  I could use `@@bson.objectid.id`...

dot-i: I picked the second one ;-)  If it's not OK, can change...

durran: I think we'd probably prefer the former `@@bson.objectid.id` since BSON is technically it's own library and doesn't necessarily need to be used the MongoDB.

durran (changes requested):
@dot-i can you also run the tests as well and fix any failures? Example:

```
[2024/01/31 21:04:34.125] > npm run check:node && npm run check:web && npm run check:web-no-bigint
[2024/01/31 21:04:34.903] created lib/bson.rn.cjs in 654ms
[2024/01/31 21:04:34.903] { web: false, noBigInt: false, nodeVersion: 'v20.11.0' }
[2024/01/31 21:04:36.187] Error: Did not find Symbol(id) on 6b61666665656b6c61746368
[2024/01/31 21:04:36.187]     at getSymbolFrom (/data/mci/db7d87dd5c732b6aca749a47f85e508c/src/test/node/tools/utils.js:200:11)
[2024/01/31 21:04:36.187]     at Suite.<anonymous> (/data/mci/db7d87dd5c732b6aca749a47f85e508c/src/test/node/object_id.test.ts:379:33)
```

src/objectid.ts
dot-i: Changed to `@@bson.objectid.id`.

dot-i:
> @dot-i can you also run the tests as well and fix any failures?

@durran Fixed!

nbbeeken:
@durran @addaleax Should we consider changing the symbol property to a string one? I don't see a motive for it anymore, combined with registering globally it doesn't seem to simulate "private" any more than typescript `private` does.

nbbeeken:
Apparently, I was the one who introduced it 😅 https://github.com/mongodb/js-bson/pull/394. 

At the time we were converting from JS to TS and in JS we felt that class property "privacy" was best encoded by attaching things as symbols and I was just following the pattern started before me in the changes necessary for that PR. Back then in the ancient era of ~3.5yrs ago, we didn't need to think of dual module hazards. 

_I think_ to solve the ESM and cross-version issue we have today there's nothing wrong with going back to a string property and using Typescript to indicate privacy.

This is the only Symbol property we have in BSON, so it is already an odd one out in that sense.

dot-i:
@nbbeeken what string property name would you propose?

```js
  get id(): Uint8Array {
    return this[kId];
  }
```

Since there's this getter, a simple `'id'` is not possible. And there's already a `__id` property for the hexString cache.

nbbeeken:
How about `buffer`? taking the idea from our Binary type, but it should be private here on the ObjectId.

dot-i:
> How about `buffer`? taking the idea from our Binary type, but it should be private here on the ObjectId.

Works for me 👍

Do I update the PR to use a `#buffer` property instead of the symbol?

nbbeeken:
Just `private buffer` works. And yes, that is exactly for the reason you call out, at this time we're avoiding using `#`private variables due to the overhead.

dot-i:
Changed to use `private buffer`, adjusted unit test accordingly.

nbbeeken (approved):
LGTM, thanks so much for the help here!

PR #645, opened 2024-02-08
feat(NODE-5909): optimize writing basic latin strings

nbbeeken:
#### What is changing?

This is the inverse opimization made in: https://github.com/mongodb/js-bson/pull/642

Should have thought of it sooner :)

- Removed fromUTF8 dead code, we always call encodeInto which is BYOB (bring your own buffer 🍻)

No

When strings are short and are only basic latin, it is faster to iterate length of the string and call charCodeAt to obtain the UTF8 byte value.

### BSON short basic latin string writing performance improved!

The BSON library's string encoding logic now attempts to optimize for basic latin (ASCII) characters. This will apply to both BSON keys and BSON values that are or contain strings. If strings are less than 6 bytes we observed approximately 100% increase in speed while around 24 bytes the performance was about 33% better. For any non-basic latin bytes or at 25 bytes or greater the BSON library will continue to use Node.js' Buffer.toString API.

The intent is to generally target the serialization of BSON keys which are often short and only use basic latin.

### DIFF
--- a/src/utils/latin.ts
+++ b/src/utils/latin.ts
@@ -59,3 +59,42 @@ export function tryLatin(uint8array: Uint8Array, start: number, end: number): st
 
   return String.fromCharCode(...latinBytes);
 }
+
+/**
+ * This function is an optimization for writing small basic latin strings.
+ * @internal
+ * @remarks
+ * ### Important characteristics:
+ * - If the string length is 0 return 0, do not perform any work
+ * - If a string is longer than 25 code units return null
+ * - If any code unit exceeds 128 this function returns null
+ *
+ * @param destination - The uint8array to serialize the string to
+ * @param source - The string to turn into UTF-8 bytes if it fits in the basic latin range
+ * @param offset - The position in the destination to begin writing bytes to
+ * @returns the number of bytes written to destination if all code units are below 128, otherwise null
+ */
+export function tryWriteLatin(
+  destination: Uint8Array,
+  source: string,
+  offset: number
+): number | null {
+  if (source.length === 0) return 0;
+
+  if (source.length > 25) return null;
+
+  if (destination.length - offset < source.length) return null;
+
+  for (
+    let charOffset = 0, destinationOffset = offset;
+    charOffset < source.length;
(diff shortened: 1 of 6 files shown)


nbbeeken:
[All the BSON granular benchmarks run with this change](https://docs.google.com/spreadsheets/d/1gRaiQGJYDnxh4TRn6HNm0KxBQkyEZBAmKEeGEPXJpuI/edit#gid=1155849188)

src/utils/latin.ts
```diff
@@ -59,3 +59,42 @@ export function tryLatin(uint8array: Uint8Array, start: number, end: number): st
...
+ * - If the string length is 0 return 0, do not perform any work
+ * - If a string is longer than 25 code units return null
+ * - If any code unit exceeds 128 this function returns null
+ *
+ * @param destination - The uint8array to serialize the string to
+ * @param source - The string to turn into UTF-8 bytes if it fits in the basic latin range
+ * @param offset - The position in the destination to begin writing bytes to
+ * @returns the number of bytes written to destination if all code units are below 128, otherwise null
+ */
+export function tryWriteLatin(
```
addaleax, line 77: This feels a bit silly, but how about naming the other method `tryReadLatin` then? (Or maybe something like `tryDecodeASCII`/`tryDecodeBasicLatin` and then this one `tryEncodeASCII`/`tryEncodeBasicLatin` – these methods don't handle Latin in the sense of a ISO-8859 character set, they handle Basic Latin aka ASCII only, so why not be precise about that?) Feel free to ignore!

addaleax (approved):

src/utils/latin.ts
nbbeeken: I agree, happy to update the naming to be clear, might as well!

addaleax (approved):

src/utils/latin.ts
nbbeeken: I have been so intentional throughout these two PRs to use the term "basic latin" and not ASCII, except for in the release notes, because as you say, why not be precise 😅, so yea anyway, I agree with seeing that completely through

PR #647, opened 2024-02-09
test(NODE-5922): replace hardcoded reference to bson#main

W-A-James:
#### What is changing?
Replaced hardcoded reference to `bson#main` when running pre-benchmark setup for `run_granular_benchmarks.js`

No

NODE-5922

### DIFF
--- a/test/bench/etc/run_granular_benchmarks.js
+++ b/test/bench/etc/run_granular_benchmarks.js
@@ -9,7 +9,7 @@ const fs = require('fs/promises');
 const path = require('path');
 const { once } = require('events');
 const { Task } = require('dbx-js-tools/packages/bson-bench');
-const { LIBRARY_SPEC } = require('../granular/common');
+const { LIBRARY_SPEC } = require('../lib/granular/common');
 
 const BENCHMARK_REGEX = /(.*)\.bench\.js$/;
 const BENCHMARK_PATH = path.resolve(`${__dirname}/../lib/granular`);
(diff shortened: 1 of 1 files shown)


test/bench/etc/run_granular_benchmarks.js
```diff
@@ -18,7 +19,7 @@ const DOCUMENT_ROOT = path.resolve(`${__dirname}/../documents`);
   // FIXME(NODE-5759): replace this with a call to Package.install()
   await new Task({
     documentPath: path.resolve(`${DOCUMENT_ROOT}/binary_small.json`),
-    library: `bson#main`,
+    library: LIBRARY_SPEC,
```
nbbeeken, line 22: Works for me, but is there a way we can remove this task from the final output since it isn't "real" in a way

nbbeeken (changes requested):

test/bench/etc/run_granular_benchmarks.js
W-A-James: It isn't a part of the output. This task runs and fails in CI since the targeted library isn't yet installed when run on a fresh system. Even if it were to run to completion, the results are not collected or written.

W-A-James: All of the actual benchmarks are run and have results collected and written using the `runSuiteAndWriteResults` helper in `test/bench/granular/common.ts`.

nbbeeken: My mistake, I thought I was seeing small binary twice, it was with/without promoteBuffers

nbbeeken (approved):

PR #656, opened 2024-03-08
feat(NODE-5958): add BSON iterating API

nbbeeken:
#### What is changing?

- Added an API that will iterate embedded document structures in BSON

No.

To encapsulate the complexity of iterating into nested structures such that one can write a function responsible for reviving each BSON value as needed and decide whether or not to create a nested structure.

### DIFF
--- a/src/parser/on_demand/parse_to_elements.ts
+++ b/src/parser/on_demand/parse_to_elements.ts
@@ -44,7 +44,10 @@ export type BSONElement = [
   length: number
 ];
 
-/** @internal Parses a int32 little-endian at offset, throws if it is negative */
+/**
+ * @internal
+ * Parses a int32 little-endian at offset, throws if it is negative
+ */
 export function getSize(source: Uint8Array, offset: number): number {
   if (source[offset + 3] > 127) {
     throw new BSONOffsetError('BSON size cannot be negative', offset);
--- a/src/parser/on_demand/parse_to_elements.ts
+++ b/src/parser/on_demand/parse_to_elements.ts
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/no-unsafe-enum-comparison */
 import { BSONOffsetError } from '../../error';
 
 /**
@@ -45,8 +44,8 @@ export type BSONElement = [
   length: number
 ];
 
-/** Parses a int32 little-endian at offset, throws if it is negative */
-function getSize(source: Uint8Array, offset: number): number {
+/** @internal Parses a int32 little-endian at offset, throws if it is negative */
+export function getSize(source: Uint8Array, offset: number): number {
   if (source[offset + 3] > 127) {
     throw new BSONOffsetError('BSON size cannot be negative', offset);
   }
--- a/src/parser/on_demand/parse_to_elements.ts
+++ b/src/parser/on_demand/parse_to_elements.ts
@@ -8,7 +8,7 @@ import { BSONOffsetError } from '../../error';
  * - `minKey` is set to 255 so unsigned comparisons succeed
  * - Modify with caution, double check the bundle contains literals
  */
-const enum t {
+const enum BSONElementType {
   double = 1,
   string = 2,
   object = 3,
@@ -82,8 +82,11 @@ function findNull(bytes: Uint8Array, offset: number): number {
  * @public
  * @experimental
  */
-export function parseToElements(bytes: Uint8Array, pOffset?: number | null): Iterable<BSONElement> {
-  const startOffset = pOffset ?? 0;
+export function parseToElements(
+  bytes: Uint8Array,
+  startOffset: number | null = 0
+): Iterable<BSONElement> {
+  startOffset ??= 0;
 
   if (bytes.length < 5) {
     throw new BSONOffsetError(
@@ -125,37 +128,51 @@ export function parseToElements(bytes: Uint8Array, pOffset?: number | null): Ite
 
     let length: number;
 
-    if (type === t.double || type === t.long || type === t.date || type === t.timestamp) {
+    if (
+      type === BSONElementType.double ||
+      type === BSONElementType.long ||
+      type === BSONElementType.date ||
+      type === BSONElementType.timestamp
+    ) {
       length = 8;
-    } else if (type === t.int) {
+    } else if (type === BSONElementType.int) {
       length = 4;
-    } else if (type === t.objectId) {
+    } else if (type === BSONElementType.objectId) {
       length = 12;
-    } else if (type === t.decimal) {
+    } else if (type === BSONElementType.decimal) {
       length = 16;
-    } else if (type === t.bool) {
+    } else if (type === BSONElementType.bool) {
       length = 1;
-    } else if (type === t.null || type === t.undefined || type === t.maxKey || type === t.minKey) {
+    } else if (
+      type === BSONElementType.null ||
+      type === BSONElementType.undefined ||
+      type === BSONElementType.maxKey ||
+      type === BSONElementType.minKey
+    ) {
       length = 0;
     }
     // Needs a size calculation
-    else if (type === t.regex) {
+    else if (type === BSONElementType.regex) {
       length = findNull(bytes, findNull(bytes, offset) + 1) + 1 - offset;
-    } else if (type === t.object || type === t.array || type === t.javascriptWithScope) {
+    } else if (
+      type === BSONElementType.object ||
+      type === BSONElementType.array ||
+      type === BSONElementType.javascriptWithScope
+    ) {
       length = getSize(bytes, offset);
     } else if (
-      type === t.string ||
-      type === t.binData ||
-      type === t.dbPointer ||
-      type === t.javascript ||
-      type === t.symbol
+      type === BSONElementType.string ||
+      type === BSONElementType.binData ||
+      type === BSONElementType.dbPointer ||
+      type === BSONElementType.javascript ||
+      type === BSONElementType.symbol
     ) {
       length = getSize(bytes, offset) + 4;
-      if (type === t.binData) {
+      if (type === BSONElementType.binData) {
         // binary subtype
         length += 1;
       }
-      if (type === t.dbPointer) {
+      if (type === BSONElementType.dbPointer) {
         // dbPointer's objectId
         length += 12;
       }
--- a/src/parser/on_demand/parse_to_elements.ts
+++ b/src/parser/on_demand/parse_to_elements.ts
@@ -82,7 +82,9 @@ function findNull(bytes: Uint8Array, offset: number): number {
  * @public
  * @experimental
  */
-export function parseToElements(bytes: Uint8Array, startOffset = 0): Iterable<BSONElement> {
+export function parseToElements(bytes: Uint8Array, pOffset?: number | null): Iterable<BSONElement> {
+  const startOffset = pOffset ?? 0;
+
   if (bytes.length < 5) {
     throw new BSONOffsetError(
       `Input must be at least 5 bytes, got ${bytes.length} bytes`,
--- a/src/parser/on_demand/parse_to_structure.ts
+++ b/src/parser/on_demand/parse_to_structure.ts
@@ -67,6 +67,8 @@ export type Container =
   | {
       kind: 'custom';
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      dest: any;
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
       [key: string]: any;
     };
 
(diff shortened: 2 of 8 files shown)


test/node/tools/utils.js
```diff
@@ -166,12 +166,12 @@ exports.int32LEToHex = int32LEToHex;
...
-  var b = Buffer.from(str, 'utf8');
-  var len = b.byteLength;
-  var out = Buffer.alloc(len + 4 + 1);
+  const b = Buffer.from(str, 'utf8');
+  const len = b.byteLength;
+  const out = Buffer.alloc(len + 4 + 1);
   out.writeInt32LE(len + 1, 0);
   out.set(b, 4);
-  out[len + 1] = 0x00;
+  out[len + 4 + 1] = 0x00;
```
nbbeeken, line 174: 😳

src/parser/on_demand/parse_to_structure.ts
```diff
@@ -0,0 +1,138 @@
...
+          previous: ctx
+        };
+
+        continue embedded;
+      }
+    }
+    ctx = ctx.previous;
+  }
+
+  return root.dest as unknown as TRoot;
```
aditi-khare-mongoDB, line 137: Is there a way to work around this type cast? like if we explicitly delcare `dest `as type `TRoot`?

src/parser/on_demand/parse_to_structure.ts
```diff
@@ -0,0 +1,138 @@
...
+  nameLength = 2,
+  offset = 3,
+  length = 4
+}
+
+/**
+ * @internal
+ * Embedded bson types
+ */
+const enum t {
```
aditi-khare-mongoDB, line 29: Even thought there are API docs, can we have more descriptive names for `e` and `t`, something like `elementOffsets` and `embeddedTypes`, respectively. Especially since the embedded bson types const could easily be confused for `e.`

src/parser/on_demand/parse_to_structure.ts
```diff
@@ -0,0 +1,138 @@
+import { type Code } from '../../code';
+import { type BSONElement, getSize, parseToElements as p } from './parse_to_elements';
+
+/** @internal TODO */
+const DEFAULT_REVIVER = () => null;
+
+/** @internal */
+function parseToElements(...args: Parameters<typeof p>): BSONElement[] {
```
aditi-khare-mongoDB, line 8: What's the logic behind having a parseToElements wrapper in parse_to_structure file with the same name as parseToElements in the ./parse_to_elements file? The naming seems a bit confusing to me at first glance.

aditi-khare-mongoDB (changes requested):

src/parser/on_demand/parse_to_elements.ts
```diff
@@ -45,8 +44,8 @@ export type BSONElement = [
   length: number
 ];
 
-/** Parses a int32 little-endian at offset, throws if it is negative */
-function getSize(source: Uint8Array, offset: number): number {
+/** @internal Parses a int32 little-endian at offset, throws if it is negative */
```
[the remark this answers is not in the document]
nbbeeken: There's nothing wrong with this formatting from the TSDoc perspective, occassionally if the docs are short enough I prefer the one liner. But I do not mind changing it, done.

src/parser/on_demand/parse_to_structure.ts
nbbeeken: Yea I think I've come up with something by using the type that root is set to in order to remove this.

src/parser/on_demand/parse_to_structure.ts
nbbeeken: I would like to keep these as single characters as we did in parseToElements. I think the context matters, `it[t.object]` does not make much sense for what `it` is in the context of the function below. If this was exported for use elsewhere in the library then I think you would want the variables to carry more context in their names. I moved them closer to the point in which they are used to get better association with the code that uses them. Is that a good enough way to improve the meaning?

src/parser/on_demand/parse_to_structure.ts
nbbeeken: Fixed, just a stylistic choice to not create something with a new name for a small piece of logic. This works too.

test/node/tools/utils.js
nbbeeken: It looks like we weren't using this anywhere except the tests I've just written 😅

src/parser/on_demand/parse_to_elements.ts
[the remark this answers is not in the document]
aditi-khare-mongoDB: Thanks!

src/parser/on_demand/parse_to_structure.ts
aditi-khare-mongoDB: Ah I see, sure, in the case that we ever do export them, we can change their names then.

aditi-khare-mongoDB (dismissed):

src/parser/on_demand/parse_to_structure.ts
```diff
@@ -0,0 +1,143 @@
...
+      dest: Code;
+      kind: 'code';
+    }
+  | {
+      kind: 'custom';
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      dest: any;
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      [key: string]: any;
+    };
```
baileympearson, line 55: Should this be parameterized instead of relying on any?

Something like this might work:
```typescript
export type Container<TContainer = unknown> = 
  | {
      dest: Record<string, unknown>;
      kind: 'object';
    }
  | {
      dest: Map<string, unknown>;
      kind: 'map';
    }
  | {
      dest: Array<unknown>;
      kind: 'array';
    }
  | {
      dest: Code;
      kind: 'code';
    }
  | {
      kind: 'custom';
      dest: TContainer
    };
```

baileympearson (changes requested):

src/parser/on_demand/parse_to_structure.ts
aditi-khare-mongoDB: @baileympearson It seems like you have a similar suggestion as I did, here's context for the discussion we had before.

baileympearson: I don't see an occurrence of `it[t.object]` in the code.  I do see the following:

```typescript
      const isEmbeddedType =
        type === t.object || type === t.array || type === t.javascriptWithScope;
```

Which makes sense with `EmbeddedTypes` instead of `t`.

I see something similar for `e`:
```typescript
      const type = it[e.type];
      const offset = it[e.offset];
```
but isn't that more clear if we just give it a good name?
```typescript
      const type = it[RawBSONElementIndexes.type];
      const offset = it[RawBSONElementIndexes.offset];
```

---

But more generally - I don't understand the arguments against descriptive variable names.  If we merged `parseToElements` that way, I suggest we fix it instead of forging ahead with unhelpful variable names.  

What downside is there to using good variable names?  Generally the only argument I hear against them is that it increases code size, but anyone in the JS community who cares about this bundles their code and has symbol minification anyways.

src/parser/on_demand/parse_to_structure.ts
```diff
@@ -0,0 +1,143 @@
...
+export function parseToStructure<
+  TRoot extends Container = {
+    dest: Record<string, unknown>;
+    kind: 'object';
+  }
+>(
+  bytes: Uint8Array,
+  startOffset?: number | null,
+  pRoot?: TRoot | null,
+  pReviver?: BSONReviver | null
```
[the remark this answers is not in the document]
nbbeeken: (See other comment, comes from Java)

I can't write `root ??= ...` because TS warns that root may have been instantiated with a different type. Maybe I'm doing something wrong but I think the problem is TS can't prevent this invocation:

```ts
parseToStructure<{ kind: 'custom'; dest: number }>(bytes);
```
Making the thing returned different from the type given. But this is not the intended use, it is supposed to automatically determine the type:
```ts
parseToStructure(bytes, undefined, { kind: 'custom', dest: 0 });
```
Maybe I have to use overrides 🤔

src/parser/on_demand/parse_to_structure.ts
```diff
@@ -0,0 +1,143 @@
...
+  /** BSON Embedded types */
+  const enum t {
+    object = 3,
+    array = 4,
+    javascriptWithScope = 15
+  }
+
+  embedded: while (ctx !== null) {
+    for (
+      let it: BSONElement | undefined = ctx.elements[ctx.elementOffset++];
```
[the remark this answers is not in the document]
nbbeeken: Yea but I don't mind changing `it`. The loop is also implemented in an iterator-like fashion (initialization moves the offset forward) so if it made sense to this could be rewritten as for-of

src/parser/on_demand/parse_to_structure.ts
nbbeeken: I don't see where this comes in handy, the container type is used in two places, the return type of parseToStructure and the input to a reviver. A reviver can't really work on parameterized type it needs to handle any/all of these depending on what nested layer it it operating on. For the return type of `parseToStructure` you do get something better than `any` if you pass in something with a better type for `dest`

src/parser/on_demand/parse_to_structure.ts
```diff
@@ -0,0 +1,143 @@
...
+  };
+
+  /** BSONElement offsets: type indicator and value offset */
+  const enum e {
+    type = 0,
+    offset = 3
+  }
+
+  /** BSON Embedded types */
+  const enum t {
```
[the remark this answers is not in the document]
nbbeeken: Resolving this for the other thread.

src/parser/on_demand/parse_to_structure.ts
nbbeeken: Changed to desciptive names, thanks for reopening the thread Aditi so we could get our thoughts in one place. I was just going with my stylistic choice, but no need to stick to it. There is no increase to code size.

src/parser/on_demand/parse_to_structure.ts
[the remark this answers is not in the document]
nbbeeken: updated

src/parser/on_demand/parse_to_elements.ts
```diff
@@ -80,7 +82,9 @@ function findNull(bytes: Uint8Array, offset: number): number {
  * @public
  * @experimental
  */
-export function parseToElements(bytes: Uint8Array, startOffset = 0): Iterable<BSONElement> {
+export function parseToElements(bytes: Uint8Array, pOffset?: number | null): Iterable<BSONElement> {
```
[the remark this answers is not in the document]
nbbeeken: https://en.wikipedia.org/wiki/Hungarian_notation

src/parser/on_demand/parse_to_structure.ts
aditi-khare-mongoDB: LGTM, I'll wait for Bailey's response to close this thread.

src/parser/on_demand/parse_to_structure.ts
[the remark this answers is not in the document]
baileympearson: Can you not just use default values? 

```typescript
export function parseToStructure<
  TRoot extends Container = {
    dest: Record<string, unknown>;
    kind: 'object';
  }
>(
  bytes: Uint8Array,
  startOffset?: number | null,
  root: TRoot | null = {
    kind: 'object',
    dest: Object.create(null) as Record<string, unknown>
  },
  reviver: BSONReviver | null = DEFAULT_REVIVER
)
```

src/parser/on_demand/parse_to_structure.ts
baileympearson: LGTM

src/parser/on_demand/parse_to_structure.ts
baileympearson: Isn't the point that a user's reviver might result in a different output shape?  Their reviver needs to all variations but the `custom` is the final result type of the parsing

src/parser/on_demand/parse_to_structure.ts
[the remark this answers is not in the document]
nbbeeken: No default values only work for undefined, I wanted to support null to match our `== null` style code for options defaulting. Also, that doesn't fix the type issue: `'TRoot' could be instantiated with a different subtype of constraint 'Container'`. As a TS library author I want to use TRoot to automagically determine the return type based on the type of root so that users never use `parseToStructure<XXX>` type parameterization syntax. I just think there isn't a nice way to allow one without the other, so new variable it is. Open to ideas if there is a way

src/parser/on_demand/parse_to_structure.ts
nbbeeken: The value returned by parseToStructure is determined by the dest field on root. The containers returned by the reviver are to control each level of nesting. There isn't a direct relation between custom and the final return type/value.

W-A-James (approved):

baileympearson (approved):

aditi-khare-mongoDB (approved):

PR #659, opened 2024-03-12
fix(NODE-6016): flip byte order depending on system endianness

nbbeeken:
#### What is changing?

Flipped the order we parse bytes from the float array depending on platform endianness

No.

Floats will not serialize or deserialize correctly on big endian systems

### Fixed float byte-wise handling on big endian systems

A recent change to the BSON library started parsing and serializing floats using a `Float64Array`. When reading the bytes from this array the ordering is dependent on the platform it is running on and we now account for that ordering.

### DIFF
--- a/test/node/utils/number_utils.test.ts
+++ b/test/node/utils/number_utils.test.ts
@@ -1,7 +1,32 @@
 import { expect } from 'chai';
 import { NumberUtils } from '../../../src/utils/number_utils';
 
+const FLOAT = new Float64Array(1);
+const FLOAT_BYTES = new Uint8Array(FLOAT.buffer, 0, 8);
+
+FLOAT[0] = -1;
+// Little endian [0, 0, 0, 0, 0, 0,  240, 191]
+// Big endian    [240, 191, 0, 0, 0, 0, 0, 0]
+const isBigEndian = FLOAT_BYTES[7] === 0;
+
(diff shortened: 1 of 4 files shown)


test/node/utils/number_utils.test.ts
```diff
@@ -1,7 +1,32 @@
...
+    });
+
+    context('setFloat64LE()', () => {
+      it('should return -1 as little endian bytes', () => {
+        const buf = new Uint8Array([0, 0, 0, 0, 0, 0, 0, 0]);
+        NumberUtils.setFloat64LE(buf, 0, -1);
+        expect(buf).to.deep.equal(new Uint8Array([0, 0, 0, 0, 0, 0, 240, 191]));
+      });
+    });
+  });
```
addaleax, line 28: Do you want to add a test that takes a BSON binary blob containing a float, deserializes it, checks that it's the right value, and vice versa?

addaleax (approved):

test/node/utils/number_utils.test.ts
nbbeeken: Sure yea integration from top to bottom is a good call.

W-A-James (approved):

baileympearson (approved):

addaleax (approved):

PR #662, opened 2024-03-18
feat(NODE-5959): make byte parsing utils available on onDemand library

nbbeeken:
#### What is changing?

- Adds utf and number parsing utils to onDemand

No

A reviver will be receiving offsets into a BSON byte sequence and will need to interpret them at JS values. These utilities provide what we need without rewriting the logic they already implement. We do not necessarily need these to be part of the final API, the driver works on Node.js so we do not need the logic for switching between web and node. We also probably only need a subset of these functions as mentioned above, for example, toHex is for display purposes and likely does not have a use for what we aim to do with this API.

### DIFF
--- a/src/utils/byte_utils.ts
+++ b/src/utils/byte_utils.ts
@@ -1,7 +1,10 @@
 import { nodeJsByteUtils } from './node_byte_utils';
 import { webByteUtils } from './web_byte_utils';
 
-/** @internal */
+/**
+ * @public
+ * @experimental
+ */
 export type ByteUtils = {
   /** Transforms the input to an instance of Buffer if running on node, otherwise Uint8Array */
   toLocalBufferType(buffer: Uint8Array | ArrayBufferView | ArrayBuffer): Uint8Array;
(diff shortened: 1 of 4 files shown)


src/utils/byte_utils.ts
```diff
@@ -1,7 +1,10 @@
 import { nodeJsByteUtils } from './node_byte_utils';
 import { webByteUtils } from './web_byte_utils';
 
-/** @internal */
+/**
+ * @public
+ * @experimental
+ */
```
aditi-khare-mongoDB, line 7: Can we add API docs to the `ByteUtils` type, to clarify what it encapsulates?

src/utils/number_utils.ts
```diff
@@ -6,12 +6,29 @@ FLOAT[0] = -1;
 // Big endian    [191, 240, 0, 0, 0, 0, 0, 0]
 const isBigEndian = FLOAT_BYTES[7] === 0;
 
+/**
+ * @experimental
+ * @public
+ */
+export type NumberUtils = {
```
aditi-khare-mongoDB, line 13: Can we add API docs to the `NumberUtils` type, to clarify what it encapsulates? Just a sentence or two.

aditi-khare-mongoDB (changes requested):

aditi-khare-mongoDB (approved):

W-A-James (approved):

PR #663, opened 2024-03-20
fix(NODE-6042): Binary.toString output with respect to position

aditi-khare-mongoDB:
### Description
`Binary.toString` and `Binary.toJSON` now respect position when printing out binary object.

#### What is changing?
See release highlights.

No.

Bug Fix.

### `Binary.toString`  and `Binary.toJSON` now respect position when printing out binary object.
When the `toString()` or `toJSON` method is called on a `Binary` instance, it will print out until the instance's position property of the object, instead of the entire buffer.

Example:
```typescript
new BSON.Binary().toString();
// old output: '\x00\x00\x00\x00...' (256 zeros)
// new output:  ' ' 
```


W-A-James (approved):
Code LGTM, but for the release note, could you just make the code block have typescript syntax highlighting?

nbbeeken (approved):

PR #665, opened 2024-03-29
fix(NODE-6059): clean up experimental APIs

nbbeeken:
#### What is changing?

- Remove parseToStructure
- Fix NumberUtils and ByteUtils `this` type issue
- Move `BSONOffsetError` to top-level export along with other errors
- Move `getSize` to NumberUtils
  - Add wrapper to support bson specific message and error type

No

A portion of these APIs are not going to be used, and there was some TS convenience issues.


W-A-James (dismissed):

src/parser/on_demand/parse_to_elements.ts
```diff
@@ -44,23 +45,13 @@ export type BSONElement = [
...
- * @experimental
- * @public
- *
- * Parses a int32 little-endian at offset, throws if it is negative
- */
-export function getSize(source: Uint8Array, offset: number): number {
-  if (source[offset + 3] > 127) {
-    throw new BSONOffsetError('BSON size cannot be negative', offset);
+const getSize = (source: Uint8Array, offset: number) => {
+  try {
```
baileympearson, line 49: Why did we move `getSize` into NumberUtils?  Also, this error handling is partially redundant.  We already throw almost the same error in the `getSize` that is in number utils.

src/utils/number_utils.ts
```diff
@@ -31,6 +36,18 @@ export type NumberUtils = {
  * @public
  */
 export const NumberUtils: NumberUtils = {
+  getSize(source: Uint8Array, offset: number): number {
```
baileympearson, line 39: Is this really any better than:

```typescript
const size = NumberUtils.parseInt32LE(...);
if (size < 0) { .... }
```

src/parser/on_demand/parse_to_elements.ts
nbbeeken: It is a number-related utility I don't agree with our previous decision to keep it separate since it is one of the foundational number parsing operations we need in this library. The error handling keeps this API throwing a consistent error type and an error that is a subclass of BSON while keeping the getSize helper BSON agnostic.

src/utils/number_utils.ts
nbbeeken: Yes, we can short-circuit this way avoiding unnecessary bit-wise operations.

src/parser/on_demand/parse_to_elements.ts
nbbeeken: From slack, improved the naming to make this utility more generic to numeric parsing.

src/utils/number_utils.ts
nbbeeken: A small one but yes measurable

W-A-James (approved):

baileympearson (approved):

PR #669, opened 2024-04-11
fix(NODE-6074): Avoiding top-level awaits in `bson` by introducing "node" bundles

kraenhansen:
#### What is changing?

I suggest introducing two node specific bundles, which performs the import of `node:crypto`.

The "realm" SDK uses a similar pattern of platform specific entrypoints declared via separate export conditions and TS project references.

See https://jira.mongodb.org/browse/NODE-6074


package.json
```diff
@@ -107,7 +109,7 @@
     "check:granular-bench": "npm run build:bench && node ./test/bench/etc/run_granular_benchmarks.js",
     "check:spec-bench": "npm run build:bench && node ./test/bench/lib/spec/bsonBench.js",
     "build:bench": "cd test/bench && npx tsc",
-    "build:ts": "node ./node_modules/typescript/bin/tsc",
+    "build:ts": "node ./node_modules/typescript/bin/tsc --build --force",
```
kraenhansen, line 112: Adding `--build` because the project now relies on TS project references.
Adding `--force` because the `composite: true` entails `incremental: true` and subsequent runs of `build:dts` would fail because the `clean_definition_files` deletes files produced by `tsc` and because of the `incremental: true` it TSC doesn't check the existence of cached files.

kraenhansen:
It seems there are some type issues with the tests that I have to sort out, but I'll have to end for today.
I'll put this out of draft when I'm ready with those changes 👍

elle-j (commented):
This would alleviate things for us on the Realm JS side 🙏

nbbeeken (changes requested):
We plan to meet about this so we can dive into more detail over a call but this change is a reversal of prior design decision to unify the javascript that is run in our supported environments. 

It is unfortunate that a language feature like top-level await still causes a hassle with bundling but we and others have dealt with this in the past by [enabling the flags needed to import such syntax](https://github.com/mongodb/js-bson/blob/main/test/bundling/webpack/webpack.config.js#L17) or [forced resolution to use the commonjs module](https://github.com/mongodb/js-bson/blob/main/package.json#L93-L95).

kraenhansen:
Okay. We'll revisit our configurations and try to make it work 👍

jk89:
+1 for some sort of solution to this top level await. I cannot use this library with my Ionic Angular project because of it. Yes the ESNext bundler can handle it but the dev bundler which I need for development does not support it.

PR #670, opened 2024-04-16
feat(NODE-6087): add Int32.fromString method

aditi-khare-mongoDB:
### Description
Add static `Int32.fromString()` method.

Yes, there are new API docs.

NODE-3660 user ticket, our validation in the Int32 is lacking. In V7, we will add `Int32.fromString(value)` validation call into the `Int32` constructor's `string` case.

### Add static `Int32.fromString` method
This method attempts to create an `Int32` type from string, and will throw a `BSONError` on any string input that is not representable as an `Int32`.
Notably, this method will also throw on the following string formats:
- Strings in non-decimal formats (exponent notation, binary, hex, or octal digits)
- Strings with non-numeric and non-leading sign characters (ex: '2.0', '24,000')
- Strings with leading and/or trailing whitespace

Strings with leading zeros, however, are allowed

### DIFF
--- a/src/int_32.ts
+++ b/src/int_32.ts
@@ -39,17 +39,21 @@ export class Int32 extends BSONValue {
    * This method will throw a BSONError on any string input that is not representable as an Int32.
    * Notably, this method will also throw on the following string formats:
    * - Strings in non-decimal formats (exponent notation, binary, hex, or octal digits)
-   * - Strings with leading zeros
-   * - Strings with decimal points (ex: '2.0')
+   * - Strings with non-numeric characters (ex: '2.0', '24,000')
    *
-   * Strings with whitespace, however, are allowed.
+   * Strings with whitespace and/or leading zeros, however, are allowed.
    *
    * @param value - the string we want to represent as an int32.
    */
   static fromString(value: string): number {
     const trimmedValue = value.trim();
-    const coercedValue = Number(trimmedValue);
-    if (coercedValue.toString() !== trimmedValue) {
+    const cleanedValue = !/[^0]+/.test(trimmedValue)
+      ? trimmedValue.replace(/^0+/, '0') // all zeros case
+      : trimmedValue.includes('-')
+        ? trimmedValue.replace(/^-0+/, '-') // negative number with leading zeros
+        : trimmedValue.replace(/^0+/, ''); // positive number with leading zeros
+    const coercedValue = Number(cleanedValue);
+    if (coercedValue.toString() !== cleanedValue) {
       throw new BSONError(`Input: '${value}' is not a valid Int32 string`);
     }
     return coercedValue;
--- a/src/int_32.ts
+++ b/src/int_32.ts
@@ -40,7 +40,7 @@ export class Int32 extends BSONValue {
    * This method will throw a BSONError on any string input that is not representable as an Int32.
    * Notably, this method will also throw on the following string formats:
    * - Strings in non-decimal formats (exponent notation, binary, hex, or octal digits)
-   * - Strings with characters other than sign or numeric characters (ex: '2.0', '24,000')
+   * - Strings non-numeric and non-leading sign characters (ex: '2.0', '24,000')
    * - Strings with leading and/or trailing whitespace
    *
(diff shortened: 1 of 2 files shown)


src/int_32.ts
```diff
@@ -32,6 +34,38 @@ export class Int32 extends BSONValue {
...
+
+    const coercedValue = Number(value);
+    if (
+      coercedValue.toString() !== cleanedValue ||
+      !Number.isSafeInteger(coercedValue) ||
+      BSON_INT32_MAX < coercedValue ||
+      BSON_INT32_MIN > coercedValue
+    ) {
+      throw new BSONError(`Input: '${value}' is not a valid Int32 string`);
+    }
```
baileympearson, line 65: Could we throw specific errors here?  Seems like low hanging fruit to give users a better experience.  For example, we could throw something like the following if the value is > the int32 max: `Input ${value} is larger than the maximum value for Int32`

test/node/int_32_tests.js
```diff
@@ -97,4 +98,50 @@ describe('Int32', function () {
...
+      ['-Infinity', '-Infinity'],
+      ['NaN', 'NaN'],
+      ['fraction', '2/3'],
+      ['commas', '34,450'],
+      ['exponentiation notation', '1e1'],
+      ['octal', '0o1'],
+      ['binary', '0b1'],
+      ['hex', '0x1'],
+      ['empty string', ''],
+      ['leading and trailing whitespace', '    89   ', 89]
```
baileympearson, line 129: According to the AC on the ticket, leading and trailing whitespace should be allowed

test/node/int_32_tests.js
```diff
@@ -97,4 +98,50 @@ describe('Int32', function () {
...
+    ];
+
+    for (const [testName, value, expectedInt32] of acceptedInputs) {
+      context(`when case is ${testName}`, () => {
+        it(`should return Int32 that matches expected value`, () => {
+          expect(Int32.fromString(value).value).to.equal(expectedInt32);
+        });
+      });
+    }
```
baileympearson, line 138: It's generally nice for tests to describe the behavior, so it's easy to understand what we expect from test output.  Currently, the output looks like this:
```
  when case is Int32.max
    should return Int32 that matches expected value
```

there are lots of ways to write this description so it's helpful, a couple of examples:

```
// no context block, just a test description
Int32.fromString(Int32.max) returns `Int32(BSON_MAX_INT_32)`

// context block
Int32.fromString(Int32.max)
  returns an Int32 representing Int32.max

// or just simple
Int32.fromString parses successfully for
	Int32.max
    Int32.min
    ...
```

If you don't mind reworking the test titles some, I think that'd be a nice improvement, so long as the test output is descriptive

test/node/int_32_tests.js
aditi-khare-mongoDB: Prohibiting whitespace is a change Neal and I discussed in office. This is because we want to stay consistent to the `decimal128.fromString` implementation. I can also ask the team for more input!

test/node/int_32_tests.js
aditi-khare-mongoDB: Bookkeeping: Discussing in slack.

src/int_32.ts
aditi-khare-mongoDB: Sounds good, I've update the error messages, but for the catch all `toString() ===` check, I've kept the generic message.

test/node/int_32_tests.js
baileympearson: 🤷 conversely, Long.fromString does support leading and trailing whitespace:

```javascript
> bson.Long.fromString('   5')
new Long('5')
> bson.Long.fromString('       5')
new Long('5')
> bson.Long.fromString('   10   ')
new Long('10')
```

aditi-khare-mongoDB: imo, it's better to be more strict now since this is a public function, and then if we decide to allow it later, we can. 
Also, `Long.fromString` is a function we're likely going to change for v7, while `decimal128.fromString` is more stable.

PR #671, opened 2024-04-16
feat(NODE-6086): add Double.fromString() method

aditi-khare-mongoDB:
### Description
Add static `Double.fromString()` method.

Yes, there are new API docs.

NODE-3660 user ticket, our validation in the Double is lacking. In V7, we will add `Double.fromString(value)` validation call into the `Double` constructor's `string` case.

### Add static `Double.fromString` method
This method attempts to create an `Double` type from a string, and will throw a `BSONError` on any string input that is not representable as a `IEEE-754 64-bit double`.
 Notably, this method will also throw on the following string formats:
   - Strings in non-decimal formats (exponent notation, binary, hex, or octal digits)
   - Strings with characters other than sign, numeric, floating point, or slash characters (Note: `'Infinity'`, `'-Infinity'`, and `'NaN'` input strings are still allowed)
   - Strings with leading and/or trailing whitespace
Strings with leading zeros, however, are also allowed.

### DIFF
--- a/src/double.ts
+++ b/src/double.ts
@@ -49,14 +49,18 @@ export class Double extends BSONValue {
   static fromString(value: string): Double {
     const coercedValue = Number(value);
     const nonFiniteValidInputs = ['Infinity', '-Infinity', 'NaN'];
-    if (
+
+    if (value.trim() !== value) {
+      throw new BSONError(`Input: '${value}' contains whitespace`);
+    } else if (value === '') {
+      throw new BSONError(`Input is an empty string`);
+    } else if (/[^-0-9.]/.test(value) && !nonFiniteValidInputs.includes(value)) {
+      throw new BSONError(`Input: '${value}' contains invalid characters`);
+    } else if (
       (!Number.isFinite(coercedValue) && !nonFiniteValidInputs.includes(value)) ||
-      (Number.isNaN(coercedValue) && value !== 'NaN') ||
-      value === '' ||
-      (/[^-0-9.]/.test(value) && !nonFiniteValidInputs.includes(value)) ||
-      value.trim() !== value
+      (Number.isNaN(coercedValue) && value !== 'NaN')
     ) {
-      throw new BSONError(`Input: '${value}' is not a valid Double string`);
+      throw new BSONError(`Input: ${value} is not representable as a Double`); // generic case
     }
     return new Double(coercedValue);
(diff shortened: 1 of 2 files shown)


billouboq:
@nbbeeken Sorry to bother you in this PR, this feels like the only way of communicating since there is no issue in this repository 😆 
And sorry Aditi to pollute your PR with this comment 🙏 

I saw that there is a branch for date optimisations here : NODE-5955-improve-date

Is it something that will get released at some point ?

src/double.ts
```diff
@@ -32,6 +33,34 @@ export class Double extends BSONValue {
...
+   * Notably, this method will also throw on the following string formats:
+   * - Strings in non-decimal formats (exponent notation, binary, hex, or octal digits)
+   * - Strings with characters other than numeric, floating point, or leading sign characters (Note: 'Infinity', '-Infinity', and 'NaN' input strings are still allowed)
+   * - Strings with leading and/or trailing whitespace
+   *
+   * Strings with leading zeros, however, are also allowed
+   *
+   * @param value - the string we want to represent as an double.
+   */
+  static fromString(value: string): Double {
```
baileympearson, line 49: same comment here as on the `Int32.fromString()` PR: can we throw descriptive errors for each case we're checking for in the `if` statement?

test/node/double.test.ts
```diff
@@ -225,6 +225,59 @@ describe('BSON Double Precision', function () {
...
+        ['leading and trailing whitespace', '    89   ', 89],
+        ['fake positive infinity', '2e308'],
+        ['fake negative infinity', '-2e308'],
+        ['fraction', '3/4'],
+        ['foo', 'foo']
+      ];
+
+      for (const [testName, value, expectedDouble] of acceptedInputs) {
+        context(`when case is ${testName}`, () => {
+          it(`should return Double that matches expected value`, () => {
```
baileympearson, line 261: Same comment here about test titles.

src/double.ts
aditi-khare-mongoDB: Yep, on it

test/node/double.test.ts
aditi-khare-mongoDB: Just changed them, going to push after my tests run!

baileympearson (approved):

PR #675, opened 2024-04-18
feat(NODE-5648): add Long.fromStringStrict()

aditi-khare-mongoDB:
### Description
Add `Long.fromStringStrict` which throws an error if the input string has invalid characters. 

#### What is changing?
See above.

Yes, we should request docs changes, and there is an API docs addition. 

User submitted bug: [NODE-5638](https://jira.mongodb.org/browse/NODE-5638).

### Add `Long.fromStringStrict` method 
The `Long.fromStringStrict` method is almost identical to the `Long.fromString` method, except it throws a `BSONError` if any of the following are true:
- input string has invalid characters, for the given radix
- the string contains whitespace
- the value the input parameters represent is too large or too small to be a 64-bit Long 

Unlike `Long.fromString`, this method does not coerce the inputs `'+/-Infinity'` and `'NaN'` to `Long.ZERO`, in any case.

**Examples:**
``` typescript
Long.fromStringStrict('1234xxx5'); // throws BSONError
Long.fromString('1234xxx5'); // coerces input and returns new Long(123400)

// when writing in radix 10, 'n' and 'a' are both invalid characters
Long.fromStringStrict('NaN'); // throws BSONError
Long.fromString('NaN'); // coerces input and returns Long.ZERO
```
> [!NOTE]
>`Long.fromStringStrict`'s functionality will be present in `Long.fromString` in the V7 BSON release.


src/long.ts
```diff
@@ -246,13 +247,43 @@ export class Long extends BSONValue {
...
    * @param radix - The radix in which the text is written (2-36), defaults to 10
+   * @param throwsError - Whether or not throwing an error is permitted
    * @returns The corresponding Long value
    */
-  static fromString(str: string, unsigned?: boolean, radix?: number): Long {
+  static fromStringHelper(
+    str: string,
+    unsigned?: boolean,
+    radix?: number,
+    throwsError?: boolean
```
baileympearson, line 285: Generally, options are named imperatively, not passively.  so this might be `throwError` instead of `throwsError`.  

I'd also suggest making the name more descriptive - it's not accurate to say this option controls whether errors are thrown or not. When `throwsError` is false, we throw errors if the string is empty or in other invalid cases.  This option specifically controls whether or not we validate characters in the string, so a name such as `validateStringCharacters` might be more appropriate.

Also, it's best practice to only make values optional when they're sometimes not provided.  In this PR, you call `fromStringHelper` in two places, both of which provide a value for `throwsError`.

Lastly, fromStringHelper isn't very descriptive.  We've [used the pattern of a shared internal method in the past](https://github.com/mongodb/node-mongodb-native/blob/eece8c1b9fcd0aa7c12722f4fe4e50e2123102b8/src/client-side-encryption/client_encryption.ts#L677) when we need two different public APIs with overlapping implementations and we chose the naming convention `private _<methodName>` to indicate that it is internal.  In this case, `private _fromString(...)` instead of `fromStringHelper`.

src/long.ts
```diff
@@ -246,13 +247,43 @@ export class Long extends BSONValue {
   }
 
   /**
+   * @internal
+   * Returns false for an string that contains invalid characters for its radix, else returns the original string.
+   * @param str - The textual representation of the Long
+   * @param radix - The radix in which the text is written (2-36), defaults to 10
+   */
+  static validateStringCharacters(str: string, radix?: number): false | string {
```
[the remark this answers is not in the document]
baileympearson: Can we also make this private?

src/long.ts
```diff
@@ -246,13 +247,43 @@ export class Long extends BSONValue {
...
+    radix = radix ?? 10;
+
+    let regexInputString = '';
+    if (radix <= 10) {
+      regexInputString = `[^-0-${radix - 1}+]`;
+    } else {
+      const validCharRangeEnd = String.fromCharCode('a'.charCodeAt(0) + (radix - 11));
+      regexInputString = `[^-0-9+(a-${validCharRangeEnd})]`;
+    }
```
baileympearson, line 264: regardless of the implementation chosen, can we add comments explaining what this regex is looking for?  regexes tend to be pretty opaque.

Could we consolidate these branches with something like this?
```suggestion
const validDigits = '0123456789abcdefghijklmnopqrstuvwxyz`.slice(0, radix);
const regex = `[^-+${validDigits}`];
...
```

src/long.ts
```diff
@@ -263,12 +294,20 @@ export class Long extends BSONValue {
...
     }
     radix = radix || 10;
+
     if (radix < 2 || 36 < radix) throw new BSONError('radix');
 
+    if (throwsError && !Long.validateStringCharacters(str, radix)) {
+      throw new BSONError(`Input: '${str}' contains invalid characters for radix: ${radix}`);
+    }
+    if (throwsError && str.trim() !== str) {
+      throw new BSONError(`Input: '${str}' contains whitespace.`);
```
baileympearson, line 304: Can we clarify that it's specifically leading or trailing whitespace?  internal whitespace will be caught in the check above

src/long.ts
```diff
@@ -291,6 +330,41 @@ export class Long extends BSONValue {
...
+   * - the value the string represents is too large or too small to be a Long
+   * @param str - The textual representation of the Long
+   * @param unsigned - Whether unsigned or not, defaults to signed
+   * @param radix - The radix in which the text is written (2-36), defaults to 10
+   * @returns The corresponding Long value
+   */
+  static fromStringStrict(str: string, unsigned?: boolean, radix?: number): Long {
+    // remove leading zeros (for later string comparison and to make math faster)
+    const cleanedStr = removeLeadingZeros(str);
+    // doing this check outside of recursive function so cleanedStr value is consistent
```
baileympearson, line 347: Could you clarify what this comment means?

src/long.ts
```diff
@@ -291,6 +330,41 @@ export class Long extends BSONValue {
...
+   * Throws an error if any of the following conditions are true:
+   * - the string contains invalid characters for the given radix
+   * - the string contains whitespace
+   * - the value the string represents is too large or too small to be a Long
+   * @param str - The textual representation of the Long
+   * @param unsigned - Whether unsigned or not, defaults to signed
+   * @param radix - The radix in which the text is written (2-36), defaults to 10
+   * @returns The corresponding Long value
+   */
+  static fromStringStrict(str: string, unsigned?: boolean, radix?: number): Long {
```
baileympearson, line 344:
```suggestion
  static fromStringStrict(string: string, unsigned?: boolean, radix?: number): Long {
```

This will be in our public API, let's make sure we use descriptive variable names since that's what shows up in autocomplete and in our docs.

(optional) I also tend to avoid abbreviations inside our code for the same reason it's best practice to ensure we use good, full parameter names in our public API.  So I'd use `cleanedString` instead of `cleanedStr`.  But this is mostly preference, feel free to ignore this part.

baileympearson (changes requested):

src/long.ts
aditi-khare-mongoDB: yep, sounds good made these changes :)

src/long.ts
baileympearson: oh, interesting.  I didn't notice that.  we can leave it as-is then

baileympearson (dismissed):

src/long.ts
```diff
@@ -291,6 +324,41 @@ export class Long extends BSONValue {
...
+    // remove leading zeros (for later string comparison and to make math faster)
+    const cleanedStr = removeLeadingZeros(str);
+    // doing this check outside of recursive function so cleanedStr value is consistent
+    const result = Long._fromString(cleanedStr, true, unsigned, radix);
+    if (result.toString(radix).toLowerCase() !== cleanedStr.toLowerCase()) {
+      throw new BSONError(
+        `Input: ${str} is not representable as ${result.unsigned ? 'an unsigned' : 'a signed'} 64-bit Long with radix: ${radix}`
+      );
```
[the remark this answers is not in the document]
nbbeeken: Otherwise this says "with radix: undefined" for most cases

test/node/long.test.ts
```diff
@@ -163,4 +163,101 @@ describe('Long', function () {
...
+      expectedStr?: string
+    ][] = [
+      ['basic no alphabet low radix', '1236', true, 8],
+      ['negative basic no alphabet low radix', '-1236', false, 8],
+      ['valid upper and lower case letters in string with radix > 10', 'eEe', true, 15],
+      ['hexadecimal letters', '126073efbcdADEF', true, 16],
+      ['negative hexadecimal letters', '-1267efbcdDEF', false, 16],
+      ['negative leading zeros', '-00000032', false, 15, '-32'],
+      ['leading zeros', '00000032', false, 15, '32'],
+      ['explicit positive leading zeros', '+00000032', false, 15, '32'],
```
nbbeeken, line 182: explicit positive without any leading zeros

test/node/long.test.ts
```diff
@@ -163,4 +163,101 @@ describe('Long', function () {
...
+      ['max unsigned binary input', Long.MAX_UNSIGNED_VALUE.toString(2), true, 2],
+      ['max unsigned decimal input', Long.MAX_UNSIGNED_VALUE.toString(10), true, 10],
+      ['max unsigned hex input', Long.MAX_UNSIGNED_VALUE.toString(16), true, 16],
+      ['max signed binary input', Long.MAX_VALUE.toString(2), false, 2],
+      ['max signed decimal input', Long.MAX_VALUE.toString(10), false, 10],
+      ['max signed hex input', Long.MAX_VALUE.toString(16), false, 16],
+      ['min signed binary input', Long.MIN_VALUE.toString(2), false, 2],
+      ['min signed decimal input', Long.MIN_VALUE.toString(10), false, 10],
+      ['min signed hex input', Long.MIN_VALUE.toString(16), false, 16],
+      ['signed zero', '0', false, 10],
```
nbbeeken, line 192: should this be `+0` and do we have a `-0` case?

test/node/long.test.ts
```diff
@@ -163,4 +163,101 @@ describe('Long', function () {
...
+      ['under min signed binary input', Long.MIN_VALUE.toString(2) + '1', false, 2],
+      ['under min signed decimal input', Long.MIN_VALUE.toString(10) + '1', false, 10],
+      ['under min signed hex input', Long.MIN_VALUE.toString(16) + '1', false, 16],
+      ['string with whitespace', '      3503a  ', false, 11]
+    ];
+
+    for (const [testName, str, unsigned, radix, expectedStr] of successInputs) {
+      context(`when the input is ${testName}`, () => {
+        it(`should return a Long represenation of the input`, () => {
+          expect(Long.fromStringStrict(str, unsigned, radix).toString(radix)).to.equal(
```
nbbeeken, line 217:
```suggestion
        it(`should return a Long representation of the input`, () => {
```

src/int_32.ts
```diff
@@ -48,11 +49,7 @@ export class Int32 extends BSONValue {
    * @param value - the string we want to represent as an int32.
    */
   static fromString(value: string): Int32 {
-    const cleanedValue = !/[^0]+/.test(value)
-      ? value.replace(/^0+/, '0') // all zeros case
-      : value[0] === '-'
-        ? value.replace(/^-0+/, '-') // negative number with leading zeros
-        : value.replace(/^\+?0+/, ''); // positive number with leading zeros
+    const cleanedValue = removeLeadingZeros(value);
```
nbbeeken, line 52:
```
BSON.Int32.fromString('+32')
Uncaught BSONError: Input: '+32' is not a valid Int32 string
```
Same issue in Long, is this intentional? I think it should be permitted

nbbeeken (changes requested):

src/int_32.ts
aditi-khare-mongoDB: This is not intentional, it is a mistake on my end. Good thing that editing the shared function will fix the problem for both!

src/long.ts
```diff
@@ -246,13 +247,37 @@ export class Long extends BSONValue {
...
+   * @param str - The textual representation of the Long
+   * @param radix - The radix in which the text is written (2-36), defaults to 10
+   */
+  private static validateStringCharacters(str: string, radix?: number): false | string {
+    radix = radix ?? 10;
+    const validCharacters = '0123456789abcdefghijklmnopqrstuvwxyz'.slice(0, radix);
+    // regex is case insensitive and checks that each character within the string is one of the validCharacters
+    const regex = new RegExp(`[^-+${validCharacters}]`, 'i');
+    return regex.test(str) ? false : str;
+  }
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Yep! Done

src/long.ts
```diff
@@ -246,13 +247,37 @@ export class Long extends BSONValue {
...
-  static fromString(str: string, unsigned?: boolean, radix?: number): Long {
+  private static _fromString(
+    str: string,
+    validateStringCharacters: boolean,
+    unsigned?: boolean,
+    radix?: number
+  ): Long {
     if (str.length === 0) throw new BSONError('empty string');
     if (str === 'NaN' || str === 'Infinity' || str === '+Infinity' || str === '-Infinity')
       return Long.ZERO;
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: We can run every validation before _fromString is called but there's certain validations that `Long.fromString` and `Long.fromStringStrict` share (no empty string, no interior hyphen, no invalid radices) that would cause some code repetition. Which is why I've only moved the round trip check outside of the function

src/long.ts
```diff
@@ -263,12 +288,20 @@ export class Long extends BSONValue {
...
     }
     radix = radix || 10;
+
     if (radix < 2 || 36 < radix) throw new BSONError('radix');
 
+    if (validateStringCharacters && !Long.validateStringCharacters(str, radix)) {
+      throw new BSONError(`Input: '${str}' contains invalid characters for radix: ${radix}`);
+    }
+    if (validateStringCharacters && str.trim() !== str) {
+      throw new BSONError(`Input: '${str}' contains leading and/or trailing whitespace.`);
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: If it contains interior whitespace, this check won't catch it, which is why this check is worded in this way.

src/long.ts
```diff
@@ -291,6 +324,41 @@ export class Long extends BSONValue {
...
+   * @param unsigned - Whether unsigned or not, defaults to signed
+   * @param radix - The radix in which the text is written (2-36), defaults to 10
+   * @returns The corresponding Long value
+   */
+  static fromStringStrict(str: string, unsigned?: boolean, radix?: number): Long {
+    // remove leading zeros (for later string comparison and to make math faster)
+    const cleanedStr = removeLeadingZeros(str);
+    // doing this check outside of recursive function so cleanedStr value is consistent
+    const result = Long._fromString(cleanedStr, true, unsigned, radix);
+    if (result.toString(radix).toLowerCase() !== cleanedStr.toLowerCase()) {
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Good catch, just fixed it!

src/long.ts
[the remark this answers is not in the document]
aditi-khare-mongoDB: Another good catch, made the change!

test/node/long.test.ts
aditi-khare-mongoDB: Yep, added it!

test/node/long.test.ts
aditi-khare-mongoDB: I can make the change, it was originally just 'signed' bc `unsigned = false`

test/node/long.test.ts
aditi-khare-mongoDB: oh no, typo! fixed it

src/long.ts
```diff
@@ -291,6 +324,41 @@ export class Long extends BSONValue {
...
+   * - the string contains invalid characters for the given radix
+   * - the string contains whitespace
+   * - the value the string represents is too large or too small to be a Long
+   * @param str - The textual representation of the Long
+   * @param unsigned - Whether unsigned or not, defaults to signed
+   * @param radix - The radix in which the text is written (2-36), defaults to 10
+   * @returns The corresponding Long value
+   */
+  static fromStringStrict(str: string, unsigned?: boolean, radix?: number): Long {
+    // remove leading zeros (for later string comparison and to make math faster)
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Bookkeeping: asked for clarification on Slack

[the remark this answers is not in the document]
aditi-khare-mongoDB: Slack discussion: add arg parsing before _fromString call and remove extra validation to only `fromStringStrict`

src/long.ts
```diff
@@ -291,6 +290,67 @@ export class Long extends BSONValue {
...
+   * - the string contains whitespace
+   * - the value the string represents is too large or too small to be a Long
+   * @param str - The textual representation of the Long
+   * @param unsigned - Whether unsigned or not, defaults to signed
+   * @param radix - The radix in which the text is written (2-36), defaults to 10
+   * @returns The corresponding Long value
+   */
+  static fromStringStrict(str: string, unsigned?: boolean, radix?: number): Long {
+    if (str === 'NaN' || str === 'Infinity' || str === '+Infinity' || str === '-Infinity')
+      return Long.ZERO;
```
nbbeeken, line 306: This would be a silent coercion from unsupported string values, I thought this was undesirable?

nbbeeken (changes requested):

src/long.ts
aditi-khare-mongoDB: Bookkeeping: Current changes only coerce  `+/-Infinity` and `NaN` cases to `Long.Zero` in `Long.fromString` when the radix does not support their respective letters.

test/node/utils/string_utils.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    ['multiple decimal points', '..', 30],
+    ['non a-z or numeric string', '~~', 36],
+    ['alphabet in radix < 10', 'a', 4],
+    ['radix does not allow all alphabet letters', 'eee', 14]
+  ];
+
+  for (const [testName, str, radix] of successInputs) {
+    context(`when the input is ${testName}`, () => {
+      it(`should return a input string`, () => {
+        expect(StringUtils.validateStringCharacters(str, radix)).to.equal(str);
```
nbbeeken, line 42: I don't see an import for expect, can we add that?

src/utils/string_utils.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * @internal
+ * Removes leading zeros and explicit plus from textual representation of a number.
+ */
+export function removeLeadingZerosandExplicitPlus(str: string): string {
+  return !/[^+?0]+/.test(str)
```
nbbeeken, line 6:
```suggestion
export function removeLeadingZerosAndExplicitPlus(str: string): string {
```
Or something more succinct `trimInsignificant`?

nbbeeken, line 6: slack q about perf

nbbeeken (changes requested):

nbbeeken (dismissed):

src/long.ts
```diff
@@ -291,6 +287,135 @@ export class Long extends BSONValue {
...
+      unsigned = !!unsignedOrRadix;
+    }
+    radix ??= 10;
+    if (str === 'NaN' && radix < 24) {
+      // radix does not support n, so coerce to zero
+      return Long.ZERO;
+    } else if ((str === 'Infinity' || str === '+Infinity' || str === '-Infinity') && radix < 35) {
+      // radix does not support y, so coerce to zero
+      return Long.ZERO;
+    }
```
baileympearson, line 415: Why are we checking the radix here?  This changes the behavior of Long.fromString - something we want to avoid in this PR.  The logic was previously:

```typescript
if (str === 'NaN' || str === 'Infinity' || str === '+Infinity' || str === '-Infinity')
      return Long.ZERO;
```

src/utils/string_utils.ts
```diff
@@ -0,0 +1,45 @@
...
+  const isNegative = str[startIndex] === '-';
+  const isExplicitlyPositive = str[startIndex] === '+';
+
+  if (isExplicitlyPositive || isNegative) {
+    startIndex += 1;
+  }
+
+  let foundInsignificantZero = false;
+
+  while (str[startIndex] === '0') {
```
baileympearson, line 21: I guess this works because indexing a string with a length that's too long just returns undefined so we break from the loop but relying on behavior such as this feels odd.  Could we add a length check and convert this to a for loop?

```typescript
for (; startIndex < str.length && str[startIndex] === '0'; ++startIndex) {
...
}
```

baileympearson (changes requested):

src/long.ts
aditi-khare-mongoDB: It does change the behavior of Long.fromString, but its another bug fix, (see updated release notes).

We previously did not allow valid inputs to radices, because we interpreted them as 
'not a number,' but in `radix >= 24`, 'nan' is a valid number. 

Example:
`Long.fromString('NaN', 27); // new Long(17060)`

src/utils/string_utils.ts
aditi-khare-mongoDB: good catch! just converted it

baileympearson (approved):

nbbeeken (approved):

PR #676, opened 2024-04-23
fix(NODE-6123): utf8 validation is insufficiently strict

aditi-khare-mongoDB:
### Description
Outside of web, our `toUtf8` function was insufficiently strict and allowed overlong encodings. 

#### What is changing?
Change our functionality to use  js`TextDecoder` to double check utf8 input when a replacement character is detected. 

No.

Drivers wide initiative to make UTF-8 validation strict and consistent. 

### UTF-8 validation now throws a `BSONError` on overlong encodings in Node.js
Specifically, this affects `deserialize` when utf8 validation is enabled, which is the default.

An overlong encoding is when the number of bytes in an encoding is inflated by padding the code point with leading 0s ([see here for more information](https://en.wikipedia.org/wiki/UTF-8#Overlong_encodings)).

### DIFF
--- a/etc/rollup/rollup-plugin-require-vendor/require_vendor.mjs
+++ b/etc/rollup/rollup-plugin-require-vendor/require_vendor.mjs
@@ -5,7 +5,7 @@ const REQUIRE_WEB_UTILS_POLYFILLS =
 const { encode: btoa, decode: atob } = require('../vendor/base64');\n`
 
 const REQUIRE_VALIDATE_UTF8_POLYFILLS = 
-  `const { TextEncoder } = require('../vendor/text-encoding');`;
+  `const { TextDecoder } = require('../vendor/text-encoding');`;
 
 export class RequireVendor {
   /**
--- a/etc/rollup/rollup-plugin-require-vendor/require_vendor.mjs
+++ b/etc/rollup/rollup-plugin-require-vendor/require_vendor.mjs
@@ -1,8 +1,11 @@
 import MagicString from 'magic-string';
 
-const REQUIRE_POLYFILLS =
-  `const { TextEncoder, TextDecoder } = require('../vendor/text-encoding');
+const REQUIRE_WEB_UTILS_POLYFILLS =
+  `const { TextEncoder } = require('../vendor/text-encoding');
 const { encode: btoa, decode: atob } = require('../vendor/base64');\n`
+
+const REQUIRE_VALIDATE_UTF8_POLYFILLS = 
+  `const { TextEncoder } = require('../vendor/text-encoding');`;
 
 export class RequireVendor {
   /**
@@ -14,17 +17,24 @@ export class RequireVendor {
    * @returns {{ code: string; map: import('magic-string').SourceMap }}
    */
   transform(code, id) {
-    if (!id.includes('validate_utf8')) {
-      return;
+    if (id.includes('validate_utf8')) {
+      // MagicString lets us edit the source code and still generate an accurate source map
+      const magicString = new MagicString(code);
+      magicString.prepend(REQUIRE_VALIDATE_UTF8_POLYFILLS);
+
+      return {
+        code: magicString.toString(),
+        map: magicString.generateMap({ hires: true })
+      };
+    } else if (id.includes('web_byte_utils')) {
+      // MagicString lets us edit the source code and still generate an accurate source map
+      const magicString = new MagicString(code);
+      magicString.prepend(REQUIRE_WEB_UTILS_POLYFILLS);
+
+      return {
+        code: magicString.toString(),
+        map: magicString.generateMap({ hires: true })
+      };
     }
-
-    // MagicString lets us edit the source code and still generate an accurate source map
-    const magicString = new MagicString(code);
-    magicString.prepend(REQUIRE_POLYFILLS);
-
-    return {
-      code: magicString.toString(),
-      map: magicString.generateMap({ hires: true })
-    };
   }
 }
--- a/etc/rollup/rollup-plugin-require-vendor/require_vendor.mjs
+++ b/etc/rollup/rollup-plugin-require-vendor/require_vendor.mjs
@@ -14,7 +14,7 @@ export class RequireVendor {
    * @returns {{ code: string; map: import('magic-string').SourceMap }}
    */
   transform(code, id) {
-    if (!id.includes('web_byte_utils')) {
+    if (!id.includes('validate_utf8')) {
       return;
     }
 
--- a/etc/rollup/rollup-plugin-require-vendor/require_vendor.mjs
+++ b/etc/rollup/rollup-plugin-require-vendor/require_vendor.mjs
@@ -4,8 +4,8 @@ const REQUIRE_WEB_UTILS_POLYFILLS =
   `const { TextEncoder } = require('../vendor/text-encoding');
 const { encode: btoa, decode: atob } = require('../vendor/base64');\n`
 
-const REQUIRE_VALIDATE_UTF8_POLYFILLS = 
-  `const { TextDecoder } = require('../vendor/text-encoding');`;
+const REQUIRE_PARSE_UTF8_POLYFILLS = 
+  `const { TextDecoder } = require('../vendor/text-encoding');\n`;
 
 export class RequireVendor {
   /**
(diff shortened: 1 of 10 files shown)


src/validate_utf8.ts
```diff
@@ -16,32 +29,17 @@ const CONTINUING_CHAR = 0x80;
...
-        continuation = 3;
-      } else {
-        return false;
-      }
+  end: number,
+  fatal: boolean
+): string {
+  if (fatal) {
+    try {
+      return new TextDecoder('utf8', { fatal }).decode(buffer.slice(start, end));
```
billouboq, line 39: the text decoder could be lazy loaded

aditi-khare-mongoDB: By 'lazy loading,' I'm thinking you mean store the decoder instances so they are created only once at the first invocation of the `validateUtf8` method (see [commit 8a07891](https://github.com/mongodb/js-bson/pull/676/commits/8a07891c057a600d8fe3c0c2cb1dbbb5f5ed182d).

aditi-khare-mongoDB: If the `TextDecoder` is still slower than `Buffer.toString`, and the following are true:
- fatal/strict validation is on
- the string contains a replacement character
- the user is on a node environment

then this may have a negative performance impact when a user opts for validating utf8 with the driver.

etc/rollup/rollup-plugin-require-vendor/require_vendor.mjs
```diff
@@ -1,9 +1,12 @@
 import MagicString from 'magic-string';
 
-const REQUIRE_POLYFILLS =
-  `const { TextEncoder, TextDecoder } = require('../vendor/text-encoding');
+const REQUIRE_WEB_UTILS_POLYFILLS =
+  `const { TextEncoder } = require('../vendor/text-encoding');
 const { encode: btoa, decode: atob } = require('../vendor/base64');\n`
 
+const REQUIRE_VALIDATE_UTF8_POLYFILLS = 
+  `const { TextDecoder } = require('../vendor/text-encoding');`;
```
nbbeeken, line 8:
```suggestion
  `const { TextDecoder } = require('../vendor/text-encoding');\n`;
```
Not a functional change but without this the code produced looks a little strange.

src/parser/deserializer.ts
```diff
@@ -604,12 +603,12 @@ function deserializeObject(
...
-        if (!validateUtf8(buffer, index, index + stringSize - 1)) {
-          throw new BSONError('Invalid UTF-8 string in BSON document');
-        }
-      }
-      const namespace = ByteUtils.toUTF8(buffer, index, index + stringSize - 1, false);
+      const namespace = ByteUtils.toUTF8(
+        buffer,
+        index,
+        index + stringSize - 1,
+        validation != null && (validation.utf8 as boolean)
```
nbbeeken, line 610: I think this may be the incorrect way to determine validation, shouldn't this be `shouldValidateKey`?

src/utils/node_byte_utils.ts
```diff
@@ -139,7 +139,7 @@ export const nodeJsByteUtils = {
       // TODO(NODE-4930): Insufficiently strict BSON UTF8 validation
       for (let i = 0; i < string.length; i++) {
         if (string.charCodeAt(i) === 0xfffd) {
-          if (!validateUtf8(buffer, start, end)) {
+          if (!validateUtf8(buffer, start, end, fatal)) {
```
nbbeeken, line 142: Because of the control flow around this we only ever call `validateUtf8` with true, which means it will throw so the if stmt doesn't make sense anymore. Can you clean that up and hard code true here instead of passing through fatal? just so it is clear that it is always on at this point.

nbbeeken (changes requested):

etc/rollup/rollup-plugin-require-vendor/require_vendor.mjs
aditi-khare-mongoDB: I see, made the change!

src/parser/deserializer.ts
aditi-khare-mongoDB: I've made the change and it doesn't affect any tests. 
It does stray from the method used by the rest of this function, but I wonder why this was implemented this way?

src/utils/node_byte_utils.ts
aditi-khare-mongoDB: Yep, cleaned it up per our conversation

src/validate_utf8.ts
```diff
@@ -1,13 +1,29 @@
...
+  new (label: 'utf8', options: { fatal: boolean; ignoreBOM?: boolean }): TextDecoder;
+};
+
+type TextEncoder = {
+  readonly encoding: string;
+  encode(input?: string): Uint8Array;
+};
+type TextEncoderConstructor = {
+  new (): TextEncoder;
+};
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: good catch! took it out

src/validate_utf8.ts
```diff
@@ -16,32 +32,19 @@ const CONTINUING_CHAR = 0x80;
...
-      } else {
-        return false;
-      }
+  end: number,
+  fatal: boolean
+): string {
+  if (fatal) {
+    try {
+      TextDecoderFatal ??= new TextDecoder('utf8', { fatal: true });
+      return TextDecoderFatal.decode(buffer.slice(start, end));
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: sg, i made the change, i will keep this in mind for the future!

test/node/byte_utils.test.ts
```diff
@@ -417,21 +419,57 @@ const toUTF8Tests: ByteUtilTest<'toUTF8'>[] = [
...
-      expect(error).to.not.exist;
-      expect(output).to.equal('abc\uFFFD');
+    name: 'throw an error if fatal is set and string contains overlong encoding',
+    inputs: [Buffer.from('11000000025f0005000000f08282ac0000', 'hex'), 0, 18, true],
+    expectation({ error }) {
+      expect(error).to.match(/Invalid UTF-8 string in BSON document/i);
+    }
+  },
+  {
+    name: 'throw an error if fatal is set and string contains invalid bytes',
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: I've added them in to explicitly check the cases Warren and I went over in my kickoff, while the web utf tests may change + depending on if we continue updating them, they might not explicitly test for the functionality that we expect.

test/node/byte_utils.test.ts
```diff
@@ -493,6 +531,51 @@ const randomBytesTests: ByteUtilTest<'randomBytes'>[] = [
   }
 ];
 
+// extra error cases copied from Web platform specs
+const toUTF8ErrorCaseTests = [
+  { input: [0xff], name: 'invalid code' },
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: We can leave a link and leave the encodings in. I've changed the name to `toUTF8WebPlatformSpecTests` how is that?

src/utils/node_byte_utils.ts
```diff
@@ -139,7 +139,7 @@ export const nodeJsByteUtils = {
       // TODO(NODE-4930): Insufficiently strict BSON UTF8 validation
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: done!

src/parse_utf8.ts
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Determines if the passed in bytes are valid utf8
+ * @param bytes - An array of 8-bit bytes. Must be indexable and have length property
+ * @param start - The index to start validating
+ * @param end - The index to end validating
+ */
+export function parseUtf8(buffer: Uint8Array, start: number, end: number, fatal: boolean): string {
+  if (fatal) {
+    try {
+      TextDecoderFatal ??= new TextDecoder('utf8', { fatal: true });
```
nbbeeken, line 27:
```suggestion
    TextDecoderFatal ??= new TextDecoder('utf8', { fatal: true });
    try {
```

If this were to throw we wouldn't want to catch and convert it to a BSON error that should only be for the actual parsing.

test/node/byte_utils.test.ts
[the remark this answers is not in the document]
nbbeeken: That seems fine although if our goal is to make sure we keep getting the same expectations covered from a user perspective then this shouldn't test the utility directly, we should pass these bytes inside a BSON document and call serialize and makes sure that the same error we expect reaches the user, via `serialize` for these bytes.

test/node/byte_utils.test.ts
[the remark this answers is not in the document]
nbbeeken: I was expecting leaving the whole link like so: `https://github.com/web-platform-tests/wpt/commit/7c9f8674d9809731e8919073d957d6233f6e0544` but this is fine too

test/node/byte_utils.test.ts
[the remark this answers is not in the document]
aditi-khare-mongoDB: yes, we can certainly add these tests in `serialize` as well!

src/parser/deserializer.ts
aditi-khare-mongoDB: Yes, I'll add a test, messaged you on Slack for more clarification

PR #677, opened 2024-04-30
fix(NODE-6144): Long.fromString incorrectly coerces valid inputs to Long.ZERO in special cases

aditi-khare-mongoDB:
### `Long.fromString` takes radix into account before coercing '+/-Infinity' and 'NaN' to `Long.ZERO`
`Long.fromString` no longer coerces the following cases to `Long.ZERO` when the provided radix supports all characters in the string:
- `'+Infinity'`, `'-Infinity'`, or `'Infinity'` when 35 <= radix  <= 36
- `'NaN'` when 24 <= radix  <= 36

``` typescript
// when writing in radix 27, 'n' and 'a' are valid characters, so 'NaN' represents the decimal number 17060
Long.fromString('NaN', 27); // new Long(17060)
Long.fromString('NaN', 10); // new Long(0) <-- Since 'NaN' is not a valid input in base 10, it gets coerced to Long.ZERO
```

### DIFF
--- a/src/long.ts
+++ b/src/long.ts
@@ -406,7 +406,11 @@ export class Long extends BSONValue {
       unsigned = !!unsignedOrRadix;
     }
     radix ??= 10;
-    if (str === 'NaN' || str === 'Infinity' || str === '+Infinity' || str === '-Infinity') {
+    if (str === 'NaN' && radix < 24) {
+      // radix does not support n, so coerce to zero
+      return Long.ZERO;
+    } else if ((str === 'Infinity' || str === '+Infinity' || str === '-Infinity') && radix < 35) {
+      // radix does not support y, so coerce to zero
       return Long.ZERO;
     }
     return Long._fromString(str, unsigned, radix);
--- a/src/long.ts
+++ b/src/long.ts
@@ -371,12 +371,28 @@ export class Long extends BSONValue {
 
   /**
    * Returns a signed Long representation of the given string, written using radix 10.
+   *
+   * If the input string is empty, this function will throw a BSONError.
+   *
+   * If input string does not have valid signed 64-bit Long representation, this method will return a coerced value:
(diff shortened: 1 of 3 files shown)


src/long.ts
```diff
@@ -406,7 +406,11 @@ export class Long extends BSONValue {
       unsigned = !!unsignedOrRadix;
     }
     radix ??= 10;
-    if (str === 'NaN' || str === 'Infinity' || str === '+Infinity' || str === '-Infinity') {
+    if (str === 'NaN' && radix < 24) {
+      // radix does not support n, so coerce to zero
+      return Long.ZERO;
+    } else if ((str === 'Infinity' || str === '+Infinity' || str === '-Infinity') && radix < 35) {
+      // radix does not support y, so coerce to zero
```
W-A-James, line 413: This doesn't account for lower case or mixed case inputs. Is this intended to be undefined behaviour? If so, we should document that somewhere, otherwise we should decide on an approach to standardize this behaviour.

```
> Long.fromString('infinity', 25) // new Long('185011843')

> Long.fromString("InFinity", 25) // new Long('185011843')

> Long.fromString("Infinity", 25) // new Long('0')

```

W-A-James (changes requested):

src/long.ts
aditi-khare-mongoDB: This is the desired behavior. If a user inputs non-valid characters in a radix, they might get undefined behavior. In the case that they input the actual js syntax for `Infinity` the driver will coerce the input into something manageable.

Since this PR intends to fix the output for valid numeric inputs, not coercions for invalid inputs, this wasn't something I considered in scope.

aditi-khare-mongoDB: I added in some more documentation to fromString though!

nbbeeken (approved):

PR #681, opened 2024-05-02
perf(NODE-6126): improve Long.fromBigInt performance

aditi-khare-mongoDB:
### Description
Refactor `Long.fromBigInt` to only use bit math, and refactor the `Long` constructor to have cleaner logic + remove `Object.assign` calls

#### What is changing?
Improve performance of `Long.fromBigInt`

No.

Use a more efficient algorithm for `Long.fromBigInt`

### Optimize performance of `Long.fromBigInt`

Internally fromBigInt was originally implemented using toString of the bigint value. Now, `Long.fromBigInt` has been refactored to use bitwise operations greatly improving performance.

### DIFF
--- a/src/long.ts
+++ b/src/long.ts
@@ -133,26 +133,51 @@ export class Long extends BSONValue {
 
   /**
    * Constructs a 64 bit two's-complement integer, given its low and high 32 bit values as *signed* integers.
-   *  See the from* functions below for more convenient ways of constructing Longs.
-   *
-   * Acceptable signatures are:
-   * - Long(low, high, unsigned?)
-   * - Long(bigint, unsigned?)
-   * - Long(string, unsigned?)
    *
    * @param low - The low (signed) 32 bits of the long
    * @param high - The high (signed) 32 bits of the long
    * @param unsigned - Whether unsigned or not, defaults to signed
    */
-  constructor(low: number | bigint | string = 0, high?: number | boolean, unsigned?: boolean) {
+  constructor(low: number, high?: number, unsigned?: boolean);
+  /**
+   * Constructs a 64 bit two's-complement integer, given a bigint representation.
+   *
+   * @param value - BigInt representation of the long value
+   * @param unsigned - Whether unsigned or not, defaults to signed
+   */
+  constructor(value: bigint, unsigned?: boolean);
+  /**
+   * Constructs a 64 bit two's-complement integer, given a string representation.
+   *
+   * @param value - 32 bit number representation of the Long
+   * @param unsigned - Whether unsigned or not, defaults to signed
+   */
+  constructor(value: number, unsigned?: boolean);
+  /**
+   * Constructs a 64 bit two's-complement integer, given a string representation.
+   *
+   * @param value - String representation of the long value
+   * @param unsigned - Whether unsigned or not, defaults to signed
+   */
+  constructor(value: string, unsigned?: boolean);
+  constructor(
+    lowOrValue: number | bigint | string = 0,
+    highOrUnsigned?: number | boolean,
+    unsigned?: boolean
+  ) {
     super();
-    if (typeof low === 'bigint') {
-      Object.assign(this, Long.fromBigInt(low, !!high));
-    } else if (typeof low === 'string') {
-      Object.assign(this, Long.fromString(low, !!high));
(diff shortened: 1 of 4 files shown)


aditi-khare-mongoDB:
[Link to sheet](https://docs.google.com/spreadsheets/d/1t_rCAm-sbVc9BwRYNFkZFKCT9eu0f53AkLilNdkwvY0/edit#gid=0)

Doesn't look like there's much change, through the initial commit

src/long.ts
```diff
@@ -243,7 +254,15 @@ export class Long extends BSONValue {
    * @returns The corresponding Long value
    */
   static fromBigInt(value: bigint, unsigned?: boolean): Long {
-    return Long.fromString(value.toString(), unsigned);
+    // eslint-disable-next-line no-restricted-globals
+    Long.FROM_BIGINT_BIT_MASK ??= BigInt(0xffffffff);
```
nbbeeken, line 258: Remove caching

src/long.ts
```diff
@@ -176,6 +176,17 @@ export class Long extends BSONValue {
...
+  /**
+   * @internal
+   * bit mask used for fromBigInt method, lazy loaded
+   */
+  static FROM_BIGINT_BIT_MASK: bigint;
+
+  /**
+   * @internal
+   * bit shift used for fromBigInt method, lazy loaded */
+  static FROM_BIGINT_BIT_SHIFT: bigint;
```
nbbeeken, line 188: not using these anymore

src/long.ts
```diff
@@ -176,6 +176,17 @@ export class Long extends BSONValue {
   /** Minimum signed value. */
```
nbbeeken, line 176: contstuctor overloads and Object.assign improvement

test/node/long.test.ts
```diff
@@ -164,6 +166,47 @@ describe('Long', function () {
...
+    beforeEach(function () {
+      if (__noBigInt__) {
+        this.currentTest?.skip();
+      }
+    });
+
+    for (const [testName, num, unsigned, expectedStr] of inputs) {
+      context(`when the input is ${testName}`, () => {
+        it(`should return a Long representation of the input`, () => {
+          expect(Long.fromBigInt(num, unsigned).toString(10)).to.equal(expectedStr);
```
nbbeeken, line 204: deep equal check instead of string

nbbeeken (changes requested):

src/long.ts
nbbeeken: We should remove the `!` from the low,high,unsigned properties, Typescript will help assert that if there's any new properties in the Long class that they must be set in the CTOR

src/long.ts
aditi-khare-mongoDB: bookkeeping: we tested this out in the huddle, and it doesn't affect performance and removing it will make the code less clunky

src/long.ts
aditi-khare-mongoDB: good catch

src/long.ts
aditi-khare-mongoDB: Yep, good call!

test/node/long.test.ts
aditi-khare-mongoDB: Sounds good!

src/long.ts
nbbeeken: There's still `!` on the properties because we still use the assign for fromString.

Something to consider, using a ternary that results in "always" fixed types might work well here, then you can just assign the results to the properties unconditionally

```js
    const {
      low: lowNumber,
      high: highNumber,
      unsigned: unsignedBoolean
    } = typeof low === 'string'
      ? Long.fromString(low, Boolean(high))
      : typeof low === 'bigint'
      ? Long.fromBigInt(low, Boolean(high))
      : typeof high === 'boolean'
      ? { low, high: 0, unsigned: high }
      : { low, high: high ?? 0, unsigned: Boolean(unsigned) };
```

nbbeeken (changes requested):
comment about `!`, req changes for viz

src/long.ts
aditi-khare-mongoDB: I see, didn't realize we were going to make any changes in the `string` branch

nbbeeken (approved):

W-A-James (approved):

PR #692, opened 2024-05-21
feat(NODE-6156): add signature to github releases

aditi-khare-mongoDB:
### Description
Sign releases in both 5.x and 6.x

#### What is changing?
Automate release signing with a detached signature and verification instructions in the README.
Example Signature Link: https://github.com/mongodb/js-bson/actions/runs/9195659055

No

SSDLC Compliance

### Add Signature to Github Releases
The Github release for `js-bson` now contains a detached signature file for the NPM package (named
`bson-X.Y.Z.tgz.sig`), on every major and patch release to 6.x and 5.x.  To verify the signature, follow the instructions in the 'Release Integrity' section of the `README.md` file.


.github/workflows/release-5.x.yml
```diff
@@ -25,6 +25,33 @@ jobs:
...
+          artifactory_username: ${{ secrets.ARTIFACTORY_USER }}
+          artifactory_password: ${{ secrets.ARTIFACTORY_PASSWORD }}
+      - if: steps.release.outputs.release_created
+        name: "Upload artifacts"
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ env.PACKAGE_FILE }}
+          path: |
+            ${{ env.PACKAGE_FILE }}
+            ${{ env.PACKAGE_FILE }}.sig
```
baileympearson, line 53: We'll need to upload the signature to the GH and make it publicly available, and it should stay there indefinitely (it will, so long as the GH release exists).  

I think we should be able to use `gh release upload` (the gh cli comes pre-installed on GHA runners): https://cli.github.com/manual/gh_release_upload

baileympearson (changes requested):
Nice work!  

One other note - Looks like only the release tarball (no signature file) is upload: https://github.com/mongodb/js-bson/actions/runs/9195659055

.github/workflows/release-5.x.yml
aditi-khare-mongoDB: Yep! My bad on forgetting this step

aditi-khare-mongoDB:
@baileympearson If you click on the bson-6.6.0.tar.gz file, it contains a bson-6.6.0.tar.gz.sig file within

baileympearson:
@aditi-khare-mongoDB Oh, you're right!  Nice.

Is is possible to upload them separately?  I don't know if it matters much, but I think we should shoot for something similar to what devtools has for their repos, where the signatures are attached separate from the file.  Here's an example: https://github.com/mongodb-js/compass/releases/tag/v1.43.0

nbbeeken:
Jumping in here, yeah I also expected the .sig to be a "top-level" artifact and that the only tgz would be the one that can be obtained from `npm`. Not that there is something wrong with a tgz on our GH releases it just isn't the tgz users are likely to ingest. The following command:

```sh
npm view bson@latest dist.tarball
```
Will get the tar url, a bit of `curl` here, and a bit of `gpg` there ✨ and you've got a validated signature. I am also thinking that the tgz we upload to npm gets replicated to artifactory instances so again the tgz on GH isn't the tgz folks care about.

.github/workflows/release-5.x.yml
```diff
@@ -25,6 +25,27 @@ jobs:
       - if: ${{ steps.release.outputs.release_created }}
         name: actions/setup
         uses: ./.github/actions/setup
+      - if: steps.release.outputs.release_created
+        run: npm pack
+        shell: bash
+      - if: steps.release.outputs.release_created
+        name: Get release version and release package file name
+        run: |
+          PACKAGE_VERSION=$(jq '.version' package.json | tr -d '"')
```
nbbeeken, line 34:
```suggestion
          PACKAGE_VERSION=$(jq --raw-output '.version' package.json)
```

.github/workflows/release.yml
```diff
@@ -17,13 +17,35 @@ jobs:
       - id: release
         uses: google-github-actions/release-please-action@v4
 
+
       # If release-please created a release, publish to npm
-      - if: ${{ steps.release.outputs.release_created }}
+      - if: steps.release.outputs.release_created
```
nbbeeken, line 22: Is this correct? I thought you needed the ${{ }} to insert the value, otherwise this would just be the string "steps.release.outputs.release_created" no?

.github/workflows/release.yml
```diff
@@ -17,13 +17,35 @@ jobs:
...
         name: actions/setup
         uses: ./.github/actions/setup
+      - if: steps.release.outputs.release_created
+        run: npm pack
+        shell: bash
+      - if: steps.release.outputs.release_created
+        name: Get release version and release package file name
+        run: |
+          PACKAGE_VERSION=$(jq '.version' package.json | tr -d '"')
```
nbbeeken, line 33:
```suggestion
          PACKAGE_VERSION=$(jq --raw-output '.version' package.json)
```

.github/workflows/release.yml
nbbeeken: Also, just an optional improvement, but since we're going to be duping this everywhere I think it is worth doing. We can make the release-please it's own "job" that outputs this, and then this job "needs" release-please. Then the whole job can be conditional upon "release_created" rather than every step having to declare this.

.github/workflows/release.yml
```diff
@@ -17,13 +17,35 @@ jobs:
...
+          echo "PACKAGE_VERSION=${PACKAGE_VERSION}" >> "$GITHUB_ENV"
+          echo "PACKAGE_FILE=bson-${PACKAGE_VERSION}.tgz" >> "$GITHUB_ENV"
+      - if: steps.release.outputs.release_created
+        name: Create detached signature
+        uses: mongodb-labs/drivers-github-tools/garasign/gpg-sign@main
```
nbbeeken, line 38: Will this be versioned? I think there is a v1 already

.github/workflows/release.yml
```diff
@@ -17,13 +17,35 @@ jobs:
...
         uses: ./.github/actions/setup
+      - if: steps.release.outputs.release_created
+        run: npm pack
+        shell: bash
+      - if: steps.release.outputs.release_created
+        name: Get release version and release package file name
+        run: |
+          PACKAGE_VERSION=$(jq '.version' package.json | tr -d '"')
+          echo "PACKAGE_VERSION=${PACKAGE_VERSION}" >> "$GITHUB_ENV"
+          echo "PACKAGE_FILE=bson-${PACKAGE_VERSION}.tgz" >> "$GITHUB_ENV"
```
nbbeeken, line 35: There's a mixture of variable usage here that I think we can consolidate a bit. So reading the [docs `GITHUB_ENV`](https://docs.github.com/en/actions/learn-github-actions/variables#passing-values-between-steps-and-jobs-in-a-workflow) will set an environment variable for all subsequent steps which means you can access it with `$PACKAGE_FILE` rather than `${{ env.PACKAGE_FILE }}` even if both work one is accessing the env variable "in shell" the other is inserting the value into the "script" string before the shell has been given the "script". 

There's also `GITHUB_OUTPUT` which is what we used in [release automation](https://github.com/mongodb/node-mongodb-native/blob/main/.github/scripts/util.mjs#L5-L20) (The JS there is equivalent to `echo "x=y" >> $GITHUB_OUTPUT`) which allows you to hand the variable to github's runner logic so you can pass values between jobs. Then only the `${{ }}` syntax is the way you can access the variable. 

I prefer the latter, because setting environment variables can have side effects, the ones we are using here probably won't but keeping the values explicitly external seems like a defensive programming approach. But either way, let's go all in on one, so if we're setting env variables then the scripts should use the env variable rather than being generated strings.

README.md
```diff
@@ -11,6 +11,19 @@ You can learn more about it in [the specification](http://bsonspec.org).
...
+
+### Release Integrity
+
+The GitHub release contains a detached signature file for the NPM package (named
+`mongodb-X.Y.Z.tgz.sig`).
+
+To verify the integrity of the downloaded package, run the following command:
+
+```shell
+gpg --verify bson-X.Y.Z.tgz.sig bson-X.Y.Z.tgz
```
nbbeeken, line 23: Just a reminder to update these instructions when we have the final `.sig` file name and  how to get the tgz.

.github/workflows/release.yml
nbbeeken: If we take this further, we can make this all a composite action that takes the variables here as inputs. Even if we don't make a composite action in this PR, writing the yaml to use outputs today is a small lift that will remove the context needed on another day to sort out which variables source from where.

nbbeeken (changes requested):

README.md
aditi-khare-mongoDB: Yep! I've updated them now, let me know your thoughts.

.github/workflows/release-5.x.yml
aditi-khare-mongoDB: Made the change!

.github/workflows/release.yml
aditi-khare-mongoDB: I've separated into two jobs, thanks!

.github/workflows/release.yml
aditi-khare-mongoDB: I've gone with the latter option; I agree, in the context of future composite actions, it'll make future work easier.

.github/workflows/release.yml
aditi-khare-mongoDB: Yep, just added it. Seems like v1 was published just yesterday

nbbeeken (approved):
Please add release highlights, otherwise LGTM

baileympearson (approved):

PR #704, opened 2024-07-19
feat(NODE-6031): add `t` and `i` to Timestamp

nbbeeken:
#### What is changing?

- add `t` and `i` getters to Timestamp
- change internals to use getters

No

- users do not have to recall whether t or i is the high or low bits
- a Timestamp can be constructed from another Timestamp

### Timestamp now has `t` and `i` properties

To make this type a bit easier to use we are surfacing the breakdown of the two internal 32 bit segments of a Timestamp value.

```ts
const ts = new Timestamp({ i: 2, t: 1 });
ts.i // 2
ts.t // 1
```

### DIFF
--- a/test/node/timestamp.test.ts
+++ b/test/node/timestamp.test.ts
@@ -22,7 +22,7 @@ describe('Timestamp', () => {
       expect(ts.t).to.equal(l.high);
     });
 
-    describe.only('when signed negative input is provided to the constructor', () => {
+    describe('when signed negative input is provided to the constructor', () => {
       it('t and i return unsigned values', () => {
         const l = new BSON.Long(-1, -2);
         // Check the assumption that Long did NOT change the values to unsigned.
--- a/test/node/timestamp.test.ts
+++ b/test/node/timestamp.test.ts
@@ -22,11 +22,21 @@ describe('Timestamp', () => {
       expect(ts.t).to.equal(l.high);
     });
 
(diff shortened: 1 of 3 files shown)


test/node/timestamp.test.ts
```diff
@@ -9,6 +9,27 @@ describe('Timestamp', () => {
...
+      const ts = new BSON.Timestamp(l);
+      expect(ts.t).to.equal(l.high);
+    });
+
+    it('t and i always return unsigned values', () => {
+      const l = new BSON.Long(-1, -2);
+      const ts = new BSON.Timestamp(l);
+      expect(ts.i).to.equal(0xffffffff); // -1 unsigned
+      expect(ts.t).to.equal(0xfffffffe); // -2 unsigned
+    });
```
aditi-khare-mongoDB, line 30:
```suggestion
context('when signed negative input is provided to the constructor', () => {
    it('t and i still return unsigned values', () => {
      const l = new BSON.Long(-1, -2);
      const ts = new BSON.Timestamp(l);
      expect(ts.i).to.equal(0xffffffff); // -1 unsigned
      expect(ts.t).to.equal(0xfffffffe); // -2 unsigned
    });
 });
```

nbbeeken: updated naming, and added an assumption check that the Long class does not make the value unsigned

aditi-khare-mongoDB (approved):

PR #707, opened 2024-08-06
perf(NODE-6246): Use buffer pool for ObjectId to significantly improve memory usage

SeanReece:
### Description

This PR significantly improves memory usage for ObjectId by introducing a buffer/Uint8Array pool. This is similar to how NodeJS `Buffer` uses an internal pool when using `allocUnsafe`, but the difference here is instead of instantiating a Buffer/Uint8Array for each ObjectId, each ObjectId holds an offset to it's pool.

### Why?

Buffer (or TypedArray) is inefficient at storing small amount of data
Buffer of size 12 takes 96 bytes in v8! An ObjectId with buffer is 128 bytes
A pointer and a Number take 16 bytes. An ObjectId using pool is 48 bytes + pool 

1000 ObjectId using pool = **60,048 bytes**
1000 ObjectIds without pool = **128,000 bytes**

**~53% reduction in memory usage**

Lots more discussion in https://github.com/mongodb/js-bson/pull/703

### Memory improvements

Using MFlix sample database. 1 Million documents

Before: **655MB**
Now: **369MB**

**~44% reduction**

### Performance tests

Since we're still making use of Buffer/Uint8Array the performance is mostly the same. There is a slight regression creating from `Buffer` (since we are copying instead of just saving the reference), but an improvement when creating from Uint8Array (since we do not need to convert to Buffer).

### Performance ideas
Now that ObjectId allocates its own Buffer/Uint8Array, we technically never need to call `toLocalBufferType` since we know we're always operating on the correct buffer type. Removing that check is outside the scope of this change but would increase performance. 

### Notes
- the pool is lazily initialized so that services that don't use ObjectId do not incur a penalty here
- Setting `ObjectId.poolSize = 1` essentially disables the pool (Each ObjectId will have its own buffer)
- I tried to keep the API as consistent as possible, including exposing `buffer` property which returns a Buffer/Uint8Array.

Adding `ObjectId.poolSize` to allow users to set their preferred ObjectId pool size. This references the # of ObjectIds that will be stored in each pool, the actual pool allocated with be 12 * poolSize. This defaults to 1000 as that seemed reasonable.

### Improve memory usage by pooling ObjectId buffers

@SeanReece brought to our attention the fact that `Uint8Array` requires a surprising amount of overhead. The 12 bytes of `ObjectId` data actually takes up 96 bytes of memory. You can find an excellent breakdown of what all that overhead is [here](https://stackoverflow.com/questions/45803829/memory-overhead-of-typed-arrays-vs-strings/45808835#45808835). 

In this release, @SeanReece has significantly improved the memory used for each `ObjectId` by introducing a `Uint8Array` pool of memory. The idea is similar to how [Node.js' `Buffer` uses an internal pool](https://nodejs.org/docs/latest/api/buffer.html#class-property-bufferpoolsize) when using `allocUnsafe` or `from` APIs that overwrite the returned memory. Each `ObjectId` uses the current shared pool and stores an offset to where its bytes begin, all operations internal to the class use this offset so there is no impact to the public API. 

#### Crunching the numbers

The default `ObjectId.poolSize` is 1000. Since it is measured in ObjectIds this means it has a default size of 12,000 bytes plus the same overhead that _each_ `ObjectId` was incurring. 

- Before pool: 
  - 1 ObjectId uses `128 B` 
  - 1000 ObjectIds use `128 KB`
  - 1,000,000 ObjectIds use: `128 MB`
- After pool: 
  - 1 ObjectId uses `40 B` + ~`12,000 B` 
  - 1000 ObjectIds use `40 KB` + ~`12 KB`
  - 1,000,000 ObjectIds use: `40 MB` + ~`12.08 MB`

As you can see the new cost for a single `ObjectId` is higher than before but as your data set grows the shared pools lead to huge savings in memory. If the initial cost is too high or your data sets are even larger the pool's size is configurable!

#### `ObjectId.poolSize`

The shared pool is not created until the first `ObjectId` is constructed so you can modify the `poolSize` to fit your needs: 

```js
import { ObjectId } from 'bson';
// or import { ObjectId } from 'mongodb';
ObjectId.poolSize = 1;
``` 

> [!TIP]
> Setting the poolSize to 1 essentially disables this change.

> [!TIP]
> You can change the `poolSize` at any time during your program, the next time the current pool's size runs out a new one will be created using the current `poolSize` value.

Thank you so much to @SeanReece for a very thorough and informative bug report and for working so hard on this improvement!

#### A note about deep equality

`ObjectId`s will no longer respond to naive deep equality checks by default. APIs like [util.deepStrictEqual](https://nodejs.org/api/assert.html#assertdeepstrictequalactual-expected-message) and [lodash.isEqual](https://lodash.com/docs/4.17.15#isEqual) ignore what is public or private API and make assumptions about what properties are considered a part of two values' equality. 

##### Workarounds

Set `ObjectId.poolSize = 1;` disabling the optimization so now each ObjectId will have its own buffer. 

Use `objectid.equals(otherOid)` wherever possible. [`lodash.isEqualWith`](https://lodash.com/docs/4.17.15#isEqualWith) allows you to define a customizer that can be used to call the ObjectId's `equals` method.

Use `BSON.serialize` if applicable. If you pass the two objects you wish to compare to the BSON serializer, the bytes returned will be exactly equal if the documents are the same on the server.


src/objectid.ts
```diff
@@ -272,18 +342,18 @@ export class ObjectId extends BSONValue {
...
+    uint8array[index + 2] = this[_pool][this[_offset] + 2];
+    uint8array[index + 3] = this[_pool][this[_offset] + 3];
+    uint8array[index + 4] = this[_pool][this[_offset] + 4];
+    uint8array[index + 5] = this[_pool][this[_offset] + 5];
+    uint8array[index + 6] = this[_pool][this[_offset] + 6];
+    uint8array[index + 7] = this[_pool][this[_offset] + 7];
+    uint8array[index + 8] = this[_pool][this[_offset] + 8];
+    uint8array[index + 9] = this[_pool][this[_offset] + 9];
+    uint8array[index + 10] = this[_pool][this[_offset] + 10];
+    uint8array[index + 11] = this[_pool][this[_offset] + 11];
```
addaleax, line 356: Not sure if it really makes a difference, but my intuition would be that it's worth storing `this[_pool]` and `this[_offset]` in separate variables?

SeanReece: Just so I understand correctly, do you mean using some temp variables like this? Would make it easier to grok for sure.

```ts
serializeInto(uint8array: Uint8Array, index: number): 12 {
  const pool = this[_pool]
  const offset = this[_offset]
   ...
  uint8array[index + 1] = pool[offset + 1];
  ...
```

addaleax: Yes, exactly 👍

src/objectid.ts
```diff
@@ -103,17 +133,38 @@ export class ObjectId extends BSONValue {
       workingId = inputId;
     }
 
+    // If we have reached the end of the pool then create a new pool
+    if (!pool || poolOffset + 12 > pool.byteLength) {
+      createPool();
+    }
+    this[_pool] = pool;
+    this[_offset] = poolOffset;
+    poolOffset += 12;
```
nbbeeken, line 142: Above I suggest changing the type of the module scoped pool to include `null` which forces us to check here if pool is `null`. I'm thinking we can migrate the condition to create a new pool into `createPool` and possibly call it `getPool` instead. The method can either return the existing a pool or a new one if adding 12 bytes goes beyond the byteLength. 

My motive is to add a bit more type truth to that global pool since we want to be sure to start the module with the pool being `null` so that users have a chance to adjust the poolSize before any ObjectIds implicitly create it.

src/objectid.ts
nbbeeken: Even if we switch to the string properties I think this is a good call

> Would make it easier to grok for sure.

Aggreed, more so for how it reads rather than any functional difference.

src/objectid.ts
```diff
@@ -72,7 +99,7 @@ export class ObjectId extends BSONValue {
    *
    * @param inputId - A 12 byte binary Buffer.
    */
```
nbbeeken, line 101: This isn't a constructor we need to add to our public API, if you stick `@internal` on it then it gets deleted from the public Typescript :) Also, can you add the inputIndex `@param` docs?

src/objectid.ts
```diff
@@ -103,17 +133,38 @@ export class ObjectId extends BSONValue {
...
+        this[_pool],
+        this[_offset]
+      );
+    } else if (ArrayBuffer.isView(workingId)) {
+      if (workingId.byteLength !== 12 && typeof inputIndex !== 'number') {
+        throw new BSONError('Buffer length must be 12 or offset must be specified');
+      }
+      if (
+        inputIndex &&
+        (typeof inputIndex !== 'number' || inputIndex < 0 || workingId.byteLength < inputIndex + 12)
```
nbbeeken, line 159: Went down a bit of a rabbit hole analyzing this condition. 😅 It stood out to me that anything falsy won't make it to the second part that does a type and bounds check. Turns out if an empty string, or false are passed in here then the next part still works `'' + 1` will still get index `1` as will `false + 1`. There are two more falsy things that will cause an issue, the easy one is a bigint: `0n` typescript would prevent that but we can also elminate that with the type check moved before this. The other is `NaN`, which _should_ never happen, but the bounds check would elimnate it if was able to reach that check.

src/utils/node_byte_utils.ts
```diff
@@ -124,8 +124,8 @@ export const nodeJsByteUtils = {
     return Buffer.from(hex, 'hex');
   },
 
-  toHex(buffer: Uint8Array): string {
-    return nodeJsByteUtils.toLocalBufferType(buffer).toString('hex');
+  toHex(buffer: NodeJsBuffer, start?: number, end?: number): string {
```
nbbeeken, line 127:
```suggestion
  toHex(buffer: Uint8Array, start?: number, end?: number): string {
```

The idea is that we do not know we have a NodeBuffer or a Uint8Array upon entering this function, so we call toLocalBufferType which is no-op-ish if we already have one, and converts (without copying) it to a NodeBuffer if it isn't one.

src/utils/node_byte_utils.ts
```diff
@@ -6,7 +6,7 @@ type NodeJsEncoding = 'base64' | 'hex' | 'utf8' | 'binary';
 type NodeJsBuffer = ArrayBufferView &
   Uint8Array & {
     write(string: string, offset: number, length: undefined, encoding: 'utf8'): number;
-    copy(target: Uint8Array, targetStart: number, sourceStart: number, sourceEnd: number): number;
+    copy(target: Uint8Array, targetStart: number, sourceStart?: number, sourceEnd?: number): number;
     toString: (this: Uint8Array, encoding: NodeJsEncoding, start?: number, end?: number) => string;
```
nbbeeken, line 10:
```suggestion
    copy(target: Uint8Array, targetStart: number, sourceStart: number, sourceEnd: number): number;
```

It appears you do not use this change? I actually do not see any usage of this API 🤔 but for here we can just reduce the diff?

nbbeeken (changes requested):
@SeanReece Looks awesome, thanks again for the contribution, just a few improvements to line up with our practices.

One thing that is not directly part of our API but is something I want to discuss with the team is how this affects deep equality. While we have tests that are failing because it is no longer true for ObjectIds that do not share the same pool that is not the equality we _really_ care about. What matters is that two separate ObjectIds that contain the same representation will turn into equal BSON sequences, that is why our ObjectId.equals method returns true for a number of object shapes/types.

I think this optimization is worth that change, but we'll see what the consensus is among maintainers.

src/objectid.ts
```diff
@@ -4,6 +4,25 @@ import { type InspectFn, defaultInspect } from './parser/utils';
 import { ByteUtils } from './utils/byte_utils';
 import { NumberUtils } from './utils/number_utils';
 
+const defaultPoolSize = 1000; // Hold 1000 ObjectId buffers in a pool
+let pool: Uint8Array;
+let poolOffset = 0;
+
+/** Internal pool accessors for objectId instance */
+/** @internal private instance pool accessor */
+export const _pool = Symbol('pool');
```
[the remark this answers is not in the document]
SeanReece: Great. I was going to ask about your stance on Symbol properties, I originally was using typescript `private` but thought that allowing direct access to an entire pool might allow users to do something stupid/dangerous. I'll switch it back to just use `private` properties.

src/objectid.ts
SeanReece: Great idea. I was thinking something similar with a `getPool` which would handle creating the pool if necessary too. There is a *small* issue with the code above currently where calling the constructor with an invalid value (ie `new ObjectId('abc')`) would still increment the poolOffset, leading to wasted space in the pool. Thinking maybe something like this

```ts
const [ pool, offset ] = getPool() // Get pool may create a new pool if necessary. It will not 
// perform checks and assign bytes to pool
...
incrementPool() // or something
```

src/objectid.ts
```diff
@@ -37,8 +56,16 @@ export class ObjectId extends BSONValue {
 
   static cacheHexString: boolean;
 
-  /** ObjectId Bytes @internal */
-  private buffer!: Uint8Array;
+  /**
+   * The size of the buffer pool for ObjectId.
+   */
+  static poolSize: number = defaultPoolSize;
```
nbbeeken, line 62: > Would it be valuable to support setting poolSize = 0 which would essentially revert back to the original buffer-per-instance implementation?

Maybe I am missing something, would `ObjectId.poolSize = 1` do that, so could we sort of "clamp" to a minimum of 1?

IIUC, currently, 0 is going to create an empty pool, which will end up just ignoring input:

```js
let b = new Uint8Array(0)
b[0] = 1;
b[0] // undefined
```

nbbeeken: And by clamp, I think I would be fine with documenting the bounds, or we can do something fancy with a setter 🤔

src/utils/node_byte_utils.ts
SeanReece: Ah I think I switched it to play around with something and forgot to change it back Thanks for catching that.

src/utils/node_byte_utils.ts
SeanReece: Another good catch. I was using copy at some point but `Uint8Array.set` is much faster. Will remove this change

src/objectid.ts
SeanReece: I really appreciate the thorough review 😄 Took me a little while to wrap my head around this but I think I've improved the check here and I've added more tests to cover these possibilities. 

- Be more explicit when byteLength === 12 and ignore inputIndex entirely
- No falsy check
- Add isNaN

Let me know if I've missed anything here.

src/objectid.ts
SeanReece: You're right, I was just overthinking it and if poolSize === 1 then we'll still perform some unnecessary `subArray(0,12)` in some cases. Which works fine.

What I ended up doing was adding that *fancy* getter/setter so poolSize must be > 0.

SeanReece:
@nbbeeken I've made some more updates with your suggestions. 

- Removed Symbol properties. Now we just have private properties `pool` and `offset`
- Add `getPool()` and `incrementPool()` functions.
- Add setter for poolSize to enforce valid value.
- Fix for buffer offset checks.
- Fix all tests + added more coverage.

As for deep equality with ObjectIds, what I've done is added a test function `assertDeepEqualsWithObjectId` which you can pass any 2 objects/arrays into and it will first deeply find and convert all ObjectIds into their hex string representations, then it'll perform the `chai` deep equality assertion. Let me know what you think of that solution.

Let me know what you think 😄 🚀

src/objectid.ts
```diff
@@ -37,8 +61,26 @@ export class ObjectId extends BSONValue {
...
+    return poolSize;
+  }
+
+  static set poolSize(size: number) {
+    const iSize = Math.ceil(size); // Ensure new pool size is an integer
+    if (iSize <= 0) {
+      throw new BSONError('poolSize must be a positive integer greater than 0');
+    }
+    poolSize = iSize;
+  }
```
nbbeeken, line 77:
```suggestion
  static set poolSize(size: number) {
    poolSize = Math.max(Number(size) >>> 0, 1);
  }
```

I like where this is going but `ceil` isn't actually going to make sure this value is an integer, `NaN` and `Infinity` are going to sneak by the following condition. Rather than adding error handling for those cases, I think this setter should just be an "always works" kind of API, so given any input it makes a number out of it, and that number must be positive and probably no greater than UInt32.max (which would be 51GB of ObjectIds 😵‍💫). Error checking could be done externally by reading the value after it has been written and checking for equality.

src/objectid.ts
```diff
@@ -170,34 +245,38 @@ export class ObjectId extends BSONValue {
    * Generate a 12 byte id buffer used in ObjectId's
    *
    * @param time - pass in a second based timestamp.
+   * @param buffer - Optionally pass in a buffer instance.
+   * @param offset - Optionally pass in a buffer offset.
    */
-  static generate(time?: number): Uint8Array {
+  static generate(time?: number, buffer?: Uint8Array, offset: number = 0): Uint8Array {
```
nbbeeken, line 251: Sorry I did not catch this the first time, we just have to give this the same treatment we did the new constructor. The overload that accepts the new parameters should be internal and the public one can remain the `time?` argument only.

src/objectid.ts
nbbeeken: Looks great! This bit is internal so its use is "conrolled" but I think it's good we made it obvious to the next maintainer if a slight math operation goes wrong and they get NaN it will crash loudly.

test/node/object_id.test.ts
```diff
@@ -5,6 +5,13 @@ import { expect } from 'chai';
 import { bufferFromHexArray } from './tools/utils';
 import { isBufferOrUint8Array } from './tools/utils';
 
+declare module '../register-bson' {
+  interface ObjectId {
+    pool: Uint8Array;
+    offset: number;
+  }
+}
```
nbbeeken, line 13: Should be able to remove this, if you were expirencing type errors it's because register-bson depends on the types being built, so if you switch branches etc, `npm run build:dts` (or just `npm install`) is a prerequiste to get the current library's types.

test/node/tools/utils.js
```diff
@@ -218,3 +218,25 @@ module.exports.sorted = (iterable, how) => {
...
+  } else if (obj !== null && typeof obj === 'object' && !Buffer.isBuffer(obj)) {
+    return Object.entries(obj).reduce((acc, [key, value]) => {
+      acc[key] = deepOidValue(value);
+      return acc;
+    }, {});
+  }
+  return obj;
+};
+
+module.exports.assertDeepEqualsWithObjectId = (actual, expected) => {
```
nbbeeken, line 238: I appreciate this solution but I'll call out the source of truth for any two BSON values being equal is that they result in the same BSON bytes regardless of the properties and internals that they contain. I think it is important we keep the tests oriented towards what users can expect to rely upon in their code. So I was thinking that the places where we need to stop change usages of `expect(x).to.deep.equal(y)` could instead be: `expect(BSON.serialize(x)).to.deep.equal(BSON.serialize(y))`. Do you think we can adjust the tests to use (something like) this easily? 


Also, I apologize I didn't get back sooner about the team's consensus on deep equality. We discussed the pros and cons, and we feel as though this change is worth introducing despite the impact it has on the deep equality assumption. We already have BSON types that are not deep equal but serialize to the same BSON (Binary) and we cannot always avoid necessary private/internal changes to our classes. We're going to make sure the release notes clearly call out the change and identify a pretty easy workaround, setting the poolSize to 1! :)

```sh
$ node -p 'var BSON = require("."); util.isDeepStrictEqual(new BSON.ObjectId("0".repeat(24)), new BSON.ObjectId("0".repeat(24)))'
false
$ node -p 'var BSON = require("."); BSON.ObjectId.poolSize = 1; util.isDeepStrictEqual(new BSON.ObjectId("0".repeat(24)), new BSON.ObjectId("0".repeat(24)))'
true
```

nbbeeken (changes requested):
Only some small things remain, thanks for this marathon contribution!

src/objectid.ts
SeanReece: I like it! My only thought here is passing in a negative number results in a huge poolSize. Which could silently lead to a pretty significant memory leak if a user accidentally passed in `-1` for example. 

`poolSize = Math.max(Number(-1) >>> 0, 1); // 4294967295`

I think adding in a `Math.abs` makes it a bit safer. WDYT?

`poolSize = Math.max(Math.abs(Number(-1)) >>> 0, 1); // 1`

test/node/tools/utils.js
SeanReece: Did my best to update any failing tests to assert equality on the BSON bytes and I've removed that `assertDeepEqualsWithObjectId` implementation.

SeanReece:
@nbbeeken Thanks for bearing with me on this marathon 😄 As always, I appreciate the thorough reviews. Hope we're looking good now 🚀

src/objectid.ts
nbbeeken: Good catch & great idea! 🚀

nbbeeken (dismissed):

nbbeeken:
LGTM! I will have the team take a final look over then I think we're good to go!

SeanReece:
@nbbeeken Thanks for the heads up! Just an FYI I noticed a significant hit to memory usage and performance when I tried using private (`#`) properties for pool/offset. It looks like that's caused by us targeting ES2021 (since # was added in ES2022) and tslib rewriting all private properties accessors to a getter/setter function which uses a WeakMap. If that's the way we want to go then we might also want to look at bumping our target to ES2022. 

FWIW I think defaulting poolSize to 1 is a good choice. Still allows users to configure based on use case, but leaves the deep equality by default. 😃

nbbeeken:
Thanks for looking into that, we were considering real `#` private properties but that has the potential to impact users of bundlers who will downlevel the code to the WeakMap implementation anyway. Lots to consider as maintainers! 

`poolSize=1` is also showing some regressions in performance from current on main, which is surprising since I would've expected it to generally be the same as before. I need to rerun and look closer into why that is. If you've still got your benchmarking setup and can check as well let me know if you're seeing the same. TIA 🙂

SeanReece:
@nbbeeken I believe the performance impact is from some extra operations required for the pool implementation, namely having to *copy* the buffer into the pool when creating ObjectId from buffer, then when retrieving the ObjectId `id`, we need to call `subArray` on the pool. I think we can improve performance in the poolSize=1 case by making some changes.

If poolSize=1
1. Do not store offset (this saves us some memory)
2. When creating from buffer (size=12), just persist the buffer
3. When retrieving the ObjectId buffer, return entire buffer

Let me run some benchmarks and see what we can do.

nbbeeken:
When you get a chance can you rebase this on main so we can get the same benchmark fixes here? TIA :)

SeanReece:
@nbbeeken Updated 👍 I've been able to get the benchmarks running locally in docker, but haven't had the time to do proper comparison of the results yet. Let me know what the results you're seeing in the benchmarks :)

nbbeeken:
Hey @SeanReece sorry about the radio silence ([it has been a busy week](https://www.mongodb.com/resources/products/alerts/potential-data-integrity-issues-sep-2024)). Thanks for continuing to get the benchmarks running on your end, those results do corroborate what I am seeing; I also still saw a regression in our objectid performance tests but it is smaller than before so we're on the right track.

I wasn't aware of how `valueOf()` affects the results of deep eq 🤔 am I understanding correctly that jest invokes that if it exists? I think chai not working is a blocker here, we have vast amount of tests that would need to be repaired if we don't make sure that we can continue to deep equality check after this is all said and done.

nbbeeken:
Just want to share the current concerns from the team here

This is a great contribution, and we are super interested in incorporating it. However, our team has a fully planned quarter ahead, so we won't be able to conduct a thorough review or make final design decisions until next quarter (begins Nov).

The PR is currently held up on the following points:
- Ensuring default performance remains optimal for users who upgrade without making changes.
- Guaranteeing deep equality is maintained by default without breaking existing functionality. We're hoping poolSize=1 gets us this without costing us point 1.
- Encapsulation of code, for which the team needs time to design a more maintainable approach, but we’d welcome your input.
- Minor improvements to testing. Mainly ensuring we have reset global state (poolSize) in before/after hooks

We intend to tackle these issues from our end and don’t expect you to continue work on this if you no longer feel inclined to unless the feedback inspires specific ideas you’d like to propose/implement.

We appreciate your patience and continued work on this. 

cc @dariakp @baileympearson @W-A-James @durran @aditi-khare-mongoDB @addaleax

SeanReece:
@nbbeeken Awesome! Does this mean we can enable the pool by default? :) Or we still want it to be opt-in?

billouboq:
What's the blocking point of this PR ?

KJ1126007:
@SeanReece  we are waiting on this :)

baileympearson:
@SeanReece Okay, so, an update on where we're at.  Thanks for your patience on this issue! 

Two weeks ago we released bson v7.  Notably, private properties did not make it into the release.

We originally introduced private properties into the ObjectId in https://github.com/mongodb/js-bson/pull/830.  After some internal testing, we realized that this has the effect of breaking lodash's deepClone() function and we decided that this breakage was too large (we discussed with the maintainer of Mongoose, who indicated that this could break many Mongoose users could be broken by this change).  We reverted the PR for v7.

Where does this leave us for this PR?  We're unfortunately between a rock and a hard place: using private properties breaks lodash's deepClone, not using private properties breaks lodash's deepEquals.  Both of these we consider supported usages.

We've decided on the following:

1. First, we'd like to investigate the performance implications of storing the ObjectId's data as a latin1 string instead of a buffer.  This would improve memory usage of ObjectIds, but we'd need to confirm this doesn't negatively impact the performance of ObjectId creation (a critical codepath in BSON).  
2. If storing a string doesn't work, we'd consider alternatives (such as allowing for different ObjectId storage representations depending on a user's desired performance characteristics or something like this).  

If this is something you are still interested in working on, we're happy to work with you on this (although the direction is admittedly a bit vague).  But we don't have the capacity to commit much more to this right now.  We will continue working on this in the future though because we do believe this is an important feature.

tadjik1:
Hi @SeanReece! After investigating this further and comparing different storage representation methods we found one that looks promising and can help us to solve this long-standing problem. Curious to hear your thoughts on it! https://github.com/mongodb/js-bson/pull/893

PR #708, opened 2024-08-22
perf(NODE-6344): Improve ObjectId.isValid(string) performance

SeanReece:
### Description

This MR improves performance of using `ObjectId.isValid('some string')` to check if a string is a valid hex representation of an ObjectId. We frequently call this function to check if a user's input is valid (for example, when creating a document that links to other document(s)).

### Performance tests

| Test                        | Ops/s (Main)         | Ops/s (New)          | % Improvement   |
|-----------------------------|----------------------|----------------------|-----------------|
| bson#new ObjectId(string)    | 3,173,643 ops/sec    | 3,299,153 ops/sec    | 3.95%           |
| bson#isValid                | 5,513,041 ops/sec    | 22,529,639 ops/sec   | 308.66%         |
| bson#isValid uppercase      | 5,498,098 ops/sec    | 21,776,208 ops/sec   | 296.07%         |
| bson#isValid invalid end       | 131,657 ops/sec      | 22,592,719 ops/sec   | 17,060.24%      |
| bson#isValid invalid start  | 132,751 ops/sec      | 152,311,298 ops/sec  | 114,634.47%     |

There is a drastic performance improvement when checking for invalid strings since throwing/catching errors is expensive.

##### Is there new documentation needed for these changes?

None

### Double check the following

- [x] Ran `npm run check:lint` script
- [x] Self-review completed using the [steps outlined here](https://github.com/mongodb/node-mongodb-native/blob/HEAD/CONTRIBUTING.md#reviewer-guidelines)
- [x] PR title follows the [correct format](https://www.conventionalcommits.org/en/v1.0.0/): `type(NODE-xxxx)[!]: description`
  - Example: `feat(NODE-1234)!: rewriting everything in coffeescript`
- [x] Changes are covered by tests
- [x] New TODOs have a related JIRA ticket

PR #717, opened 2024-09-16
chore(NODE-6382): add benchmarking for helpers

nbbeeken:
#### What is changing?

- Adds a benchmark suite that will output to our trendcharting tool.

No

- There are a number of code paths that don't relate to serialize/deserialize but can still be useful for critical paths in an application. It would be useful to improve those where we can and ensure we maintain the improvements where they are made.
- Help with the change proposed: https://github.com/mongodb/js-bson/pull/708

### DIFF
--- a/test/bench/custom/benchmarks.mjs
+++ b/test/bench/custom/benchmarks.mjs
@@ -1,0 +1,20 @@
+/* eslint-disable strict */
+import { BSON } from '../../../lib/bson.mjs';
+
+const ObjectId_isValid = [
+  function objectid_isvalid_strlen() {
+    BSON.ObjectId.isValid('a');
+  },
+  function objectid_isvalid_bestcase_false() {
+    BSON.ObjectId.isValid('g6e84ebdc96f4c0772f0cbbf');
+  },
+  function objectid_isvalid_worstcase_false() {
+    BSON.ObjectId.isValid('66e84ebdc96f4c0772f0cbbg');
+  },
+  function objectid_isvalid_true() {
+    BSON.ObjectId.isValid('66e84ebdc96f4c0772f0cbbf');
+  }
+];
+
+// Add benchmarks here:
+export const benchmarks = [...ObjectId_isValid];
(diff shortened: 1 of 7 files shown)


test/bench/custom/main.mjs
```diff
@@ -0,0 +1,39 @@
...
+  .on('cycle', function logBenchmark(event) {
+    console.log(String(event.target));
+  })
+  .on('complete', function outputPerfSend() {
+    const data = Array.from(this).map(bench => ({
+      info: { test_name: bench.name },
+      metrics: [{ name: 'ops_per_sec', value: bench.hz }]
+    }));
+    console.log(util.inspect(data, { depth: Infinity, colors: true }));
+    fs.writeFileSync('results.json', JSON.stringify(data), 'utf8');
```
W-A-James, line 37: Can we rename this to something like `customBenchmarkResults.json` so it's easier to check the results and build tables locally if need be?

W-A-James (changes requested):

test/bench/custom/main.mjs
nbbeeken: done!

W-A-James (dismissed):

test/bench/custom/benchmarks.mjs
```diff
@@ -0,0 +1,20 @@
+/* eslint-disable strict */
+import { BSON } from '../../../lib/bson.mjs';
+
+const ObjectId_isValid = [
+  function objectid_isvalid_strlen() {
+    BSON.ObjectId.isValid('a');
+  },
+  function objectid_isvalid_bestcase_false() {
```
dariakp, line 8: what makes this the best case? and the other the worst case?

nbbeeken: Most or least short-circuiting. Updated the test names and added comments to clarify!

dariakp: Test case names lgtm now - thanks!

W-A-James (approved):

PR #721, opened 2024-10-17
test(NODE-6438): check that BSON undefined is returned from deserialize

nbbeeken:
#### What is changing?

- Prevent deserialize from changing the value returned for BSON undefined
- Reorganize tests around ignoreUndefined

Yes

I noticed no test fails if I change `BSON_DATA_UNDEFINED` handling, seems worthy of coverage given how the serializer special cases that same value.


test/node/bson_undefined.test.ts
```diff
@@ -0,0 +1,106 @@
...
+    it('returns a javascript undefined value', () => {
+      const doc = BSON.deserialize(bsonDocWithUndefined);
+      expect(doc).to.have.own.property('a').that.is.undefined;
+    });
+  });
+
+  describe('when serialize is given a javascript object that contains undefined', () => {
+    describe('when ignoreUndefined is set to false', function () {
+      it('serializes to document with a set to BSON null (type=10)', () => {
+        const jsObject = BSON.deserialize(bsonDocWithUndefined);
```
baileympearson, line 26:
```suggestion
        const jsObject = { a: undefined };
```

Is there a reason we can't use an object literal?

test/node/bson_undefined.test.ts
```diff
@@ -0,0 +1,106 @@
...
+        const bytes = BSON.serialize(jsObject, { ignoreUndefined: false });
+        expect(bytes).to.have.lengthOf(8);
+        expect(bytes).to.have.own.property('4', BSON_DATA_NULL);
```
baileympearson, line 29: What is happening in this assertion?  Does `haveOwnProperty` check for index access if the element is a Buffer?

Could we use parseToElementsToArray to re-parse the serialized bytes and confirm the serialized bytes contain one BSON null?

test/node/bson_undefined.test.ts
```diff
@@ -0,0 +1,106 @@
...
+        const jsObject = BSON.deserialize(bsonDocWithUndefined);
+        const bytes = BSON.serialize(jsObject);
+        expect(bytes).to.deep.equal(Uint8Array.of(5, 0, 0, 0, 0));
+      });
+    });
+  });
+
+  describe('when undefined appears inside an array', function () {
+    describe('when ignoreUndefined is set to true', function () {
+      it('does not ignore undefined values', function () {
```
baileympearson, line 52:
```suggestion
      it('serializes undefined values as null', function () {
```

May as well be consistent with our description

baileympearson (changes requested):

test/node/bson_undefined.test.ts
nbbeeken: good point, fixed

test/node/bson_undefined.test.ts
nbbeeken: good use of parseToElements

test/node/bson_undefined.test.ts
nbbeeken: fixed

baileympearson (approved):

PR #727, opened 2024-11-13
fix(NODE-6536): Binary.read never returns number[] and reads beyond content

nbbeeken:
#### What is changing?

- Fix Binary.read return type to be correctly Uint8Array
- Use subarray instead of slice so there isn't inconsistent view/copy between node and web
- Prevent reading beyond serializable user data
- Remove dead code from deserializer (buffer.slice is always defined)
- Searched for our usages of slice when we mean to use subarray to get views not copies. This has no impact on Node, fixes browser environments making copies where we did not intend to

No

Correctness and consistency

### Binary's `read()` returns a view of Uint8Array

Binary's `read()` return type claimed it would return `number[]` or Uint8Array which was true in previous BSON versions that didn't _always_ store a Uint8Array on the buffer property like `Binary` does today.

`read()`'s length parameter did not respect the position value allowing reading bytes beyond the data that is actually stored in the Binary. This has been corrected.

Additionally this method returned a view in Node.js environemnets and a copy in Web environments. it has been fixed to always return a view. 

### DIFF
--- a/src/binary.ts
+++ b/src/binary.ts
@@ -61,8 +61,47 @@ export class Binary extends BSONValue {
   /** User BSON type */
   static readonly SUBTYPE_USER_DEFINED = 128;
 
+  /**
+   * The bytes of the Binary value.
+   *
+   * The format of a Binary value in BSON is defined as:
+   * ```txt
+   * binary	::= int32 subtype (byte*)
+   * ```
+   *
+   * This `buffer` is the "(byte*)" segment.
+   *
+   * Unless the value is subtype 2, then deserialize will read the first 4 bytes as an int32 and set this to the remaining bytes.
+   *
+   * ```txt
+   * binary	::= int32 unsigned_byte(2) int32 (byte*)
+   * ```
+   *
+   * @see https://bsonspec.org/spec.html
+   */
   public buffer: Uint8Array;
+  /**
+   * The binary subtype.
+   *
+   * Current defined values are:
+   *
+   * - `unsigned_byte(0)` Generic binary subtype
+   * - `unsigned_byte(1)` Function
+   * - `unsigned_byte(2)` Binary (Deprecated)
+   * - `unsigned_byte(3)` UUID (Deprecated)
+   * - `unsigned_byte(4)` UUID
+   * - `unsigned_byte(5)` MD5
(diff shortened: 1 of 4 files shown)


src/binary.ts
```diff
@@ -219,7 +257,7 @@ export class Binary extends BSONValue {
 
   toUUID(): UUID {
     if (this.sub_type === Binary.SUBTYPE_UUID) {
-      return new UUID(this.buffer.slice(0, this.position));
+      return new UUID(this.buffer.subarray(0, this.position));
```
baileympearson, line 260: I actually think that we should return a copy here, a new UUID should own its bytes.

src/binary.ts
```diff
@@ -160,16 +199,15 @@ export class Binary extends BSONValue {
...
    * @param length - the number of bytes to read.
    */
-  read(position: number, length: number): BinarySequence {
+  read(position: number, length: number): Uint8Array {
     length = length && length > 0 ? length : this.position;
-
-    // Let's return the data based on the type we have
-    return this.buffer.slice(position, position + length);
+    const end = position + length;
+    return this.buffer.subarray(position, end > this.position ? this.position : end);
```
baileympearson, line 210:
```suggestion
    return this.buffer.subarray(position, Math.min(this.position, end));
```

baileympearson (changes requested):

src/binary.ts
nbbeeken: I prefer the ternary to a function call

Should I look into the C preprocessor idea? 😉

src/binary.ts
nbbeeken: We can file a ticket about auditing copying/viewing but for now I am fixing that this is a view in node and copy in the browser, changing it to always copy has performance implications and shouldn't be done as a "fix" unless there's a bug caused by it. It would be a feature to "ensure uuid has it's own copy of bytes" if that's what we want from this API.

I can file that, does that sound good?

baileympearson: Yup, that sounds good.

baileympearson (approved):

src/binary.ts
nbbeeken: https://jira.mongodb.org/browse/NODE-6548

PR #730, opened 2024-11-14
feat(NODE-6537): add support for binary vectors

nbbeeken:
#### What is changing?

Add vector helper APIs to Binary.

No

New Binary Vector `sub_type` ease of interoperability with native types.

### BSON Binary Vector Support!

The `Binary` class has new helpers to assist with using the newly minted Vector `sub_type` of Binary `sub_type == 9` :tada:! For more on how these types can be used with MongoDB take a look at [How to Ingest Quantized Vectors](https://www.mongodb.com/docs/atlas/atlas-vector-search/ingest-quantized-vectors/)!

Here's a summary of the API:

```ts
class Binary {
  toInt8Array(): Int8Array;
  toFloat32Array(): Float32Array;
  toPackedBits(): Uint8Array;

  static fromInt8Array(array: Int8Array): Binary;
  static fromFloat32Array(array: Float32Array): Binary;
  static fromPackedBits(array: Uint8Array, padding: number = 0): Binary;
}
```

Relatively self-explanatory: each one supports converting **to** and constructing **from** a native Javascript data type that corresponds to one of the three vector types: `Int8`, `Float32`, `PackedBit`. 

#### Vector Bytes Format

When a Binary is `sub_type` 9 the first two bytes are set to important metadata about the vector.

- `binary.buffer[0]` - The `datatype` that indicates what the following bytes are.
- `binary.buffer[1]` - The `padding` amount, a value 0-7 that indicates how many bits to ignore in a `PackedBit` vector.

##### `static fromPackedBits(array: Uint8Array, padding: number = 0)`

When handling packed bits, the last byte may not be entirely used. For example, a PackedBit vector = `[0xFF, 0xF0]` with padding = `4` ignores those last four 0s making the bit vector logically equal to 12 ones.

```txt
    F    F    F    0
[1111 1111 1111]   // ignored: the four 0s are padding
```
> [!IMPORTANT]
> When using the `fromPackedBits` method to set your padding amount to avoid inadvertently extending your bit vector. 

##### Unpacking Bits :luggage:

Packed bits get special treatment with two styles of conversion methods to suit your vector-y needs. `toBits` will return individually addressable bits shifted apart into an array. `fromBits` takes the same format in reverse and packs the bits into bytes. 

Notice there is no argument to set the `padding`. That is because it can be determined by the array's length. Recall those 12 ones from the previous example, well, the padding has to be 4 to reach a multiple of 8. 

```ts
class Binary {
  toBits(): Int8Array;
  static fromBits(bits: ArrayLike<number>): Binary;
}
```

> [!CAUTION]
> We _highly_ encourage using _**ONLY**_ these methods to interact with vector data and avoid operating directly on the byte format. Other Binary class methods (`put()`, `write()` `read()`, and `value()`) and direct access of data in a Binary's `buffer` beyond the 1st index should only be used in _exceptional_ circumstances and with _extreme_ caution after closely consulting the [BSON Vector specification](https://github.com/mongodb/specifications/blob/master/source/bson-binary-vector/bson-binary-vector.md?rgh-link-date=2024-11-14T14%3A26%3A54Z).
>
> **Details to keep in mind**
> - A javascript engine's endianness is platform dependent whereas BSON is always in little-endian format so if viewing bytes as Float32s take care to re-order bytes as needed.
> - Int8 vectors are signed bytes but `read()` always returns unsigned bytes.
> - The vector data begins at offset `2`.  


src/binary.ts
```diff
@@ -310,6 +334,207 @@ export class Binary extends BSONValue {
...
+
+    for (let bitOffset = 0; bitOffset < bits.length; bitOffset++) {
+      const byteOffset = Math.floor(bitOffset / 8);
+      const bit = bits[bitOffset];
+
+      if (bit !== 0 && bit !== 1) {
+        throw new BSONError(
+          `Invalid bit value at ${bitOffset}: must be 0 or 1, found ${bits[bitOffset]}`
+        );
+      }
```
nbbeeken, line 500: This provides a reasonable guard against passing packed values instead of individual bits.

src/binary.ts
```diff
@@ -310,6 +334,207 @@ export class Binary extends BSONValue {
...
+
+  /** @internal */
+  static validateVector(vector: Binary): void {
+    if (vector.sub_type !== this.SUBTYPE_VECTOR) return;
+
+    const size = vector.position;
+    const d_type = vector.buffer[0];
+    const padding = vector.buffer[1];
+
+    if (
```
nbbeeken, line 519: These are errors tested by the spec, our library must not produce (serialize, stringify) vectors that violate these conditions.

.evergreen/run-big-endian-test.sh
```diff
@@ -1,5 +1,22 @@
-#!/usr/bin/env bash
+#! /usr/bin/env bash
```
addaleax, line 1: I know this is completely insignificant but that makes it even more interesting how this change became part of the diff :) Did some autoformatter do this?

src/binary.ts
```diff
@@ -310,6 +325,213 @@ export class Binary extends BSONValue {
...
+
+    return new this(bytes, Binary.SUBTYPE_VECTOR);
+  }
+
+  /** @internal */
+  static validateVector(vector: Binary): void {
+    if (vector.sub_type !== this.SUBTYPE_VECTOR) return;
+
+    const size = vector.position;
+    const d_type = vector.buffer[0];
```
addaleax, line 513: Are we not doing validation for other `d_type` values than the 3 listed ones for future extensibility? (That would make sense to me)

src/binary.ts
```diff
@@ -310,6 +325,213 @@ export class Binary extends BSONValue {
...
+    const padding = vector.buffer[1];
```
addaleax, line 514: Should there be validation that the padding exists, or some type of check that ensures the vector contains at least the two metadata bytes? (I'd lean towards "no" but wanted to check)

src/binary.ts
```diff
@@ -310,6 +325,213 @@ export class Binary extends BSONValue {
     const subTypeArg = inspect(this.sub_type, options);
     return `Binary.createFromBase64(${base64Arg}, ${subTypeArg})`;
   }
+
+  /**
+   * If this Binary represents a Int8 Vector,
+   * returns a copy of the bytes in a new Int8Array.
```
addaleax, line 331: I'd mention in the doc comments that these methods throw if they are not of the right subtype and ... idk, subsubtype or whatever we call it?

src/binary.ts
```diff
@@ -310,6 +325,213 @@ export class Binary extends BSONValue {
...
+      for (let i = 2; i < binaryBytes.byteLength; i += 4) {
+        const byte0 = binaryBytes[i];
+        const byte1 = binaryBytes[i + 1];
+        const byte2 = binaryBytes[i + 2];
+        const byte3 = binaryBytes[i + 3];
+        binaryBytes[i] = byte3;
+        binaryBytes[i + 1] = byte2;
+        binaryBytes[i + 2] = byte1;
+        binaryBytes[i + 3] = byte0;
+      }
```
addaleax, line 458: Do you want to make use of the Node.js `Buffer.prototype.swapXX` methods if they are available by moving this to `NumberUtils`/`ByteUtils`?

A couple of advantages to that:
- This is already duplicating code a bit and it might be nice to use this as an opportunity to encapsulate the whole "if big endian then swap" logic in both places out of the way
- Once you start supporting 64-bit types this becomes really noisy
- The Node.js implementation uses a C++ helper for larger buffers

src/binary.ts
```diff
@@ -310,6 +325,213 @@ export class Binary extends BSONValue {
...
+    buffer.set(array, 2);
+    return new this(buffer, this.SUBTYPE_VECTOR);
+  }
+
+  /**
+   * Constructs a Binary representing an Packed Bit Vector.
+   * @param array - The array of 1s and 0s to pack into the Binary instance
+   */
+  public static fromBits(bits: ArrayLike<number>): Binary {
+    const byteLength = Math.ceil(bits.length / 8);
```
addaleax, line 482:
```suggestion
    const byteLength = (bits.length + 7) >>> 3; // ceil(bits.length / 8)
```

if nothing else, it lets you avoid the JS engine converting between integers and doubles here. If you feel like the potential overflow is a concern, I'd just add an explicit validation check

src/binary.ts
```diff
@@ -310,6 +325,213 @@ export class Binary extends BSONValue {
...
+    const bytes = new Uint8Array(byteLength + 2);
+    bytes[0] = Binary.VECTOR_TYPE.PackedBit;
+
+    const remainder = bits.length % 8;
+    bytes[1] = remainder === 0 ? 0 : 8 - remainder;
+
+    for (let bitOffset = 0; bitOffset < bits.length; bitOffset++) {
+      const byteOffset = Math.floor(bitOffset / 8);
```
addaleax, line 490:
```suggestion
      const byteOffset = bitOffset  >>> 3; // floor(bitOffset / 8)
```

.evergreen/run-big-endian-test.sh
```diff
@@ -1,5 +1,22 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+set -o xtrace
+
+# If you get an error you may have an outdated buildkit version
+# Try running this:
+# docker buildx rm builder && docker buildx create --name builder --bootstrap --use
+
+docker buildx build \
```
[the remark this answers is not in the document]
baileympearson: I'd prefer to run the tests on actual bigendian hardware, instead of spawning a docker container to run them.

.github/docker/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM node:22 AS build
+
+WORKDIR /bson
+COPY . .
+
+RUN rm -rf node_modules && npm install && npm test
+
+FROM scratch
+
+COPY --from=build /bson/docs/ /
```
baileympearson, line 10: What is the purpose of copying files out of the container?

baileympearson (changes requested):

test/node/binary.test.ts
```diff
@@ -249,4 +249,267 @@ describe('class Binary', () => {
...
+        binary.buffer[3] = 0;
+        binary.buffer[4] = 128;
+        binary.buffer[5] = 191;
+        expect(binary.toFloat32Array()).to.deep.equal(new Float32Array([-1]));
+      });
+
+      it('returns Float32Array when sub_type is vector and d_type is FLOAT32', () => {
+        const float32Array = new Float32Array([1.1, 2.2, 3.3]);
+        const binary = Binary.fromFloat32Array(float32Array);
+        expect(binary.toFloat32Array()).to.deep.equal(float32Array);
```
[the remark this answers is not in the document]
nbbeeken: It does check it

src/binary.ts
```diff
@@ -58,9 +68,18 @@ export class Binary extends BSONValue {
...
+  static readonly SUBTYPE_VECTOR = 9;
   /** User BSON type */
   static readonly SUBTYPE_USER_DEFINED = 128;
 
+  /** d_type of a Binary Vector (subtype: 9) */
+  static readonly VECTOR_TYPE = Object.freeze({
+    Int8: 0x03,
+    Float32: 0x27,
+    PackedBit: 0x10
+  } as const);
```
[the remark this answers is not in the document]
nbbeeken: If there are no strong feelings then we can leave it here.

src/utils/node_byte_utils.ts
```diff
@@ -159,5 +160,9 @@ export const nodeJsByteUtils = {
     return nodeJsByteUtils.toLocalBufferType(buffer).write(source, byteOffset, undefined, 'utf8');
   },
 
-  randomBytes: nodejsRandomBytes
+  randomBytes: nodejsRandomBytes,
+
+  swap32(buffer: Uint8Array): NodeJsBuffer {
+    return nodeJsByteUtils.toLocalBufferType(buffer).swap32();
```
addaleax, line 166: You should be able to just `Buffer.prototype.swap32.call(buffer);` this

nbbeeken: I would do that for all the functions in this file, but we use the explicit `toLocalBufferType` because of this:

```txt
> Buffer.prototype.toString.call(new Uint8Array([61]))
Uncaught TypeError: this.utf8Slice is not a function
```

It seems like swap32 to works so we don't necessarily have to blanket avoid calling prototype methods like you suggest but I'm wary that Node core may feel entitled to only make swap32 work _only_ for Buffer instances. Am I being overly cautious?

addaleax: Well ... you add another typed array instance that way, but it's probably no big deal.

I'd honestly consider just fixing `.toString()` and adding tests to Node.js core to ensure that all of the methods work on Uint8Arrays myself, maybe I'll do that if I find a piece of extra time 😂

nbbeeken: > Well ... you add another typed array instance that way

Yea and I'm sure our whole library could get a nice perf boost if we removed all the unnecessary view creations. 


I can make this change for swap32 if we're confident it won't break but now I'm thinking maybe I file a ticket for a perf improvement where we feature detect the ability to skip over the `toLocalBufferType` for the potential future node versions where this is fixed for `toString` and tested all the others. (more indirection!).

There is cost added to startup time but we can "test" `prototype.X.call`-style calls don't throw and then lock in versions of the functions that do that instead 🤔

addaleax (dismissed):
No blocking concerns from me

src/utils/node_byte_utils.ts
nbbeeken: https://jira.mongodb.org/browse/NODE-6550

baileympearson (approved):

src/binary.ts
```diff
@@ -310,6 +325,213 @@ export class Binary extends BSONValue {
...
+
+      const shift = 7 - (bitOffset % 8);
+      bytes[byteOffset + 2] |= bit << shift;
+    }
+
+    return new this(bytes, Binary.SUBTYPE_VECTOR);
+  }
+
+  /** @internal */
+  static validateVector(vector: Binary): void {
```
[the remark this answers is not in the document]
baileympearson: It doesn't, the free function does.  But I also think that given a choice between equivalent Binary.validate(binary) and binary.validate(), the instance method is preferable.

PR #742, opened 2024-12-16
fix(NODE-6608): calculateObjectSize returns the wrong value for bigint

nbbeeken:
#### What is changing?

- Introduce a bigint condition to the switch statement in calculateObjectSize that adds 8 to the size
- Introduce a symbol condition that returns 0
- Introduce a default condition that throws an error if there are future js types that the library cannot understand

No

This fixes a bug where the serializer produces something of a different size than what the calculate function claims it will.

### Fix `calculateObjectSize` not accounting for `BigInt` value size

`BSON.calculateObjectSize` was missing a condition for `BigInt` values, meaning it did not account for them in the same way that it would for `Long` values. This has been corrected.

We also added a new default condition that will catch any new values that may be returned by `typeof` in the future and will throw an error rather than returning an inaccurate size.


test/node/parser/calculate_size.test.ts
```diff
@@ -24,4 +24,16 @@ describe('calculateSize()', () => {
       })
     ).to.throw(BSONVersionError, /Unsupported BSON version/i);
   });
+
+  it('returns 8 bytes (+7 meta bytes) size for a bigint value', function () {
```
addaleax, line 28:
```suggestion
  it('returns 8 bytes (+8 meta bytes) size for a bigint value', function () {
```

right? 8 for the bigint, and then 1 for the 'long' type tag, 1+1 for the name, 4 for the document length, 1 trailing nul byte to finish the document?

addaleax (approved):

addaleax (approved):

dariakp (approved):

PR #745, opened 2025-01-07
chore(NODE-6634): pin NPM to 10 when Node version is 18

durran:
### Description

Pins NPM to 10 with Node.js 18.

#### What is changing?

Updates dependency script to ensure the pinning.

None

NODE-6634

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -166,16 +166,17 @@ tasks:
           NODE_LTS_VERSION: 22
       - func: install dependencies
       - func: run tests
-  - name: node-tests-latest
-    tags: ["node"]
-    commands:
-      - func: fetch source
-        vars:
-          NODE_LTS_VERSION: latest
-      - func: install dependencies
-      - func: run tests
-        vars:
-          TEST_TARGET: node
+  # TODO(NOXE-6662): Fix tests on latest.
+  # - name: node-tests-latest
+  #   tags: ["node"]
+  #   commands:
+  #     - func: fetch source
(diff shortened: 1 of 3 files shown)


nbbeeken (approved):
LGTM, but some new issue has cropped up in latest. I imagine this would repro on main if we just re-ran the latest task so its external, can you just confirm and capture the issue in a ticket? 

Should we skip latest here to keep the CI green?

durran:
> LGTM, but some new issue has cropped up in latest. I imagine this would repro on main if we just re-ran the latest task so its external, can you just confirm and capture the issue in a ticket?
> 
> Should we skip latest here to keep the CI green?

I confirmed on main and added NODE-6662. I'll skip here when back on Wednesday - we also still need https://github.com/mongodb-labs/drivers-evergreen-tools/pull/573 to go in first so I can remove the direct clone from that branch here as well.

.evergreen/config.yml
```diff
@@ -166,16 +166,17 @@ tasks:
...
-    tags: ["node"]
-    commands:
-      - func: fetch source
-        vars:
-          NODE_LTS_VERSION: latest
-      - func: install dependencies
-      - func: run tests
-        vars:
-          TEST_TARGET: node
+  # TODO(NOXE-6662): Fix tests on latest.
```
nbbeeken, line 169:
```suggestion
  # TODO(NODE-6662): Fix tests on latest.
```

nbbeeken (approved):

PR #748, opened 2025-02-10
fix(NODE-6735, NODE-6711): add BSON vector validation to EJSON stringification, serialization and conversion to native types

nbbeeken:
#### What is changing?

NODE-6735 and NODE-6711

### Prevent the creation of incorrectly sized float32 vectors

This adds validation to our `BSON.serialize` and `EJSON.stringify` methods that will prevent creating float 32 vectors that are not a multiple of 4. Previously created vectors that do not meet this validation will still be `deserialized` and `parsed` so they can be fixed. 

Additionally, the `toFloat32Array()`, `toInt8Array()`, and `toPackedBits()` methods now perform the same validation that serialize does to prevent use of incorrectly formatted Binary vector values. (For example, a packed bits vector with more than 7 bits of padding)

Vectors of an incorrect length could only be made manually (directly constructing the bytes and calling `new Binary`). We recommend using `toFloat32Array` and `fromFloat32Array` when interacting with Vectors in MongoDB as they handle the proper creation and translation of this data type.  

### DIFF
--- a/test/node/bson_binary_vector.spec.test.ts
+++ b/test/node/bson_binary_vector.spec.test.ts
@@ -52,10 +52,8 @@ function dtypeToHelper(dtype_hex: string) {
       return 'fromPackedBits';
     case '0x03' /* int8 */:
       return 'fromInt8Array';
-      break;
     case '0x27' /* float32 */:
       return 'fromFloat32Array';
-      break;
     default:
       throw new Error(`Unknown dtype_hex: ${dtype_hex}`);
   }
--- a/test/node/bson_binary_vector.spec.test.ts
+++ b/test/node/bson_binary_vector.spec.test.ts
@@ -185,6 +185,22 @@ function testVectorInvalidBSONBytes(test: VectorTest, expectedErrorMessage: stri
       expect(thrownError?.message).to.match(new RegExp(expectedErrorMessage));
     });
 
+    if (toHelper === 'toPackedBits') {
+      it(`Binary.toBits() throw a BSONError`, function () {
+        let thrownError: Error | undefined;
+        const bin = BSON.deserialize(Buffer.from(test.canonical_bson!, 'hex'));
+
+        try {
+          bin.vector.toBits();
+        } catch (error) {
+          thrownError = error;
+        }
+
+        expect(thrownError, thrownError?.stack).to.be.instanceOf(BSONError);
+        expect(thrownError?.message).to.match(new RegExp(expectedErrorMessage));
+      });
+    }
+
     it(`EJSON.stringify() throw a BSONError`, function () {
       let thrownError: Error | undefined;
       const bin = BSON.deserialize(Buffer.from(test.canonical_bson!, 'hex'));
--- a/test/node/bson_binary_vector.spec.test.ts
+++ b/test/node/bson_binary_vector.spec.test.ts
@@ -176,7 +176,7 @@ function testVectorInvalidInputValues(test: VectorTest, expectedErrorMessage: st
 }
 
 function testVectorInvalidBSONBytes(test: VectorTest, expectedErrorMessage: string) {
-  describe('when creating a Binary Vector instance from invalid bytes', () => {
+  describe('when encoding a Binary Vector made from invalid bytes', () => {
     it(`BSON.serialize() throw a BSONError`, function () {
       let thrownError: Error | undefined;
       const bin = BSON.deserialize(Buffer.from(test.canonical_bson!, 'hex'));
--- a/test/node/bson_binary_vector.spec.test.ts
+++ b/test/node/bson_binary_vector.spec.test.ts
@@ -188,7 +188,7 @@ function testVectorInvalidBSONBytes(test: VectorTest, expectedErrorMessage: stri
   });
 }
 
-describe.only('BSON Binary Vector spec tests', () => {
+describe('BSON Binary Vector spec tests', () => {
   const tests: Record<string, VectorSuite> = Object.create(null);
 
   for (const file of fs.readdirSync(path.join(__dirname, 'specs/bson-binary-vector'))) {
--- a/test/node/bson_binary_vector.spec.test.ts
+++ b/test/node/bson_binary_vector.spec.test.ts
@@ -110,33 +110,71 @@ const invalidTestExpectedError = new Map()
   .set('INT8 with float inputs', 'unsupported_error')
   .set('Vector with float values PACKED_BIT', 'unsupported_error');
 
+const invalidTestsWhereHelpersDoNotThrow = new Set()
+  .add('FLOAT32 with padding')
+  .add('INT8 with padding');
+
 function testVectorInvalidInputValues(test: VectorTest, expectedErrorMessage: string) {
   describe('when creating a BSON Vector given invalid input values', () => {
-    it(`either BSON.serialize() or Binary.${dtypeToHelper(test.dtype_hex)}() throws a BSONError`, function () {
+    it(`BSON.serialize() throws a BSONError`, function () {
       let thrownError: Error | undefined;
+
+      let bin;
       try {
-        const bin = make(test.vector!, test.dtype_hex, test.padding);
+        bin = make(test.vector!, test.dtype_hex, test.padding);
+      } catch (error) {
+        thrownError = error;
+      }
+
+      if (thrownError?.message.startsWith('unsupported_error')) {
+        expect(
+          expectedErrorMessage,
+          'We expect a certain error message but got an unsupported error'
+        ).to.equal('unsupported_error');
+        return;
+      }
+
+      try {
         BSON.serialize({ bin });
       } catch (error) {
         thrownError = error;
       }
 
+      expect(thrownError, thrownError?.stack).to.be.instanceOf(BSONError);
+      expect(thrownError?.message).to.match(new RegExp(expectedErrorMessage));
+    });
+
+    it(`Binary.${dtypeToHelper(test.dtype_hex)}() throws a BSONError`, function () {
+      let thrownError: Error | undefined;
+      try {
+        make(test.vector!, test.dtype_hex, test.padding);
+      } catch (error) {
+        thrownError = error;
+      }
+
+      if (invalidTestsWhereHelpersDoNotThrow.has(test.description)) {
+        expect(thrownError).to.not.exist;
+        return;
+      }
+
       if (thrownError?.message.startsWith('unsupported_error')) {
         expect(
           expectedErrorMessage,
           'We expect a certain error message but got an unsupported error'
         ).to.equal('unsupported_error');
-      } else {
-        expect(thrownError, thrownError?.stack).to.be.instanceOf(BSONError);
-        expect(thrownError?.message).to.match(new RegExp(expectedErrorMessage));
+        return;
       }
+
+      expect(thrownError, thrownError?.stack).to.be.instanceOf(BSONError);
+      expect(thrownError?.message).to.match(new RegExp(expectedErrorMessage));
     });
 
-    it(`either EJSON.stringify() or Binary.${dtypeToHelper(test.dtype_hex)}() throws a BSONError`, function () {
+    it(`EJSON.stringify() throws a BSONError`, function () {
       let thrownError: Error | undefined;
+
+      let bin;
       try {
-        const bin = make(test.vector!, test.dtype_hex, test.padding);
-        BSON.EJSON.stringify({ bin });
+        bin = make(test.vector!, test.dtype_hex, test.padding);
       } catch (error) {
         thrownError = error;
       }
@@ -146,10 +184,17 @@ function testVectorInvalidInputValues(test: VectorTest, expectedErrorMessage: st
           expectedErrorMessage,
           'We expect a certain error message but got an unsupported error'
         ).to.equal('unsupported_error');
-      } else {
-        expect(thrownError, thrownError?.stack).to.be.instanceOf(BSONError);
-        expect(thrownError?.message).to.match(new RegExp(expectedErrorMessage));
+        return;
       }
+
+      try {
+        BSON.EJSON.stringify({ bin });
+      } catch (error) {
+        thrownError = error;
+      }
+
+      expect(thrownError, thrownError?.stack).to.be.instanceOf(BSONError);
+      expect(thrownError?.message).to.match(new RegExp(expectedErrorMessage));
     });
   });
 }
--- a/test/node/bson_binary_vector.spec.test.ts
+++ b/test/node/bson_binary_vector.spec.test.ts
@@ -9,7 +9,7 @@ const { toHex, fromHex } = BSON.onDemand.ByteUtils;
 type VectorHexType = '0x03' | '0x27' | '0x10';
 type VectorTest = {
   description: string;
-  vector?: (number | string)[];
+  vector?: number[];
   valid: boolean;
   dtype_hex: VectorHexType;
   padding?: number;
@@ -46,7 +46,22 @@ function fixBits(f: number | string): number {
   return f;
 }
 
-function make(vector: (number | string)[], dtype_hex: VectorHexType, padding?: number): Binary {
+function dtypeToHelper(dtype_hex: string) {
+  switch (dtype_hex) {
+    case '0x10' /* packed_bit */:
+      return 'fromPackedBits';
+    case '0x03' /* int8 */:
+      return 'fromInt8Array';
+      break;
+    case '0x27' /* float32 */:
+      return 'fromFloat32Array';
+      break;
+    default:
+      throw new Error(`Unknown dtype_hex: ${dtype_hex}`);
+  }
+}
+
+function make(vector: number[], dtype_hex: VectorHexType, padding?: number): Binary {
   let binary: Binary;
   switch (dtype_hex) {
     case '0x10' /* packed_bit */:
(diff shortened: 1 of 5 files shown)


test/node/bson_binary_vector.spec.test.ts
```diff
@@ -1,14 +1,15 @@
...
 import * as fs from 'fs';
 import * as path from 'path';
-import { BSON, BSONError, Binary } from '../register-bson';
+import { BSON, BSONError, Binary, EJSON } from '../register-bson';
 import { expect } from 'chai';
 
 const { toHex, fromHex } = BSON.onDemand.ByteUtils;
 
 type VectorHexType = '0x03' | '0x27' | '0x10';
 type VectorTest = {
```
baileympearson, line 10: Suggest: taking the time to codify the required test fields into the type system:

```
type VectorValidTest = {
  ...
  vector: number[],
  valid: true,
  cannonical_bson: true
}

type VectorInvalidTest = {
  ...
  valid: false,
  cannonical_bson: true
} 

type VectorTest = VectorValidTest | VectorInvalidTest;
``` 

Also, are there actually tests where we parse `vector` as `string[]`?

test/node/bson_binary_vector.spec.test.ts
```diff
@@ -147,47 +224,22 @@ describe('BSON Binary Vector spec tests', () => {
...
-                expectedErrorMessage,
-                'We expect a certain error message but got an unsupported error'
-              ).to.be.false;
-              this.skip();
-            }
-
-            expect(thrownError).to.be.instanceOf(BSONError);
-            expect(thrownError?.message).to.match(new RegExp(expectedErrorMessage));
-          });
+          if (test.vector != null && test.canonical_bson != null) {
```
baileympearson, line 227: suggest: undo this change. the changes here don't DRY up the tests (we have two test functions), obfuscate the actual test implementation and make it more challenging to compare with the driver's [specification](https://github.com/mongodb/specifications/tree/master/source/bson-binary-vector/tests#to-prove-correct-in-an-invalid-case-validfalse-one-must)

baileympearson (changes requested):

test/node/bson_binary_vector.spec.test.ts
nbbeeken: I am not seeing the pattern that would clearly split the type between valid and invalid, because there are invalid int8 tests that have a vector field but float32 tests don't. That's why there's logic to test if vector exists we test making a binary out of it, then (in)valid determines whether we're looking for an error or not

test/node/bson_binary_vector.spec.test.ts
nbbeeken: - if the vector field is present, raise an exception when attempting to encode a document from the numeric values,
    dtype, and padding.
- if the canonical_bson field is present, raise an exception when attempting to deserialize it into the corresponding
    numeric values, as the field contains corrupted data.

The if statements here line up with these spec bullets. "testVectorBuilding" is the encoding bullet, I can update the naming to match that, I think encoding is a bit of a bloated term, but the focus is on the API that creates the binary not what our parsers / serializers return.

baileympearson: That is true.  But the additional describe blocks and organization obfuscates the actual tests.

My perspective on this issue comes from my experience reviewing these changes and comparing them with the spec.  I found it less straightforward to review than the previous implementation.

nbbeeken: Before:
```
float32
      ...
      invalid
        ✔ bson: FLOAT32 with padding
        ✔ extended json: FLOAT32 with padding
```

After:
```
BSON Binary Vector spec tests
    float32
      ...
      invalid
        both manual vector building and re-serializing
          creating an instance of a Binary class using parameters
            ✔ bson: FLOAT32 with padding
            ✔ ejson: FLOAT32 with padding
          creating an instance of a Binary class using canonical_bson
            ✔ bson deserialize: FLOAT32 with padding
            ✔ ejson stringify: FLOAT32 with padding
        vector re-serializing
          creating an instance of a Binary class using canonical_bson
            ✔ bson deserialize: Insufficient vector data with 3 bytes FLOAT32
            ✔ ejson stringify: Insufficient vector data with 3 bytes FLOAT32
        vector re-serializing
          creating an instance of a Binary class using canonical_bson
            ✔ bson deserialize: Insufficient vector data with 5 bytes FLOAT32
            ✔ ejson stringify: Insufficient vector data with 5 bytes FLOAT32
```

The additional describe blocks are meant to make our intention clearer so that when there's a breakage its clearer what component is means to work a certain way and how it is tested. So for example before when it was "invalid bson" it didn't inform you what part of BSON that meant was invalid. Now it make a distinction between "creating an instance of a Binary ( using parameters | using canonical bson )" 

Happy to make those titles clearer, the point is to draw out the distinction between the API users invoke when making a Binary vector 'by hand' and when we create it from bytes. 

This information isn't really captured in the spec titles and I don't know if it could be given that one test serves for multiple assertions. The distinction is important because it is where great care must be taken to not introduce a breaking change where existing vectors in the server fail to deserialize (rinse repeat for existing EJSON).

baileympearson: Okay, if I understand what you're saying - this change is largely to make the test output clearer?  If so, I think we could reorganize a bit.  Could we group by test first, followed by assertions?  I find that easier to understand. 

Before:
```
    int8
      invalid
        manual vector building
          creating an instance of a Binary class using parameters
            - bson: Overflow Vector INT8
            - ejson: Overflow Vector INT8
        manual vector building
          creating an instance of a Binary class using parameters
            - bson: Underflow Vector INT8
            - ejson: Underflow Vector INT8
        both manual vector building and re-serializing
          creating an instance of a Binary class using parameters
            ✔ bson: INT8 with padding
            ✔ ejson: INT8 with padding
          creating an instance of a Binary class using canonical_bson
            ✔ bson deserialize: INT8 with padding
            ✔ ejson stringify: INT8 with padding
        manual vector building
          creating an instance of a Binary class using parameters
            - bson: INT8 with float inputs
            - ejson: INT8 with float inputs
```

After:
```
int8
  invalid
    overflow vector int8
        instantiation from Array<number> does not throw
        BSON serialization throws
        EJSON serialization throws
...
    INT8 with padding
        instantiation from Array<number> does not throw
        instantiation from Uint8Array does not throw
        BSON serialization throws
        EJSON serialization throws
```

(I also rephrased `manual vector building` to `instantiation from Array<number>`).

While I was re-writing these tests, I noticed the "re-serialization" tests aren't mentioned anywhere in the spec?  I think they're attempting to test

```
if the canonical_bson field is present, raise an exception when attempting to deserialize it into the corresponding numeric values, as the field contains corrupted data.
```

But the actual test looks like:
```
const bin = BSON.deserialize(Buffer.from(test.canonical_bson!, 'hex'));

try {
BSON.serialize(bin);
} catch (error) {
thrownError = error;
}
```

Which is throwing the error during serialization, not deserialization.  I think maybe what this is supposed to be testing is our `toXArray` helpers instead of serialize?

baileympearson: Also, while we're re-structuring tests, is it possible to 

1. add a skip reason to the skipped tests
2. omit them entirely

It isn't clear from the test output that some tests are intentionally skipped because they don't actually exist.  Ex:

```
    int8
      invalid
        manual vector building
          creating an instance of a Binary class using parameters
            - bson: Overflow Vector INT8
            - ejson: Overflow Vector INT8
```

nbbeeken:
```
BSON Binary Vector spec tests
    float32
      valid
        Simple Vector FLOAT32
          ✔ calling Binary.fromFloat32Array() with input numbers and serializing it does not throw
          ✔ creating a Binary instance from BSON bytes does not throw
        Vector with decimals and negative value FLOAT32
          ✔ calling Binary.fromFloat32Array() with input numbers and serializing it does not throw
          ✔ creating a Binary instance from BSON bytes does not throw
      invalid
        FLOAT32 with padding
          when creating a Binary Vector instance from invalid bytes
            ✔ BSON.serialize() throw a BSONError
            ✔ EJSON.stringify() throw a BSONError
          when creating a BSON Vector given invalid input values
            ✔ either BSON.serialize() or Binary.fromFloat32Array() throws a BSONError
            ✔ either EJSON.stringify() or Binary.fromFloat32Array() throws a BSONError
```

I think I've got a better title approach that captures what you're suggesting, let me know. And I removed the skipping

nbbeeken: The toXArray methods have been updated to perform the same validation, do we think this change is safe?

test/node/bson_binary_vector.spec.test.ts
```diff
@@ -49,7 +46,22 @@ function fixBits(f: number | string): number {
...
+  switch (dtype_hex) {
+    case '0x10' /* packed_bit */:
+      return 'fromPackedBits';
+    case '0x03' /* int8 */:
+      return 'fromInt8Array';
+      break;
+    case '0x27' /* float32 */:
+      return 'fromFloat32Array';
+      break;
+    default:
```
baileympearson, line 59: dead code
```suggestion
      return 'fromInt8Array';
    case '0x27' /* float32 */:
      return 'fromFloat32Array';
    default:
```

test/node/bson_binary_vector.spec.test.ts
baileympearson: I'm not sure honestly.  I think you could argue it is a fix - we should have been validating the arrays we provided users so we don't give them corrupted data.

But also, assuming users do have invalid vectors now, this means they can't access the vectors' underlying bytes to fix the invalid bytes in their vectors.

baileympearson (changes requested):

test/node/bson_binary_vector.spec.test.ts
nbbeeken: If the validations related to corrupted data, I would feel confident it was a fix, however, specifically, the only additions here are padding fields being set for vector types that it shouldn't be set for.

> assuming users do have invalid vectors now, this means they can't access the vectors' underlying bytes to fix the invalid bytes

We do not validate in the Binary ctor or in the deserializer intentionally so that values could be fixed. So if we keep the validations toFloat32Array will throw if the padding field is non-zero, one could just do `binary.buffer[1] = 0` to make toFloat32Array work again and it would fix the data if they write it back to the db.

test/node/bson_binary_vector.spec.test.ts
nbbeeken: done!

test/node/bson_binary_vector.spec.test.ts
nbbeeken: Here are the cases we're considering:

These currently DO throw, but a RangeError:

  2) BSON Binary Vector spec tests float32 invalid **Insufficient vector data with 3 bytes FLOAT32** when creating a Binary Vector instance from invalid bytes Binary.toFloat32Array() throw a BSONError
  3) BSON Binary Vector spec tests float32 invalid **Insufficient vector data with 5 bytes FLOAT32** when creating a Binary Vector instance from invalid bytes Binary.toFloat32Array() throw a BSONError


These ones are just padding on a vector type that doesn't support it:

  1) BSON Binary Vector spec tests float32 invalid **FLOAT32 with padding** when creating a Binary Vector instance from invalid bytes Binary.toFloat32Array() throw a BSONError
  4) BSON Binary Vector spec tests int8 invalid **INT8 with padding** when creating a Binary Vector instance from invalid bytes Binary.toInt8Array() throw a BSONError
  5) BSON Binary Vector spec tests packed_bit invalid **Padding specified with no vector data PACKED_BIT** when creating a Binary Vector instance from invalid bytes Binary.toPackedBits() throw a BSONError
  
This one has no impact on `toPackedBits` because regardless of the padding value you still return the same number of bytes, however, this would be important to assert before `toBits()`:
 
  6) BSON Binary Vector spec tests packed_bit invalid **Exceeding maximum padding PACKED_BIT** when creating a Binary Vector instance from invalid bytes Binary.toPackedBits() throw a BSONError

baileympearson (changes requested):

test/node/bson_binary_vector.spec.test.ts
```diff
@@ -87,21 +97,131 @@ const invalidTestExpectedError = new Map()
...
+  .set('Overflow Vector PACKED_BIT', 'unsupported_error')
+  .set('Underflow Vector PACKED_BIT', 'unsupported_error')
+  .set('Overflow Vector INT8', 'unsupported_error')
+  .set('Underflow Vector INT8', 'unsupported_error')
+  .set('INT8 with float inputs', 'unsupported_error')
+  .set('Vector with float values PACKED_BIT', 'unsupported_error');
+
+function testVectorInvalidInputValues(test: VectorTest, expectedErrorMessage: string) {
+  describe('when creating a BSON Vector given invalid input values', () => {
+    it(`either BSON.serialize() or Binary.${dtypeToHelper(test.dtype_hex)}() throws a BSONError`, function () {
```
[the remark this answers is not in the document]
nbbeeken: I'm a little confused, this is testing the `fromX` helpers so what logs are you referring to?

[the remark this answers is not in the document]
baileympearson: > I'm a little confused, this is testing the fromX helpers so what logs are you referring to?

It also isn't just testing the from helpers - if I comment out the serialize call in the test, 5 tests fail because no error was thrown.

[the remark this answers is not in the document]
nbbeeken: Gotcha but how do I create a vector from the number array in the tests in order to test serialize without calling the `fromX` helpers?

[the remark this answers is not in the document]
nbbeeken: >if I comment out the serialize call in the test,

I meant this is testing the `fromX` and serialize as opposed to the `toX` ones you were mentioning, that's a different suite. But I don't see how you can separate the two, and logically that seems like it isn't the goal of the test:

- raise an exception when attempting to encode a document from the numeric values

test/node/bson_binary_vector.spec.test.ts
```diff
@@ -114,88 +115,68 @@ const invalidTestsWhereHelpersDoNotThrow = new Set()
...
+): { status: 'returned'; result: T } | { status: 'thrown'; result: Error } {
+  try {
+    return { status: 'returned', result: fn() };
+  } catch (error) {
+    return { status: 'thrown', result: error };
+  }
+}
+
 function testVectorInvalidInputValues(test: VectorTest, expectedErrorMessage: string) {
   describe('when creating a BSON Vector given invalid input values', () => {
```
baileympearson, line 124: suggest: breaking the vector instantiation from the serialization tests.

```
 BSON Binary Vector spec tests
    float32
      invalid
        FLOAT32 with padding
          when creating a Binary Vector instance from invalid bytes
            ✔ BSON.serialize() throw a BSONError
            ✔ Binary.toFloat32Array() throw a BSONError
            ✔ EJSON.stringify() throw a BSONError
```
to
```
 BSON Binary Vector spec tests
    float32
      invalid
        FLOAT32 with padding
          when creating a Binary Vector instance from invalid bytes
            ✔ Binary.toFloat32Array() throw a BSONError
          when serializing a Binary Vector instance constructed with invalid bytes
            ✔ BSON.serialize() throw a BSONError
            ✔ EJSON.stringify() throw a BSONError
```

serialization/stringification do not construct vectors, the test titles are misleading.

test/node/bson_binary_vector.spec.test.ts
```diff
@@ -87,21 +97,152 @@ const invalidTestExpectedError = new Map()
...
+    const errorType = expectedErrorMessage === 'unsupported_error' ? Error : BSONError;
+    const errorName = expectedErrorMessage === 'unsupported_error' ? 'Error' : 'BSONError';
+
+    const check = outcome => {
+      expect(outcome).to.exist;
+      expect(outcome.status).to.equal('thrown');
+      expect(outcome.result).to.be.instanceOf(errorType);
+      expect(outcome.result)
+        .to.have.property('message')
+        .that.matches(new RegExp(expectedErrorMessage));
```
baileympearson, line 153:
```suggestion
      expect(outcome.result).to.match(new RegExp(expectedErrorMessage));
```

(optional) regular expressions convert their arguments to strings, so matching errors automatically asserts on the error message of the error

baileympearson (changes requested):

test/node/bson_binary_vector.spec.test.ts
baileympearson: Still seeing misleading titles in the tests I think.  Ex:

```
    packed_bit
      invalid
        Padding specified with no vector data PACKED_BIT
          when creating a Binary Vector instance from invalid bytes
            ✔ BSON.serialize() throw a BSONError
            ✔ Binary.toPackedBits() throw a BSONError
            ✔ Binary.toBits() throw a BSONError
            ✔ EJSON.stringify() throw a BSONError
```

nbbeeken: What's the misleading part? (note this comment thread is on the "invalid values" but the sample output is "invalid bytes")

We use the bytes that are "bad" in the test case to make a manually built Binary (to avoid our validations), then we test that the helper methods and the functions that would put this Binary back into the transmittable format (bson/ejson) throw an error to prevent transmission of an invalid Binary.

nbbeeken: From slack: we mention making an instance but that's not the part that errors its the turning it into some other form (ejson/bson/toX)

baileympearson (approved):

PR #749, opened 2025-02-13
fix(NODE-6074): Removes top-level await in bson with separate node and browser ESM bundles

lourd:
#### What is changing?

The bundling process has been changed to make two separate ESM bundles for browsers and Node.js respectively.

No

The main motivation for this change is that the top level async-await usage breaks hot refresh in Next.js when importing this library on the client, as documented in this issue https://github.com/vercel/next.js/issues/61452 with a reproduction in this repo https://github.com/flo-dao/repro-fast-refresh.

The code does execute fine, it's specifically that the hot refresh behavior is broken. This is very problematic on applications that are very stateful and leverage hot refresh to have a quick iteration cycle. The current state is that when changing any module that imports `"bson"` it causes a full-page refresh, and you lose all of the state driving the UI you were developing

This has been raised before in this repo in https://github.com/mongodb/js-bson/pull/669 with a corresponding ticket that was closed without resolution https://jira.mongodb.org/browse/NODE-6074.

### DIFF
--- a/etc/rollup/rollup-plugin-require-rewriter/require_rewriter.mjs
+++ b/etc/rollup/rollup-plugin-require-rewriter/require_rewriter.mjs
@@ -1,44 +1,46 @@
 import MagicString from 'magic-string';
 
-const CRYPTO_IMPORT_ESM_SRC = `const nodejsRandomBytes = await (async () => {
+const CRYPTO_IMPORT_ESM_SRC = `import { randomBytes as nodejsRandomBytes } from 'crypto';`;
+const BROWSER_ESM_SRC = `const nodejsRandomBytes = nodejsMathRandomBytes;`;
+const CODE_TO_REPLACE = `const nodejsRandomBytes = (() => {
     try {
-        return (await import('crypto')).randomBytes;`;
-
-export class RequireRewriter {
-  /**
-   * Take the compiled source code input; types are expected to already have been removed
-   * Look for the function that depends on crypto, replace it with a top-level await
-   * and dynamic import for the crypto module.
-   *
-   * @param {string} code - source code of the module being transformed
-   * @param {string} id - module id (usually the source file name)
-   * @returns {{ code: string; map: import('magic-string').SourceMap }}
-   */
-  transform(code, id) {
-    if (!id.includes('node_byte_utils')) {
-      return;
+        return require('crypto').randomBytes;
     }
-    if (!code.includes('const nodejsRandomBytes')) {
-      throw new Error(`Unexpected! 'const nodejsRandomBytes' is missing from ${id}`);
+    catch {
+        return nodejsMathRandomBytes;
     }
+})();`;
 
-    const start = code.indexOf('const nodejsRandomBytes');
-    const endString = `return require('crypto').randomBytes;`;
-    const end = code.indexOf(endString) + endString.length;
+export function requireRewriter({ isBrowser = false } = {}) {
+  return {
+    /**
+     * Take the compiled source code input; types are expected to already have been removed
+     * Look for the function that depends on crypto, replace it with a top-level await
+     * and dynamic import for the crypto module.
+     *
+     * @param {string} code - source code of the module being transformed
+     * @param {string} id - module id (usually the source file name)
+     * @returns {{ code: string; map: import('magic-string').SourceMap }}
+     */
+    transform(code, id) {
+      if (!id.includes('node_byte_utils')) {
+        return;
+      }
+      const start = code.indexOf(CODE_TO_REPLACE);
+      if (start === -1) {
+        throw new Error(`Unexpected! Code meant to be replaced is missing from ${id}`);
+      }
 
-    if (start < 0 || end < 0) {
-      throw new Error(
-        `Unexpected! 'const nodejsRandomBytes' or 'return require('crypto').randomBytes;' not found`
-      );
+      const end = start + CODE_TO_REPLACE.length;
+
+      // MagicString lets us edit the source code and still generate an accurate source map
+      const magicString = new MagicString(code);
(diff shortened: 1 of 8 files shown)


lourd:
Pinging @nbbeeken for review 🙏

dariakp:
@lourd Hi there, thanks for bringing this to our attention! Neal is unavailable, but the team will re-triage this issue next week and get back to you with a reply as soon as we determine the best course of action.

baileympearson:
Hey @lourd, thanks for opening this PR.  

The current package structure was an intentional decision to avoid separate ESM bundles for Node and the browser because the current ESM module can be run in both.  In our next major release, we plan to drop support for Node16 and Node18.  Node20+ support global crypto by default, which will allow us to remove the import entirely (and the top-level await in the ESM module by extension).  We'd really like to keep the existing bundle structure, if possible, until we can make this change.  We tend to make our major releases in late summer (roughly August).

We'd like to know more about the problems that this is causing for your users.

- How many users seem to be impacted?  It seems like a relatively narrow subset of users - users with data heavy client-side applications that also use BSON in the browser?
- What are the actual consequences of this issue?  If I understand correctly, it seems like it just reloads extra modules and would require that users re-navigate their application to get to the place they were before the refresh.
- Is there truly no workaround here?  @nbbeeken suggested perhaps targeting the CJS module and using a crypto polyfill might resolve the issue.  Or something else?

And finally - is the current experience acceptable for the next ~5ish months, knowing that a proper resolution to this issue is coming in bson@7.0?

lourd:
Hey @baileympearson, for sure, thanks for the consideration and context!

> - How many users seem to be impacted?

At least anyone using Next.js and the bson library on the client. I suspect that it really encompasses anyone using React+Fast Refresh with Webpack, as it's not Next.js-specific. Users are affected whether they import `bson` directly or indirectly. I haven't tried with other bundlers like Vite to see if they're affected, too. There's a good chance they may be, too.

> - What are the actual consequences of this issue?

Fast Refresh is broken on the pages where `bson` is imported. Fast Refresh is a huge boon to React devs, it keeps you in the flow and saves a lot of time from having to get the UI back to the state you're changing every time you tweak something. It's not a matter of just reloading extra modules. I know on my team at least this change has saved on the order of 30-60 minutes per day for the folks who were working on the complex interfaces in our application where Fast Refresh has been broken. This is a situation where it takes roughly 60 seconds to get the UI to a specific state, multiplied by 30-60 tweaks in that day.

> - Is there truly no workaround here? @nbbeeken suggested perhaps targeting the CJS module and using a crypto polyfill might resolve the issue. Or something else?

I tried targeting the CJS module as I saw suggested in your other issues first, by overriding the `resolve` of the webpack config, but that didn't work, it broke the build for some reason. Our current workaround is to apply the changes from this PR as a patch to the bson package in our project.

> And finally - is the current experience acceptable for the next ~5ish months, knowing that a proper resolution to this issue is coming in bson@7.0?

It's not acceptable for us, hence going ahead and patching the package locally.

What's the cost-benefit tradeoff of having separate ESM bundles for Node and the browser? It seems like a relatively trivial change.

rkistner:
I'm also affected by this, so adding my view. I'm part of a team maintaining libraries for browsers and react-native, that have the bson package as a dependency, and are users are affected by this issue.

For example, when using vite (which is quite common), users need to add the `vite-plugin-top-level-await` plugin just because of bson. Then there are been issues like [this](https://github.com/Menci/vite-plugin-top-level-await/issues/61) with the plugin, completely breaking the build in some cases.

I've seen this also cause issues with Angular's build system, although I can't find the details now.

Overall, there are generally workarounds, but this does cause pain for web users.

nbbeeken:
Hey @lourd and @rkistner, apologies for the momentary radio silence, we are going to consider the change. We must look into additional testing to ensure this change is safe for all the platforms and products we support. We have current items in flight that need the team's attention but this is next up on our agenda, so stay tuned, thank you for your patience

W-A-James:
Hi @lourd, thanks for the contribution!
To accept these changes, however, we'd need you to add testing specifically for the behaviour you're adding here. As it is now, we could add a top-level await to the browser bundle and none of the existing tests would fail so it would be easy for us to accidentally reintroduce this behaviour in future.

lourd:
Ok @W-A-James. I went ahead and pushed a commit removing the `topLevelAwait` config from the Webpack bundling test. But, to double check that was correct, I ran the bundling test on `main` without my changes, just removing that configuration, and the bundling worked without an issue.

Is there a different testing strategy I should use? One could go very blunt, writing a test that uses typescript or babel or something to create an AST of each of the files and throw an error if a top-level await statement exists, but that seems like overkill.

W-A-James (changes requested):
@lourd some quick changes here.

Also I'm seeing test failures in `release.test.ts`:
```
 1) Release bson-6.10.3.tgz
       should not have extraneous files:

      AssertionError: Extra files: package/lib/bson.browser.mjs.map, package/lib/bson.node.mjs.map, package/lib/bson.browser.mjs, package/lib/bson.node.mjs: expected [ 'package/lib/bson.browser.mjs.map', 'package/lib/bson.node.mjs.map', 'package/lib/bson.browser.mjs', 'package/lib/bson.node.mjs' ] to have a length of +0 but got 4
      + expected - actual

      -4
      +0

      at Context.<anonymous> (test/node/release.test.ts:107:40)
      at processImmediate (node:internal/timers:483:21)
```

Should be an easy fix of adding the new artefacts to the the `REQUIRED_FILES` array.

And as far as the regression testing, one way we could do this is to use the node's `[--experimental-print-required-tla](https://nodejs.org/api/cli.html#--experimental-print-required-tla)` flag on the bundle to check if it includes any top-level awaits.

nbbeeken (approved):
LGTM! Along with the changes we'll land on top of this to finalize the testing. Thank you @lourd for the contribution 🎉

baileympearson (approved):

PR #751, opened 2025-02-14
test(NODE-6756): add tags to benchmarks

W-A-James:
#### What is changing?
* Update lockfile to use new bson-bench version
* Add tags to granular benchmarks via `getTypeTestTags` and `getMixedTestTags` helper
* Add tags to spec benchmarks
* Add tags to custom benchmarks
* Add script to run baseline benchmark to package.json
* Update custom benchmark runner to support cpu baseline benchmark
* Update readme to reflect changes in custom benchmark runner

- [X] Ran `npm run check:lint` script


package-lock.json
W-A-James, line 1: regened lockfile to get latest version of bson-bench

package.json
```diff
@@ -102,9 +102,10 @@
     "check:tsd": "npm run build:dts && tsd",
     "check:web": "WEB=true mocha test/node",
     "check:web-no-bigint": "WEB=true NO_BIGINT=true mocha test/node",
-    "check:granular-bench": "npm run build:bench && node ./test/bench/etc/run_granular_benchmarks.js",
-    "check:spec-bench": "npm run build:bench && node ./test/bench/lib/spec/bsonBench.js",
-    "check:custom-bench": "npm run build && node ./test/bench/custom/main.mjs",
+    "check:granular-bench": "npm run build:bench && npm run check:baseline-bench && node ./test/bench/etc/run_granular_benchmarks.js",
+    "check:spec-bench": "npm run build:bench && npm run check:baseline-bench && node ./test/bench/lib/spec/bsonBench.js",
+    "check:custom-bench": "npm run build && npm run check:baseline-bench && node ./test/bench/custom/main.mjs",
+    "check:baseline-bench": "node ./test/bench/etc/cpuBaseline.js",
```
W-A-James, line 108: Running the baseline bench script before each of our benchmark tasks to ensure that the results file exists

test/bench/custom/benchmarks.mjs
```diff
@@ -1,22 +1,26 @@
...
+    },
+    /** wrong character at the end, could be the least short circuited code path */
+    function objectid_isvalid_invalid_hex_at_end() {
+      BSON.ObjectId.isValid('66e84ebdc96f4c0772f0cbbg');
+    },
+    function objectid_isvalid_valid_hex_string() {
+      BSON.ObjectId.isValid('66e84ebdc96f4c0772f0cbbf');
+    }
+  ]
+};
```
W-A-James, line 23: Changed format here to make it easy to add tags to benchmarks

test/bench/etc/cpuBaseline.js
```diff
@@ -0,0 +1,60 @@
...
+    const data = {};
+    for (const b of Array.from(this)) {
+      if (b.name !== 'cpuBaseline') continue;
+      data.name = b.name;
+      data.stats = b.stats;
+      data.times = b.times;
+      data.hz = b.hz;
+      data.count = b.count;
+      data.cycles = b.cycles;
+      data.megabytes_per_second = taskSize / b.stats.mean;
```
W-A-James, line 54: saving additional stats here in case they prove useful in future, happy to remove all but the `hz` (ops per second) and `megabytes_per_second` measurements

test/bench/etc/cpuBaseline.js
```diff
@@ -0,0 +1,60 @@
+'use strict';
+const fs = require('fs');
+const { sep } = require('path');
+const benchmark = require('benchmark');
+
+const stableRegionMean = 42.82;
```
W-A-James, line 6: Same value used in driver

test/bench/custom/main.mjs
```diff
@@ -20,20 +20,71 @@ const systemInfo = () =>
...
+      const { tags } = suiteConfig;
+      for (const bench of Array.from(suite)) {
+        const result = processBenchmarkResult(bench, tags, metadata);
+        result.metrics.push({
+          name: 'normalized_throughput',
+          value: bench.hz / cpuBaselineResult,
+          metadata
+        });
+        data.push(result);
+      }
```
W-A-James, line 62: Note here that the normalized throughput here is the ops_per_second of the actual benchmark over the ops_per_second of the baseline benchmark

test/bench/granular/binary.bench.ts
W-A-James, line 1: The changes across the <type>.bench.ts benchmarks should be the same:
* import the `getTypetestTags`
* create tags by calling `getTypeTestTags` on a per-document basis
* Add tags into each invocation of `Suite.task`

test/bench/custom/readme.md
```diff
@@ -28,9 +32,9 @@ The JSON emitted at the end of the benchmarks must follow our performance tracki
 The JSON must be an array of "`Test`"s:
 
 ```ts
-type Metric = { name: string, value: number }
+type Metric = { name: string, value: number, metadata: { improvement_diraction: 'up' | 'down' } }
```
nbbeeken, line 35:
```suggestion
type Metric = { name: string, value: number, metadata: { improvement_direction: 'up' | 'down' } }
```

test/bench/granular/date.bench.ts
```diff
@@ -17,15 +18,17 @@ async function main() {
   const suite = new Suite('Date');
   const testDocs = await getTestDocs('date');
 
-  for (const operation of OPERATIONS) {
-    for (const documentPath of testDocs) {
+  for (const documentPath of testDocs) {
+    const tags = getTypeTestTags(documentPath);
+    for (const operation of OPERATIONS) {
```
nbbeeken, line 23: Not immediately obvious what flipping the order of the loops does?

test/bench/spec/bsonBench.ts
```diff
@@ -1,77 +1,105 @@
...
 });
 
-suite.run().then(
-  () => {
-    const results = suite.results.map(result => {
+suite
+  .run()
+  .then(() => {
+    return readFile(join(__dirname, '..', '..', 'etc', 'cpuBaseline.json'), 'utf8');
+  }, console.error)
```
nbbeeken, line 84: The console.error's are risky no? don't we want to crash?

test/bench/spec/bsonBench.ts
```diff
@@ -1,77 +1,105 @@
...
+  .then(cpuBaseline => {
+    if (!cpuBaseline) throw new Error('could not find cpu baseline');
```
nbbeeken, line 86: Ah related to the other comment, if we just let the error throw then we'll know why it's missing maybe? (probably file not found, but who knows!)

test/bench/spec/bsonBench.ts
```diff
@@ -1,77 +1,105 @@
...
+      metrics: (PerfSendResult['metrics'][0] & { metadata?: Metadata })[];
+    }[];
+    const results = suiteResults.map(result => {
       const rv = { ...result };
       rv.metrics = rv.metrics.filter(metric => metric.type === 'MEAN');
       return rv;
     });
+
+    const metadata: Metadata = { improvement_direction: 'up' };
     // calculte BSONBench composite score
```
nbbeeken, line 103:
```suggestion
    // calculate BSONBench composite score
```

test/bench/spec/bsonBench.ts
```diff
@@ -1,77 +1,105 @@
...
+  tags: [ALERT_TAG]
 });
 
-suite.run().then(
-  () => {
-    const results = suite.results.map(result => {
+suite
+  .run()
+  .then(() => {
+    return readFile(join(__dirname, '..', '..', 'etc', 'cpuBaseline.json'), 'utf8');
```
nbbeeken, line 83: Can you move this into the block below, make it an `async () =>` function and `await` this? then stack traces are a bit nicer, nothing wrong with an async function chained to a promise.

package.json
nbbeeken, line 108: Totally optional suggestion, should we capture these necessary lifecycle steps in the subdirectories these are written in? if we add a package.json like the driver one has we can use `prestart` and `start`: `npm --prefix test/bench/XXX start`

Maybe it doesn't gain that much over this, feel free to stick with what we have!

test/bench/etc/run_granular_benchmarks.js
```diff
@@ -28,6 +28,9 @@ const DOCUMENT_ROOT = path.resolve(`${__dirname}/../documents`);
     .run()
     .catch(() => null);
 
+  // Check for benchmark results
+  const cpuBaselineData = require(`${__dirname}${path.sep}cpuBaseline.json`);
```
nbbeeken, line 32: `require` is already script relative, so "./cpuBaseline.json" should be all that you need

But also, if you have the `path` module, might as well use `path.join()` there may be additional rules to "joining" that that methods handles that are more than putting the `.sep` in place.

test/bench/custom/main.mjs
```diff
@@ -20,20 +24,71 @@ const systemInfo = () =>
...
+  };
+}
+
+let completedSuites = 0;
+function completeSuite() {
+  const metadata = { improvement_direction: 'up' };
+  if (++completedSuites >= collectedSuites.length) {
+    let cpuBaselineResults;
+    try {
+      cpuBaselineResults = JSON.parse(fs.readFileSync(`${__dirname}/../etc/cpuBaseline.json`));
```
nbbeeken, line 45: Optional, should be able to use await import here to get the same convenience require provides in the other benchmark. This works so no worries if you want to stay the course.

nbbeeken (changes requested):

nbbeeken:
The granular benchmarks timed out after 15min, it says the base commit takes 5min which I find a bit surprising, did we accidentally change iterations here? and should we do it intentionally later for a larger sample size? 😅

test/bench/granular/date.bench.ts
W-A-James: I just didn't want to call `getTypeTestTags` more often than we needed to since the only parameter is the document path. It's admittedly a micro-optimization since this is just the benchmark run setup

.evergreen/perf_send.mjs
```diff
@@ -0,0 +1,63 @@
...
+  revision_order_id: order,
+  build_variant: variant,
+  version_id: version
+} = process.env;
+
+
+fs.readFile(resultFile, { encoding: 'utf8' })
+  .then(
+    results => {
+      const body = {
```
nbbeeken, line 29: Let's use async/await

.evergreen/perf_send.mjs
```diff
@@ -0,0 +1,63 @@
...
+          "Content-Type": "application/json"
+        },
+        body: JSON.stringify(body)
+      })
+    }
+  )
+  .then(async resp => {
+    console.log(resp);
+    if (resp.status !== 200) {
+      throw new Error(`Got status code: ${resp.status}\nResponse body: ${await resp.json()}`);
```
nbbeeken, line 56: If it's not 200, will the reply be json? if we end having to debug this I want to be able to see the content so maybe just `.text()` in case this isn't JSON

.evergreen/perf_send.mjs
```diff
@@ -0,0 +1,63 @@
...
+    }
+    console.log("Successfully posted results");
+  })
+  .catch(e => {
+    console.error(e);
+    process.exit(1);
+  });
```
nbbeeken, line 63: I'm guessing this is for testing, we should be able to just let this throw to the top-level.

.evergreen/perf_send.mjs
```diff
@@ -0,0 +1,63 @@
...
+  project,
+  task_id,
+  task_name,
+  revision_order_id: order,
+  build_variant: variant,
+  version_id: version
+} = process.env;
+
+
+fs.readFile(resultFile, { encoding: 'utf8' })
```
nbbeeken, line 26:
```suggestion
fs.readFile(resultFile, 'utf8')
```

nbbeeken (changes requested):

nbbeeken:
evergreen retry

nbbeeken (approved):

PR #752, opened 2025-02-18
fix(NODE-6764): incorrect negative bigint handling

W-A-James:
### Description

Correct regression to bigint deserialziation introduced in #649 

### ⚠️ Fixed potential data corruption bug when `useBigInt64` is enabled

After refactoring to improve deserialization performance in #649, we inadvertently broke the deserialization of negative Longs into `bigint`. This issue has been fixed here.

Thanks to @rkistner for reporting this bug!

- [X] Ran `npm run check:lint` script

### DIFF
--- a/src/utils/number_utils.ts
+++ b/src/utils/number_utils.ts
@@ -90,7 +90,8 @@ export const NumberUtils: NumberUtils = {
       eslint-disable-next-line no-restricted-globals
       -- This is allowed since this helper should not be called unless bigint features are enabled
      */
-    return (BigInt(hi) << BigInt(32)) + BigInt(lo);
+    const intermediate = (BigInt(hi) << BigInt(32)) + BigInt(lo);
+    return BigInt.asIntN(64, intermediate);
   },
 
   /** Reads a little-endian 64-bit float from source */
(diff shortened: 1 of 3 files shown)


rkistner (approved):
Thanks for the quick fix! The change looks good to me, and should fix the issue I had

test/node/bigint.test.ts
```diff
@@ -105,6 +105,39 @@ describe('BSON BigInt support', function () {
 
       it(description, test);
     }
+
+    describe('edge case tests', function () {
```
durran, line 109: Just a little nit. Instead of just calling these edge case tests can we better describe what we are testing here?

durran (changes requested):

src/utils/number_utils.ts
```diff
@@ -87,10 +87,12 @@ export const NumberUtils: NumberUtils = {
     const hi = NumberUtils.getUint32LE(source, offset + 4);
 
     /*
-      eslint-disable-next-line no-restricted-globals
+      eslint-disable no-restricted-globals
       -- This is allowed since this helper should not be called unless bigint features are enabled
      */
-    return (BigInt(hi) << BigInt(32)) + BigInt(lo);
+    const intermediate = (BigInt(hi) << BigInt(32)) + BigInt(lo);
+    return BigInt.asIntN(64, intermediate);
```
nbbeeken, line 94: From slack, let's fix the bit math instead of using asIntN

nbbeeken (changes requested):

nbbeeken (approved):

PR #791, opened 2025-04-22
chore(NODE-6921): perf tests default to cwd and errexit

nbbeeken:

Use cwd for BSON by default


nbbeeken:
Testing error: https://spruce.mongodb.com/version/6807edaab356e500077d17c2/

test/bench/spec/bsonBench.ts
```diff
@@ -78,6 +78,8 @@ suite.task({
 });
 
 suite.run().then(async () => {
+  if (suite.errors.length) process.exitCode = 1;
```
baileympearson, line 81: Its best practice to immediately exit after setting `exitCode`.  There's nothing preventing something that runs after this from assigning exitCode to a different value.

test/bench/granular/common.ts
```diff
@@ -61,6 +61,7 @@ export function getMixedTestTags(documentPath: string) {
 export async function runSuiteAndWriteResults(suite: Suite) {
   const targetDirectory = path.resolve(`${__dirname}/../../etc`);
   await suite.run();
+  if (suite.errors.length) process.exitCode = 1;
```
baileympearson, line 64: (same comment here as below)

test/bench/granular/common.ts
```diff
@@ -71,7 +72,7 @@ export function readEnvVars(): { warmup: number; iterations: number; library: st
   const rv = {
     warmup: Number.isSafeInteger(envWarmup) && envWarmup > 0 ? envWarmup : 100_000,
     iterations: Number.isSafeInteger(envIterations) && envIterations > 0 ? envIterations : 10_000,
-    library: libraryPath ? `bson:${libraryPath}` : 'bson#main'
+    library: libraryPath ? `bson:${libraryPath}` : `bson:${process.cwd()}`
```
baileympearson, line 75: from the AC:

> How do we fix the fact that the test names include the branch "main" the name needs to remain the same in order to get history.

I think this will break the perf tooling's continuity.  We can default to `cwd()`, but we still need to name anything that runs on `main` as `main`.

baileympearson:
> Testing error: https://spruce.mongodb.com/version/6807edaab356e500077d17c2/

@nbbeeken why did the spec benchmarks not fail?

baileympearson (changes requested):

test/bench/spec/bsonBench.ts
nbbeeken: > Its best practice to immediately

Where are you getting this practice from?

> nothing preventing something that runs after this from assigning exitCode to a different value

That's correct, but may be desirable

test/bench/granular/common.ts
nbbeeken: This was already fixed in the last PR, if you open the trend charts you can see the history hasn't been restarted and there is no branch name in the tests

nbbeeken:
> why did the spec benchmarks not fail?

I hadn't added the if stmt to the spec runner when I ran that patch, if you check the changes tab. I can run another patch to confirm

test/bench/spec/bsonBench.ts
baileympearson: > > Its best practice to immediately

> Where are you getting this practice from? Mocha does not exit when a test fails

I don't know of anywhere that this is stated, this feels more like common sense.  Let me rephrase though, I meant mainly that if you intend to exit a process with an exit code, then you should exit immediately to ensure you exit with that intended code.

nbbeeken: I must lack some of that sense that is common 🙂

Node.js exits when the current exec stack is empty and there are no ref-ed resources in the event loop. When you `exit()`, you preempt remaining resources from handling their events and you cover up leaked resources that would keep Node.js running.

As a compromise, I think we should instead leave the error code setting up to Node the way we normally do by throwing an error, this also stops us from running the logic that comes after these checks to not perform unnecessary work.

baileympearson (approved):

PR #870, opened 2026-02-10
fix(NODE-7436): EJSON.stringify type signature

chdanielmueller:
#### Summary of Changes

This pull request enhances the flexibility and robustness of the `EJSON.stringify` function by supporting multiple parameter signature combinations, similar to `JSON.stringify`.
It introduces a new internal type guard to accurately distinguish between options and replacer parameters, and adds comprehensive tests to ensure correct behavior across all supported overloads.
It supports a previously advertised function signature which was not implemented. `EJSON.stringify(object, { relaxed: false }, 2)`

##### Notes for Reviewers

The tests should contain every previously possible and also previously advertised but not working combination of inputs.
The changes within the function have to do with the previous typescript signature which has been used but did not work. See https://github.com/mongodb/js-bson/blob/0712bb15653093766a80bbd4ba104353cd4581e3/test/bench/etc/generate_documents.ts#L27

Off Topic: Documented within [NODE-7437](https://jira.mongodb.org/browse/NODE-7437)
> I noticed that the array replacement signature for JSON.stringify replaces the serialized keys.
> ```typescript
> const testDoc = {
>   objectId: ObjectId.createFromHexString('111111111111111111111111')
> };
> 
> const result = EJSON.stringify(testDoc, ['objectId', '$oid']) // {"objectId":{"$oid":"111111111111111111111111"}}
> const result2 = EJSON.stringify(testDoc, ['objectId']) // {"objectId":{}}
> ```
> Maybe if the replacer is an Array it should add the data types to the array.

### `EJSON.stringify` now correctly handles all parameter combinations

Previously, calling `EJSON.stringify(value, options, space)` silently discarded the `space` parameter, producing unformatted output. This has been fixed so all documented call signatures work as expected:

```typescript
// This now correctly produces indented canonical EJSON
EJSON.stringify(doc, { relaxed: false }, 2);
```

Thanks to @chdanielmueller for working on this problem!

### DIFF
--- a/src/extended_json.ts
+++ b/src/extended_json.ts
@@ -59,6 +59,30 @@ export type EJSONParseOptions = EJSONOptionsBase & {
 export type EJSONOptions = EJSONSerializeOptions & EJSONParseOptions;
 
 /** @internal */
+function isEJSONSerializeOptions(value: unknown): value is EJSONSerializeOptions {
+  if (value == null || typeof value !== 'object') {
+    return false;
+  }
+    
+  // Check that all properties, if present, are of the correct type
+  if ('legacy' in value && typeof value.legacy !== 'boolean') {
+    return false;
+  }
+  if ('relaxed' in value && typeof value.relaxed !== 'boolean') {
+    return false;
+  }
+  if ('ignoreUndefined' in value && typeof value.ignoreUndefined !== 'boolean') {
+    return false;
+  }
+  
+  // Check that there are no invalid properties (only known EJSON serialize options)
+  const validKeys = ['legacy', 'relaxed', 'ignoreUndefined'];
+  const keys = Object.keys(value);
+  
+  return keys.every(key => validKeys.includes(key));
+}
+
+/** @internal */
 type BSONType =
   | Binary
   | Code
@@ -464,28 +488,45 @@ function parse(text: string, options?: EJSONParseOptions): any {
  *
  * // prints '{"int32":10}'
  * console.log(EJSON.stringify(doc));
+ * 
+ * // prints '{"int32":{"$numberInt":"10"}}' with 2 space indentation
+ * console.log(EJSON.stringify(doc, { relaxed: false }, 2));
  * ```
  */
+function stringify(value: any, replacer?: (number | string)[] | ((this: any, key: string, value: any) => any) | null, space?: string | number, options?: EJSONSerializeOptions): string;
+function stringify(value: any, replacer?: (number | string)[] | ((this: any, key: string, value: any) => any) | null, options?: EJSONSerializeOptions): string;
+function stringify(value: any, options?: EJSONSerializeOptions, space?: string | number): string;
 function stringify(
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   value: any,
-  replacer?:
+  replacerOrOptions?:
     | (number | string)[]
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
     | ((this: any, key: string, value: any) => any)
+    | null
     | EJSONSerializeOptions,
-  space?: string | number,
+  spaceOrOptions?: string | number | EJSONSerializeOptions,
   options?: EJSONSerializeOptions
 ): string {
-  if (space != null && typeof space === 'object') {
-    options = space;
-    space = 0;
-  }
-  if (replacer != null && typeof replacer === 'object' && !Array.isArray(replacer)) {
-    options = replacer;
+  let replacer: ((this: any, key: string, value: any) => any) | (number | string)[] | null | undefined;
+  let space: string | number | undefined;
+  
+  // Check if second parameter is options
+  if (isEJSONSerializeOptions(replacerOrOptions)) {
+    options = replacerOrOptions;
     replacer = undefined;
-    space = 0;
+    space = typeof spaceOrOptions === 'number' || typeof spaceOrOptions === 'string' ? spaceOrOptions : undefined;
   }
+  // Check if third parameter is options
+  else if (isEJSONSerializeOptions(spaceOrOptions)) {
+    options = spaceOrOptions;
+    replacer = replacerOrOptions;
+    space = undefined;
+  }
+  // Standard case: replacer, space, options
+  else {
+    replacer = replacerOrOptions;
+    space = typeof spaceOrOptions === 'number' || typeof spaceOrOptions === 'string' ? spaceOrOptions : undefined;
+  }
+  
   const serializeOptions = Object.assign({ relaxed: true, legacy: false }, options, {
     seenObjects: [{ propertyName: '(root)', obj: null }]
   });
--- a/src/extended_json.ts
+++ b/src/extended_json.ts
@@ -63,7 +63,7 @@ function isEJSONSerializeOptions(value: unknown): value is EJSONSerializeOptions
   if (value == null || typeof value !== 'object') {
     return false;
   }
-    
+
   // Check that all properties, if present, are of the correct type
   if ('legacy' in value && typeof value.legacy !== 'boolean') {
     return false;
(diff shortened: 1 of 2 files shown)


PavelSafronov:
@chdanielmueller , thanks so much for your contribution! The team is going to prioritize this PR and [NODE-7436](https://jira.mongodb.org/browse/NODE-7436) at our next triage session.

src/extended_json.ts
```diff
@@ -58,6 +58,30 @@ export type EJSONParseOptions = EJSONOptionsBase & {
...
+  }
+  if ('relaxed' in value && typeof value.relaxed !== 'boolean') {
+    return false;
+  }
+  if ('ignoreUndefined' in value && typeof value.ignoreUndefined !== 'boolean') {
+    return false;
+  }
+
+  // Check that there are no invalid properties (only known EJSON serialize options)
+  const validKeys = ['legacy', 'relaxed', 'ignoreUndefined'];
```
tadjik1: This creates a maintenance footgun with to compiler warning to catch if `EJSONSerializeOptions` gains a new property, or if detection silently breaks. In this case the new property would be treated as a replacer and passed into JSON.stringify.

src/extended_json.ts
```diff
@@ -58,6 +58,30 @@ export type EJSONParseOptions = EJSONOptionsBase & {
...
+  // Check that all properties, if present, are of the correct type
+  if ('legacy' in value && typeof value.legacy !== 'boolean') {
+    return false;
+  }
+  if ('relaxed' in value && typeof value.relaxed !== 'boolean') {
+    return false;
+  }
+  if ('ignoreUndefined' in value && typeof value.ignoreUndefined !== 'boolean') {
+    return false;
+  }
```
tadjik1: The strict key check causes silent misbehavior on typos. If something like `{relaxd: false}` will be passed - this guard returns `false` and the object will silently be forwarded to `JSON.stringify` as a replacer - producing wrong output, the existing `typeof replacer === 'object' && !Array.isArray(replacer)` would correctly treat it as options. This is more forgiving and safer.

test/node/extended_json.test.ts
```diff
@@ -733,4 +733,153 @@ describe('Extended JSON', function () {
       expect(() => EJSON.stringify(input)).to.throw(BSONError);
     });
   });
+
+  context('stringify: Parameter signature combinations', function () {
```
tadjik1, line 737: The test coverage here is great, and these tests are valuable independent of which implementation approach is chosen. Please keep them even with a simplified fix.

src/extended_json.ts
```diff
@@ -464,28 +489,65 @@ function parse(text: string, options?: EJSONParseOptions): any {
...
+    | null
+    | undefined;
+  let space: string | number | undefined;
+  /* eslint-enable @typescript-eslint/no-explicit-any */
+
+  // Check if second parameter is options
+  if (isEJSONSerializeOptions(replacerOrOptions)) {
+    options = replacerOrOptions;
     replacer = undefined;
-    space = 0;
```
tadjik1, line 487: I think this is the original bug and removing just this line (avoid unconditional discarding the `space` value( would make everything work as expected.

src/extended_json.ts
```diff
@@ -464,28 +489,65 @@ function parse(text: string, options?: EJSONParseOptions): any {
  *
  * // prints '{"int32":10}'
  * console.log(EJSON.stringify(doc));
+ *
+ * // prints '{"int32":{"$numberInt":"10"}}' with 2 space indentation
+ * console.log(EJSON.stringify(doc, { relaxed: false }, 2));
  * ```
  */
 function stringify(
```
tadjik1, line 473: These overloads add type-level documentation of supported call patterns, but they also add maintenance hurdle. Since the runtime checks are needed regardless, and the existing signature with the union type already communicates that options can go in the second position - this doesn't provide enough value to justify the complexity.

tadjik1 (changes requested):
hi @chdanielmueller, thanks a lot for the contribution. The bug is real, well-documented, and the test coverage you added is great.
The core issue is that `space = 0` discards the user's `space` value when options are passed in the `replacer` position.

However, the fix is significantly more complex than it needs to be. The entire bug can be fixed by deleting one line (`space = 0)`. The existing code already treats any non-array object in replacer position as options.

I would suggest minimal fix:
```diff
if (replacer != null && typeof replacer === 'object' && !Array.isArray(replacer)) {
  options = replacer;
  replacer = undefined;
- space = 0
}
```

The overloads, type guard, and restructured function body are unnecessary for this fix and introduce maintenance overhead.

test/node/extended_json.test.ts
chdanielmueller: Thanks, keeping the tests as requested.

src/extended_json.ts
chdanielmueller: That's a great catch. I did not think of that.

I simplified the type guard to the old implementation.
I added a new test to validate the behavior in the presence of typos and to ensure this decision is explicitly captured.
see d3047c3461e05949bee1a3db80bcdfda5941bed5

src/extended_json.ts
chdanielmueller: Removed by d3047c3461e05949bee1a3db80bcdfda5941bed5

src/extended_json.ts
chdanielmueller: I appreciate the concern about complexity. From my perspective, the overloads provide meaningful value: they catch invalid parameter combinations (e.g., passing space before options when replacer is omitted) at compile-time and make IDE autocomplete show three clear patterns instead of a confusing union type.
In my opinion the ~15 lines of declarations add negligible maintenance cost while significantly improving type safety and DX.

src/extended_json.ts
chdanielmueller: Type guard removed completely in ee148581986ec5c1bfdcc2ef83fcd84bb400a903

src/extended_json.ts
chdanielmueller: Whilst I think, that my approach would have improved readability for the next developer I see your concern.
I did revert the implementation to the previous implementation but kept the new variable names as an indicator that it can be either or.
ee148581986ec5c1bfdcc2ef83fcd84bb400a903

test/node/extended_json.test.ts
```diff
@@ -733,4 +733,163 @@ describe('Extended JSON', function () {
...
+  "int32Number": {
+    "$numberInt": "300"
+  },
+  "name": "test"
+}`);
+    });
+
+    it('should work with string space parameter', function () {
+      const result = EJSON.stringify(testDoc, null, '\t', { relaxed: false });
+      expect(result).to.equal(`{
```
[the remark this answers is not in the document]
chdanielmueller: Only primitive values are supported. This adheres to the implementation of JSON.stringify

chdanielmueller:
Hi @tadjik1 
Thank you for the feedback and the review. I did address all your comments with the new changes with one exception: https://github.com/mongodb/js-bson/pull/870#discussion_r2863395998
Giving it back to you for another review 😄

tadjik1:
hi @chdanielmueller, thanks a lot for updating PR 👍 Just wanted to confirm that this work is not lost, and it's tracked. I will re-review changes as soon as possible.

src/extended_json.ts
tadjik1: Thanks, agree 👍

src/extended_json.ts
```diff
@@ -443,6 +443,7 @@ function parse(text: string, options?: EJSONParseOptions): any {
   });
 }
 
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
tadjik1, line 446: For the team review: please, note, this is necessary as replacer has `any` defined in multiple places (as `value` parameter, `this`, and as an return value). It can be done differently by "extracting" replacer function into separate type, so disable any would work for 1 line, but in my opinion it won't provide much value.

tadjik1 (dismissed):

PavelSafronov (dismissed):
Changes look good, thanks for this!

nbbeeken (dismissed):

tadjik1 (approved):

PavelSafronov (approved):

PR #880, opened 2026-04-13
feat(NODE-4793): add Symbol.toStringTag to all BSON types

sarthaksoni25:
#### Summary of Changes
Adds `Symbol.toStringTag` to all BSON types by implementing the getter in the `BSONValue` base class, delegating to `this._bsontype`. Since `_bsontype` is abstract on `BSONValue` and implemented by every subclass, all 13 BSON types inherit the correct tag automatically.

Before:
Object.prototype.toString.call(new ObjectId()) // '[object Object]'

After:
Object.prototype.toString.call(new ObjectId()) // '[object ObjectId]'

#### Notes for Reviewers
Follows the same pattern as the existing `bsonType` symbol getter added in NODE-7255. The implementation is a single getter on `BSONValue` — no changes needed in individual subclasses.

- [x] Self-review completed
- [x] PR title follows the correct format: `feat(NODE-4793): add Symbol.toStringTag to all BSON types`
- [x] Changes are covered by tests (14 new tests added in `bson_type_classes.test.ts`)
- [x] No new TODOs

### DIFF
--- a/test/node/bson_type_classes.test.ts
+++ b/test/node/bson_type_classes.test.ts
@@ -98,6 +98,17 @@ describe('BSON Type classes common interfaces', () => {
           ));
       }
 
+      if (TypeClass.name !== 'UUID') {
+        it(`defines Symbol.toStringTag equal to its name`, () =>
+          expect(TypeClass.prototype).to.have.property(Symbol.toStringTag, TypeClass.name));
+      } else {
+        it(`UUID inherits Symbol.toStringTag from Binary`, () =>
+          expect(Object.getPrototypeOf(TypeClass.prototype)).to.have.property(
+            Symbol.toStringTag,
+            'Binary'
+          ));
+      }
+
       it(`defines a Symbol.for('@@mdb.bson.version') property equal to 7`, () =>
         expect(TypeClass.prototype).to.have.property(Symbol.for('@@mdb.bson.version'), 7));
 
--- a/src/bson_value.ts
+++ b/src/bson_value.ts
@@ -29,6 +29,9 @@ export abstract class BSONValue {
   public get [bsonType](): this['_bsontype'] {
     return this._bsontype;
   }
+  public get [Symbol.toStringTag](): string {
+    return this._bsontype;
+  }
 
   /** @internal */
   get [BSON_VERSION_SYMBOL](): typeof BSON_MAJOR_VERSION {


test/node/bson_type_classes.test.ts
```diff
@@ -98,6 +98,17 @@ describe('BSON Type classes common interfaces', () => {
...
 
+      if (TypeClass.name !== 'UUID') {
+        it(`defines Symbol.toStringTag equal to its name`, () =>
+          expect(TypeClass.prototype).to.have.property(Symbol.toStringTag, TypeClass.name));
+      } else {
+        it(`UUID inherits Symbol.toStringTag from Binary`, () =>
+          expect(Object.getPrototypeOf(TypeClass.prototype)).to.have.property(
+            Symbol.toStringTag,
+            'Binary'
+          ));
```
tadjik1, line 109: Approach is right - matches the existing `bsonType` getter pattern on `BSONValue`. One suggestion: override `Symbol.toStringTag` on `UUID` to return `'UUID'`. `_bsontype` is a wire-protocol key and must stay `'Binary'` (serializer dispatches on it), but `Symbol.toStringTag` is a developer-facing identity hint with no wire impact - making it reflect the JS class gives nicer debug output (`[object UUID]`) without breaking anything. The test would then assert `toStringTag === TypeClass.name` uniformly across all types.

tadjik1 (changes requested):
hi @sarthaksoni25, thanks for you work, everything looks correct, simple and clean implementation, well done! 
I left one suggestion, please let me know what you think.

test/node/bson_type_classes.test.ts
nbbeeken:
```js
> Object.prototype.toString.call(Buffer.alloc(0))
'[object Uint8Array]'
```

It is not always the case that the string tag matches the class name, because UUID is a trivial subclass of Binary and Binary is the relevant BSON type it may make sense to keep these matching the `_bsontype` field. It also is not only a developer hint here's an example where it's a [load bearing type check](https://github.com/mongodb/js-bson/blob/0712bb15653093766a80bbd4ba104353cd4581e3/src/parser/utils.ts#L1-L25).

An additional benefit to adding this field is it adds the same functionality we added when we introduced the `@@mdb.bson.type` symbol but without the need to import/create something from the library.  

Just my opinion here :) not strongly held

addaleax:
Should this be considered semver-major?

tadjik1:
hi @sarthaksoni25, after discussing this change with the team we have decided against it. the main reason is that it can break existing functionality of external packages that might rely on the string representation of BSON instances. and potential benefits do not justify complexity of the migration.
Once again, thanks a lot for your contribution!

sarthaksoni25:
Thanks for the review and context, @tadjik1 and @nbbeeken! Makes sense given the semver-major risk. Happy to revisit when the time is right.

PR #881, opened 2026-04-13
fix(NODE-5116): validate null bytes in EJSON keys during serialization

sarthaksoni25:
#### Summary of Changes
Adds null byte validation to `EJSON.stringify` and `EJSON.serialize` when processing document keys. Previously, BSON serialization would throw a `BSONError` for keys containing null bytes, but EJSON serialization would silently produce output that could never be round-tripped back through BSON.

Before:
EJSON.stringify({ 'a\x00b': 1 }) // '{"a\u0000":1}' — silent, invalid output
BSON.serialize({ 'a\x00b': 1 })  // throws BSONError ✅

After:
EJSON.stringify({ 'a\x00b': 1 }) // throws BSONError ✅
EJSON.serialize({ 'a\x00b': 1 }) // throws BSONError ✅

#### Notes for Reviewers
The check is added in `serializeDocument` in `extended_json.ts`, mirroring the existing null byte check already present in `EJSON.parse`. The error message is identical to the one used in `parse` for consistency.

- [x] Self-review completed
- [x] PR title follows the correct format: `fix(NODE-5116): validate null bytes in EJSON keys during serialization`
- [x] Changes are covered by tests (3 new tests added in `extended_json.test.ts`)
- [x] No new TODOs

### DIFF
--- a/src/extended_json.ts
+++ b/src/extended_json.ts
@@ -351,6 +351,11 @@ function serializeDocument(doc: any, options: EJSONSerializeInternalOptions) {
     // It's a regular object. Recursively serialize its property values.
     const _doc: Document = {};
     for (const name of Object.keys(doc)) {
+      if (name.indexOf('\x00') !== -1) {
+        throw new BSONError(
+          `BSON Document field names cannot contain null bytes, found: ${JSON.stringify(name)}`
+        );
+      }
       options.seenObjects.push({ propertyName: name, obj: null });
       try {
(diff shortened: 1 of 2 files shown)


dariakp:
Hi @sarthaksoni25 - thanks so much for your submission! Unfortunately, since this is a breaking change, we cannot accept your contribution at this point in time. We can definitely revisit this when the team is ready to work on our next major version release, but for now I am closing this PR.

PR #884, opened 2026-05-06
fix(NODE-7575): reset ObjectId state when building startup snapshot

addaleax:
#### Summary of Changes

Use the relevant Node.js API to ensure that different processes indeed have different counter and process-unique values in their ObjectId classes, as intended, even when using Node.js's startup snapshot feature.

### ObjectId generates unique values across Node.js startup snapshot processes

`ObjectId` now resets its random state (process-unique bytes and counter seed) when a process restores from a Node.js [startup snapshot](https://nodejs.org/api/v8.html#startup-snapshot-api). Previously these values were frozen into the snapshot blob, causing processes restored from the same snapshot to generate colliding ObjectIds.

### DIFF
--- a/src/objectid.ts
+++ b/src/objectid.ts
@@ -4,9 +4,6 @@ import { type InspectFn, defaultInspect } from './parser/utils';
 import { ByteUtils } from './utils/byte_utils';
 import { NumberUtils } from './utils/number_utils';
 
-// Unique sequence for the current process (initialized on first use)
-let PROCESS_UNIQUE: Uint8Array | null = null;
-
 /** ObjectId hexString cache @internal */
 const __idCache = new WeakMap(); // TODO(NODE-6549): convert this to #__id private field when target updated to ES2022
 
@@ -33,7 +30,28 @@ export class ObjectId extends BSONValue {
   }
 
   /** @internal */
-  private static index = Math.floor(Math.random() * 0xffffff);
+  private static index = 0;
+
+  /** Unique sequence for the current process (initialized on first use)
+   * @internal
+   */
+  private static PROCESS_UNIQUE: Uint8Array | null = null;
+
+  /** @internal */
+  private static resetState = (): void => {
+    this.index = Math.floor(Math.random() * 0x1000000);
+    this.PROCESS_UNIQUE = null;
+  };
+
+  static {
+    this.resetState();
+    // https://nodejs.org/api/v8.html#startup-snapshot-api
+    // @ts-expect-error Node.js types not present since this is an optional API
+    const { startupSnapshot } = globalThis?.process?.getBuiltinModule('v8') ?? {};
+    if (startupSnapshot?.isBuildingSnapshot()) {
+      startupSnapshot?.addDeserializeCallback(this.resetState);
+    }
+  }
 
   static cacheHexString: boolean;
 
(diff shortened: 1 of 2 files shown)


src/objectid.ts
```diff
@@ -33,7 +30,28 @@ export class ObjectId extends BSONValue {
...
+  private static resetState = (): void => {
+    this.index = Math.floor(Math.random() * 0x1000000);
+    this.PROCESS_UNIQUE = null;
+  };
+
+  static {
+    this.resetState();
+    // https://nodejs.org/api/v8.html#startup-snapshot-api
+    // @ts-expect-error Node.js types not present since this is an optional API
+    const { startupSnapshot } = globalThis?.process?.getBuiltinModule('v8') ?? {};
```
addaleax, line 50: Accessing the API this way was recommended here: https://github.com/nodejs/node/pull/63138#issuecomment-4384154881

I would agree that this seems like a reasonable compromise

src/objectid.ts
```diff
@@ -33,7 +30,28 @@ export class ObjectId extends BSONValue {
...
+  private static index = 0;
+
+  /** Unique sequence for the current process (initialized on first use)
+   * @internal
+   */
+  private static PROCESS_UNIQUE: Uint8Array | null = null;
+
+  /** @internal */
+  private static resetState = (): void => {
+    this.index = Math.floor(Math.random() * 0x1000000);
```
addaleax, line 42: Drive-by fix: `Math.random()` never returns `1`, and `0xffffff` should be a valid counter value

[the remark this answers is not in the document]
addaleax: Good catch! `getInc()` also needs an update, done

test/node/startup_snapshot.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    const [oid1, oid2] = [stdout1[1], stdout2[1]].map(
+      oid => oid.match(/^(?<ts>\w{8})(?<uniq>\w{10})(?<counter>\w{6})$/)!.groups!
+    );
+
+    // Less than 20 seconds between timestamps should be plenty of leeway
+    expect(Math.abs(parseInt(oid2.ts, 16) - parseInt(oid1.ts, 16))).to.be.lessThan(20);
+    // Distinct process unique values
+    expect(oid1.uniq).to.not.equal(oid2.uniq);
+    // Distinct counter values
+    expect(oid1.counter).to.not.equal(oid2.counter);
```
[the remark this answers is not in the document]
addaleax: Yeah, this is one in 16 million CI runs that fails, I can live with that

nbbeeken (dismissed):

src/objectid.ts
```diff
@@ -33,7 +30,28 @@ export class ObjectId extends BSONValue {
...
+    this.PROCESS_UNIQUE = null;
```
PavelSafronov: We're setting `index` to a random value in this function, can we also set the PROCESS_UNIQUE value?
`this.PROCESS_UNIQUE = ByteUtils.randomBytes(5);`

addaleax: @PavelSafronov Sure, initializing this eagerly now: f8fe04c91b83859626c07f8a27cbd485cf49c23f

PavelSafronov (approved):

PR #886, opened 2026-05-20
fix(NODE-5937): rewrite deserializeObject as iterative, not recursive

PavelSafronov:
#### Summary of Changes

Rewrite `deserialize` method so it's no longer recursive.

The AC Neal wrote is pretty succinct:
> Replace recursive deserializeObject() with an iterative implementation using an explicit stack (array of frames with push/pop)
> Each stack frame holds the document being built, current buffer position, size, options, and a continuation describing what to do when the child completes (object assignment, array assignment + terminator check, or code-with-scope assembly)

So the new approach is to parse everything in a single pass, but when we are faced with nested objects, we change where the deserialized values are placed:
1. if there are no nested objects, the while-loop populates a single document with all of the fields
2. if there are nested objects, the while-loop parses their contents and puts the properties into a nested document, which itself is then added to the "original" document (which can be either the root object, or the preceding container object) when we get to the last index the current frame is expected to parse

Once I wrote the `deserialize` method, I used Claude to make similar changes to `serializeInto` (the frame stack approach is identical), `internalCalculateObjectSize`, which we labeled as stretch goals for this ticket. Claude's code was updated, modified, and commented by me.

EJSON has not been updated. EJSON is more involved and will be completed later. We have created a follow-up JIRA ticket for that: https://jira.mongodb.org/browse/NODE-7595

##### Notes for Reviewers

Latest `run-spec-benchmarks` run: https://spruce.corp.mongodb.com/task/node_bson_perf_run_spec_benchmarks_patch_ba29d33ff0ed5ba6fc63e598fe8b7e7879e4753f_6a10d1d207c6d40007b93b59_26_05_22_21_59_47/trend-charts?execution=0

Screenshot 2026-05-22 at 3 07 08 PM

zscore mostly went down, and even the downward move is within the +-5% of last time

Avoid stack overflow when parsing heavily-nested content.

### `BSON` operations no longer throw on deeply nested documents

Core serialization and deserialization function (`serialize`, `deserialize`, `calculateObjectSize`) have been rewritten as iterative algorithms, so a sufficiently-nested document will no longer exhaust the call stack.

#### Behavior change: Code with Scope scopes are no longer promoted to DBRef

As a side effect of the rewrite, a `Code with Scope` scope that happens to carry `$ref`, `$id`, and `$db` keys is no longer promoted to a DBRef instance. The scope is returned as a plain object, which is the correct behavior: a scope represents JavaScript variable bindings, not a document reference.

### DIFF
--- a/src/parser/deserializer.ts
+++ b/src/parser/deserializer.ts
@@ -125,12 +125,14 @@ export function internalDeserialize(
 interface NestedParsingFrame {
   // Document that we will fill out as we parse the nested object
   holdingDocument: Document;
+  // The name of the key we will set the parsed object on once we finish parsing the nested object, this is used in the onComplete callback to know where to set the parsed nested object in the parent document
+  propertyName: string | number;
   // The index in the buffer where the current object ends, used to know when we are done parsing the current object and can pop the stack
   lastIndex: number;
   // Whether the current frame is parsing an array, used to know whether to interpret keys as strings or array indices
   isArray: boolean;
   // The next array index to use if this frame is parsing an array, used to assign numeric keys to array elements without having to utf-8 decode the key from the buffer
-  arrayIndex?: number;
+  arrayIndex: number;
   // When true, all objects in this frame will be returned as raw bson buffers without parsing.
   // This is used when the fieldsAsRaw option is used on a parent object, and is inherited by nested frames.
   // It can also be set to true if the global raw option is set, but it cannot be set to true for a frame if the global raw option is false.
@@ -141,8 +143,15 @@ interface NestedParsingFrame {
   globalUTFValidation: boolean;
   // The utf-8 validation setting for this frame, used to determine whether to utf-8 validate keys in this frame. This is determined based on the global utf-8 validation setting and the specific keys specified in the validation option.
   validationSetting: boolean;
-  // A callback function that will be called when the frame is completed, used to set the value in the parent object once we have finished parsing the current object.
-  onComplete: () => void;
+  functionString: string | null; // only used for Code with Scope
+  // One of 3 supported types:
+  // - constants.BSON_DATA_OBJECT
+  // - constants.BSON_DATA_ARRAY
+  // - constants.BSON_DATA_CODE_W_SCOPE
+  elementType:
+    | typeof constants.BSON_DATA_OBJECT
+    | typeof constants.BSON_DATA_ARRAY
+    | typeof constants.BSON_DATA_CODE_W_SCOPE;
 }
 
 const allowedDBRefKeys = /^\$ref$|^\$id$|^\$db$/;
@@ -250,13 +259,41 @@ function deserializeObject(
 
   let isPossibleDBRef = isArray ? false : null;
 
+  // How we set the value in the destination object, with special handling for __proto__ to avoid prototype pollution vulnerabilities
+  const setValue = (name: string | number, value: unknown) => {
+    const dest =
+      nestedParsingStack.length === 0
+        ? rootObject
+        : nestedParsingStack[nestedParsingStack.length - 1].holdingDocument;
+    if (name === '__proto__') {
+      Object.defineProperty(dest, name, {
+        value,
+        writable: true,
+        enumerable: true,
+        configurable: true
+      });
+    } else {
+      dest[name] = value;
+    }
+  };
+
+  const toPotentialDbRef = (doc: Document): DBRef | Document => {
+    if (isDBRefLike(doc)) {
+      const copy = Object.assign({}, doc) as Partial<DBRefLike>;
+      delete copy.$ref;
+      delete copy.$id;
+      delete copy.$db;
+      const dbref = new DBRef(doc.$ref, doc.$id, doc.$db, copy);
+      return dbref;
+    }
+    return doc;
+  };
+
   // While we have more left data left keep parsing
   while (!done) {
     // Current frame, if one is present
     const currentFrame =
       nestedParsingStack.length === 0 ? null : nestedParsingStack[nestedParsingStack.length - 1];
-    // The object that we will be setting the current key value on, this is either the root object if we are at the top level, or the holdingDocument of the current frame if we are nested
-    const destObject = currentFrame === null ? rootObject : currentFrame.holdingDocument;
 
     // Read the type
     const elementType = buffer[index++];
@@ -269,7 +306,28 @@ function deserializeObject(
         if (index === currentFrame.lastIndex) {
           // Current index matches the last index of the frame, so we pop the frame and set the value in the parent document
           nestedParsingStack.pop();
-          currentFrame.onComplete();
+          // finish the frame
+          let result: Document = currentFrame.holdingDocument;
+          switch (currentFrame.elementType) {
+            case constants.BSON_DATA_OBJECT:
+              // if this is a DBRef, we need to construct a DBRef object instead of a plain object
+              if (currentFrame.isPossibleDBRef) {
+                result = toPotentialDbRef(result);
+              }
+              break;
+            case constants.BSON_DATA_ARRAY:
+              // nothing to do, the holding document is already an array and the keys were set as numeric indices
+              break;
+            case constants.BSON_DATA_CODE_W_SCOPE:
+              // the holding document is the scope, we need to construct a Code object with the function string and scope
+              result = new Code(currentFrame.functionString!, currentFrame.holdingDocument);
+              break;
+            default:
+              throw new BSONError('Unexpected element type in frame stack');
+          }
+          const propName = currentFrame.propertyName;
+          // set the value in the parent document
+          setValue(propName, result);
           continue;
         } else {
           // Current index does not match the last index of the frame, the document is malformed
@@ -302,23 +360,9 @@ function deserializeObject(
     // Represents the key
     const name = isArray
       ? currentFrame && currentFrame.isArray
-        ? currentFrame.arrayIndex!++
+        ? currentFrame.arrayIndex++
         : arrayIndex++
       : ByteUtils.toUTF8(buffer, index, i, false);
-
-    // How we set the value in the destination object, with special handling for __proto__ to avoid prototype pollution vulnerabilities
-    const setValue = (value: unknown) => {
-      if (name === '__proto__') {
-        Object.defineProperty(destObject, name, {
-          value,
-          writable: true,
-          enumerable: true,
-          configurable: true
-        });
-      } else {
-        destObject[name] = value;
-      }
-    };
 
     // shouldValidateKey is true if the key should be validated, false otherwise.
     // Within a nested frame the original code passed a collapsed boolean validation option,
@@ -396,23 +440,16 @@ function deserializeObject(
         isDeferredValue = true;
         nestedParsingStack.push({
           holdingDocument: {},
+          elementType: constants.BSON_DATA_OBJECT,
+          propertyName: name,
+          functionString: null,
           lastIndex: index + objectSize,
           isArray: false,
+          arrayIndex: 0,
           raw: false,
           isPossibleDBRef: null, // we don't know if this is a DBRef until we parse the keys, so we start with null and set to false if we encounter a key that is not valid for a DBRef
           globalUTFValidation: true,
-          validationSetting: shouldValidateKey,
-          onComplete: function () {
-            let result = this.holdingDocument;
-            if (this.isPossibleDBRef && isDBRefLike(result)) {
-              const copy = Object.assign({}, result) as Partial<DBRefLike>;
-              delete copy.$ref;
-              delete copy.$id;
-              delete copy.$db;
-              result = new DBRef(result.$ref, result.$id, result.$db, copy);
-            }
-            setValue(result);
-          }
+          validationSetting: shouldValidateKey
         });
         index = index + 4;
       }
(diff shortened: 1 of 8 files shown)


src/parser/deserializer.ts
```diff
@@ -288,45 +427,58 @@ function deserializeObject(
...
-      }
-      value = deserializeObject(buffer, _index, arrayOptions, true);
-      index = index + objectSize;
-
-      if (buffer[index - 1] !== 0) throw new BSONError('invalid array terminator byte');
-      if (index !== stopIndex) throw new BSONError('corrupted array bson');
+      // fieldsAsRaw match: push with raw=true so embedded objects inside come back as raw bytes.
+      // Also propagate raw from the parent frame (nested arrays inside a raw array stay raw).
+      const arrayRaw = !!(fieldsAsRaw && fieldsAsRaw[name]) || (currentFrame?.raw ?? false);
+      isDeferredValue = true;
```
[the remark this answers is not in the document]
tadjik1: This is real problem, I guess we can use the same mechanism as few lines above:
```
 const objectSize = NumberUtils.getInt32LE(buffer, index);

if (objectSize <= 0 || objectSize > buffer.length - index)
  throw new BSONError('bad embedded document length in bson');
}
```

src/parser/deserializer.ts
```diff
@@ -201,26 +244,108 @@ function deserializeObject(
...
+              break;
+            default:
+              throw new BSONError('Unexpected element type in frame stack');
+          }
+          // set the value in the parent document (setValue reads currentFrame, now pointing to parent)
+          setValue(completedFrame.propertyName, result);
+          continue;
+        } else {
+          // Current index does not match the last index of the frame, the document is malformed
+          throw new BSONError('Bad BSON Document: object not properly terminated');
```
tadjik1, line 333: Before, with recursion, in these cases we threw `corrupt array bson`, but now it seems we throw this error only for root, and "Bad BSON ..." - for nested. This is regression for error messages.

src/parser/deserializer.ts
```diff
@@ -130,6 +162,15 @@ function deserializeObject(
   options: DeserializeOptions,
   isArray = false
 ) {
+  // Settings configured from options parameter
+
+  // Strip the prototype chain so inherited properties don't affect option reads.
+  options = Object.assign({}, options);
```
nbbeeken: "Strip the prototype chain" is an interesting comment for an API that is notorious for having the power to modify the prototype 😅. 

this should work just the same and comes with less surprises
```js
options = { ...options }
```

src/parser/serializer.ts
```diff
@@ -528,60 +408,60 @@ function serializeSymbol(buffer: Uint8Array, key: string, value: BSONSymbol, ind
...
+function toKvPairs(object: Document): [string, unknown][] {
+  if (Array.isArray(object)) {
+    const kvPairs = new Array<[string, unknown]>(object.length);
+    for (let i = 0; i < object.length; i++) {
+      kvPairs[i] = [`${i}`, object[i]];
+    }
+    return kvPairs;
+  } else if (object instanceof Map || isMap(object)) {
+    return Array.from((object as Map<unknown, unknown>).entries()) as [string, unknown][];
+  } else {
```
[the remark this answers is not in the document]
PavelSafronov: `object` is a `Document`, which by definition has string properties.
We can add the recommended validation, but that will be a behavior change: we'll start to throw BSONError instead of throwing a TypeError or silently getting a corrupted value. My preference would be to keep the current behavior and follow up with an investigation and a potential behavior change.

src/parser/deserializer.ts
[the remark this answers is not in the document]
PavelSafronov: This is another behavior change: on `main` we don't make these validations and allow the parsing errors to happen later.
This changes the type of error that we're throwing, by making it more specific. I think it's worthwhile making this breaking change here, but we can also add a TODO and keep the current behavior.

src/parser/deserializer.ts
PavelSafronov: Excellent catch! One line fix to keep the current behavior, and adding tests to cover this.

src/parser/deserializer.ts
```diff
@@ -201,26 +244,111 @@ function deserializeObject(
...
+  };
+
+  const toPotentialDbRef = (doc: Document): DBRef | Document => {
+    if (isDBRefLike(doc)) {
+      const copy = Object.assign({}, doc) as Partial<DBRefLike>;
+      delete copy.$ref;
+      delete copy.$id;
+      delete copy.$db;
+      const dbref = new DBRef(doc.$ref, doc.$id, doc.$db, copy);
+      return dbref;
```
nbbeeken: Just drive-by modernization

```suggestion
      const copy = { $ref, $id, $db, ...fields } as Partial<DBRefLike>;
      return new DBRef($ref, $id, $db, fields);
```

src/parser/serializer.ts
```diff
@@ -528,60 +408,60 @@ function serializeSymbol(buffer: Uint8Array, key: string, value: BSONSymbol, ind
...
-    output.$db = value.db;
+function toKvPairs(object: Document): [string, unknown][] {
+  if (Array.isArray(object)) {
+    const kvPairs = new Array<[string, unknown]>(object.length);
+    for (let i = 0; i < object.length; i++) {
+      kvPairs[i] = [`${i}`, object[i]];
+    }
+    return kvPairs;
+  } else if (object instanceof Map || isMap(object)) {
+    return Array.from((object as Map<unknown, unknown>).entries()) as [string, unknown][];
```
nbbeeken:
```suggestion
    return Array.from((object as Map<unknown, unknown>)) as [string, unknown][];
```

entries is redundant right?

src/parser/serializer.ts
```diff
@@ -619,336 +499,184 @@ export function serializeInto(
     path = new Set();
   }
 
-  // Push the object to the path
   path.add(object);
 
-  // Start place to serialize into
+  const stack: SerializationFrame[] = [
+    {
+      kvPairs: toKvPairs(object),
```
nbbeeken: For Maps and Arrays we previously did not make an additional array before iterating them, I am wary of the added garbage collection, do you think there's a way we can still iterate the containers directly here?

src/parser/serializer.ts
```diff
@@ -528,60 +408,60 @@ function serializeSymbol(buffer: Uint8Array, key: string, value: BSONSymbol, ind
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    if (typeof (target as any)?.toBSON === 'function') {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      target = (target as any).toBSON() as Document;
+      if (target != null && typeof target !== 'object') {
+        throw new BSONError('toBSON function did not return an object');
+      }
+    }
+    const keys = Object.keys(target);
+    const kvPairs = new Array<[string, unknown]>(keys.length);
```
nbbeeken: In my experience, creating an Array with a size is going to make it sparse which has special handling since it's an "odd" case in most JS. It's almost always better to create `[]` and `push` to it so it remains dense

test/node/parser/serializer.test.ts
```diff
@@ -3,6 +3,22 @@ import { bufferFromHexArray } from '../tools/utils';
 import { expect } from 'chai';
 import { BSONVersionError } from '../../register-bson';
 
+function buildDeeplyNestedObject(depth: number): Record<string, unknown> {
+  let inner: Record<string, unknown> = {};
+  for (let i = 0; i < depth; i++) {
+    inner = { a: inner };
+  }
+  return inner;
+}
```
nbbeeken: theres a test utils file we could de-dupe this into

src/parser/deserializer.ts
PavelSafronov: Made a slightly different change, and added tests for this specific bloc.

src/parser/serializer.ts
PavelSafronov: Yup, dropping.

src/parser/serializer.ts
PavelSafronov: Fixed.

test/node/parser/serializer.test.ts
PavelSafronov: De-duped.

src/parser/serializer.ts
PavelSafronov: Yup, absolutely, reworking that now to use an iterator.

src/parser/serializer.ts
```diff
@@ -528,60 +408,54 @@ function serializeSymbol(buffer: Uint8Array, key: string, value: BSONSymbol, ind
...
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      target = (target as any).toBSON() as Document;
+      if (target != null && typeof target !== 'object') {
+        throw new BSONError('toBSON function did not return an object');
+      }
+    }
+    const keys = Object.keys(target);
+    for (let i = 0; i < keys.length; i++) {
+      yield [keys[i], target[keys[i]]];
+    }
```
addaleax: `yield* Object.entries(target)`? and that should even cover arrays without issue as well

addaleax (commented):
Regarding perf, I'd actually be pretty curious to see if it would help to make this a linked list (i.e. have a `prev` field or similar on the stack frames) instead of an array, but that seems like something that could be a separate thing to try out

src/parser/serializer.ts
PavelSafronov: Updated.

nbbeeken (dismissed):
Functionally lgtm, I think it would make sense to run the driver perf too for a complete picture if we want to see if anything over there shows signs of change. 

You should be able to put a commit into the package.json and run a patch 🤔 

but all-in-all seems like things weren't too complicated, nice work :)

src/parser/deserializer.ts
```diff
@@ -547,10 +698,6 @@ function deserializeObject(
       const _index = index;
       // Decode the size of the object document
       const objectSize = NumberUtils.getInt32LE(buffer, index);
-      // Decode the scope object
-      const scopeObject = deserializeObject(buffer, _index, options, false);
```
tadjik1, line 551: I think this part has new behaviour, locally I have ran the following test:
```
it('keeps scope as a plain object', () => {
  const oid = new BSON.ObjectId();
  const scope = { $ref: 'col', $id: oid, $db: 'db' };
  const original = new BSON.Code('function () {}', scope);
  const bytes = BSON.serialize({ code: original });
  const { code } = BSON.deserialize(bytes);
  expect(code).to.have.property('_bsontype', 'Code');
  expect(code.scope).to.not.have.property('_bsontype');
});
```
and it behaves differently on main vs. this branch. currently this test passes, but on main
```
keeps scope as a plain object:
     AssertionError: expected new DBRef("col", new ObjectId("6a1e7ecd8bb2260d2a730334"), "db") to not have property '_bsontype'
      at Context.<anonymous> (test/node/parser/deserializer.test.ts:114:38)
      at processImmediate (node:internal/timers:504:21)
```

So, basically, code-with-scope no loger promotes a dbref-shaped scope to DBRef instance, in the new code, the scope stays a plain object.
```
case constants.BSON_DATA_CODE_W_SCOPE:
    result = new Code(completedFrame.functionString!, completedFrame.holdingDocument);
    break;
```

It doesn't change the bytes, only the in-memory class type. Not sure if it was intended, the new behaviour seems more correct, but worth flagging (or restore the original and create follow-up ticket).

Either way - I think adding this or similar test is worth doing, wdyt?

src/parser/deserializer.ts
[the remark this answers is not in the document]
tadjik1: I don't have strong opinion here, we do have this check for `Object` path (and throw `bad embedded document length in bson` there), so maybe fixing the consistency here would be better. But you are right, the error type is going to change here.
I would recommend adding validation here, but either way works for me 👍

src/parser/serializer.ts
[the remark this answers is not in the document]
tadjik1: Agreed, let's keep behavior unchanged in this PR and file a follow-up. The unsound cast is pre-existing on main, and changing the error type touches user-visible behavior. I'll file the follow-up ticket.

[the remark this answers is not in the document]
tadjik1: Ah, wait, found ticket https://jira.mongodb.org/browse/NODE-3027, it was closed as "work as designed" in 2021. 
The team already decided non-string Map keys are unsupported by design. No follow-up ticket needed.

tadjik1:
@PavelSafronov can you please also release highlight block in the PR?

src/parser/deserializer.ts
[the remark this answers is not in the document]
PavelSafronov: I'm inclined to change the error to be more descriptive. Making that change.

src/parser/deserializer.ts
PavelSafronov: Tests added.

src/parser/deserializer.ts
PavelSafronov: To summarize: current (main) code promotes a plain object to a DBRef if the properties look like a DBRef. This behavior was never intended for Code scopes, but accidentally "snuck  in".
Adding in your suggested test and it is passing.
Sounds important enough to add to the release notes.

tadjik1:
Thanks, everything looks good to me! Can you please also check failing linter? I believe there is something with formatting:
```
[2026/06/02 20:18:52.112] /data/mci/3b75c5a5450a7c4ac5df2e6ec16d9807/src/test/node/parser/deserializer.test.ts
[2026/06/02 20:18:52.112]   273:54  error  Replace `BSON.BSONError,·'bad·embedded·array·length·in·bson'` with `⏎········BSON.BSONError,⏎········'bad·embedded·array·length·in·bson'⏎······`  prettier/prettier
[2026/06/02 20:18:52.112]   280:54  error  Replace `BSON.BSONError,·'bad·embedded·array·length·in·bson'` with `⏎········BSON.BSONError,⏎········'bad·embedded·array·length·in·bson'⏎······`  prettier/prettier
[2026/06/02 20:18:52.112]   287:54  error  Replace `BSON.BSONError,·'bad·embedded·array·length·in·bson'` with `⏎········BSON.BSONError,⏎········'bad·embedded·array·length·in·bson'⏎······`  prettier/prettier
[2026/06/02 20:18:52.112] ✖ 3 problems (3 errors, 0 warnings)
```

But code-wise I believe we are complete, great job 🚀

tadjik1 (approved):
great work @PavelSafronov, thanks a lot! 🥇 

I'm waiting for driver's perf tests to complete before merging

tadjik1:
@nbbeeken perf tests on driver haven't shown any regressions

nbbeeken (approved):

PR #887, opened 2026-05-25
feat(NODE-7069): deprecate Long members inapplicable to Timestamp

johnmtll:
#### Summary of Changes

Timestamp extends Long for storage convenience, but its semantics are strictly a (t, i) pair — not a general 64-bit integer. Inherited Long members that don't map cleanly to that model (arithmetic, bitwise ops, sign-related methods, raw conversions, and bit accessors) are now marked @deprecated, with declare-only redeclarations so no runtime behavior changes. The LongWithoutOverrides type is narrowed from an open denylist to an explicit allowlist (TimestampKept), preventing future Long additions from silently appearing on Timestamp's public surface. This lays the groundwork for a future refactor where Long becomes a private backing field rather than a base class.

##### Notes for Reviewers

Sets up a future refactor toward wrapping (Long as a backing field, as
suggested in https://jira.mongodb.org/browse/NODE-2624)
which the type surface already accommodates.

There's also no practical need to convert Timestamp to/from number, since the conversion is imprecise (lossful).

### Deprecated inapplicable `Long` members on `Timestamp`

`Timestamp` inherits from `Long` for storage convenience, but it is semantically a `(t, i)` pair, not a general 64-bit integer. If you have called `Long` methods directly on `Timestamp` instances (for example `toNumber()`, `getHighBits()`, `getLowBits()`, or aithmetic and bitwise operations), those are now marked `@deprecated` with no runtime behavior changes. The `.t` and `.i` accessors provide direct, semantically consistent access to the two timestamp components.

- [ ] Lint is passing (`npm run check:lint`)

### DIFF
--- a/src/timestamp.ts
+++ b/src/timestamp.ts
@@ -312,7 +312,7 @@ export class Timestamp extends LongWithoutOverridesClass {
   /** @deprecated Not applicable to Timestamp; use `.t` instead. */
   declare high: Long['high'];
 
-  /** @deprecated Not applicable to Timestamp. Use `_bsontype === 'Timestamp'` to identify a Timestamp. */
+  /** @deprecated Not applicable to Timestamp. Use `bsontype === 'Timestamp'` to identify a Timestamp. */
   declare __isLong__: Long['__isLong__'];
   /** @deprecated Not applicable to Timestamp. */
   declare getNumBitsAbs: Long['getNumBitsAbs'];
--- a/src/timestamp.ts
+++ b/src/timestamp.ts
@@ -27,6 +27,7 @@ export type TimestampOverrides =
 type TimestampKept =
   | 'toBigInt'
   | 'toString'
+  | 'compare'
   | 'equals'
   | 'eq'
   | 'notEquals'
@@ -310,8 +311,6 @@ export class Timestamp extends LongWithoutOverridesClass {
   /** @deprecated Not applicable to Timestamp; use `.t` instead. */
   declare high: Long['high'];
 
-  /** @deprecated Use `.equals()`, `.lessThan()`, or `.greaterThan()` for comparisons. */
-  declare compare: Long['compare'];
   /** @deprecated Use `.equals()`, `.lessThan()`, or `.greaterThan()` for comparisons. */
   declare comp: Long['comp'];
   /** @deprecated Compare `.t` and `.i` against `0` explicitly. */
--- a/src/timestamp.ts
+++ b/src/timestamp.ts
@@ -41,11 +41,7 @@ type TimestampKept =
   | 'gt'
   | 'greaterThanOrEqual'
   | 'gte'
-  | 'ge'
-  | 'compare'
-  | 'comp'
-  | 'isZero'
-  | 'eqz';
+  | 'ge';
 
 /** @public */
 export type LongWithoutOverrides = new (
@@ -310,6 +306,15 @@ export class Timestamp extends LongWithoutOverridesClass {
   declare low: Long['low'];
   /** @deprecated Not applicable to Timestamp; use `.t` instead. */
   declare high: Long['high'];
+
+  /** @deprecated Use `.equals()`, `.lessThan()`, or `.greaterThan()` for comparisons. */
+  declare compare: Long['compare'];
+  /** @deprecated Use `.equals()`, `.lessThan()`, or `.greaterThan()` for comparisons. */
+  declare comp: Long['comp'];
+  /** @deprecated Compare `.t` and `.i` against `0` explicitly. */
+  declare isZero: Long['isZero'];
+  /** @deprecated Compare `.t` and `.i` against `0` explicitly. */
+  declare eqz: Long['eqz'];
 
   /** @deprecated Not applicable to Timestamp. Use `bsontype === 'Timestamp'` to identify a Timestamp. */
   declare __isLong__: Long['__isLong__'];
--- a/src/timestamp.ts
+++ b/src/timestamp.ts
@@ -25,8 +25,6 @@ export type TimestampOverrides =
  * the public type — preventing accidental behavioral bleed-through.
  */
 type TimestampKept =
-  | 'high'
-  | 'low'
   | 'toString'
   | 'equals'
   | 'eq'
@@ -309,6 +307,11 @@ export class Timestamp extends LongWithoutOverridesClass {
   declare getLowBits: Long['getLowBits'];
   /** @deprecated Not applicable to Timestamp; use `.i` for the increment ordinal. */
   declare getLowBitsUnsigned: Long['getLowBitsUnsigned'];
+
+  /** @deprecated Not applicable to Timestamp; use `.i` instead. */
+  declare low: Long['low'];
+  /** @deprecated Not applicable to Timestamp; use `.t` instead. */
+  declare high: Long['high'];
 
   /** @deprecated Not applicable to Timestamp. Use `_bsontype === 'Timestamp'` to identify a Timestamp. */
   declare __isLong__: Long['__isLong__'];
--- a/src/timestamp.ts
+++ b/src/timestamp.ts
@@ -17,7 +17,7 @@ export type TimestampOverrides =
 
 /**
  * @public
- * 
+ *
  * Inherited `Long` members surfaced on `Timestamp`.
  * When `Long` gains a new member: add it here if it should pass through, or
  * add a `@deprecated declare` line on the `Timestamp` class if it should be
--- a/src/timestamp.ts
+++ b/src/timestamp.ts
@@ -16,7 +16,7 @@ export type TimestampOverrides =
   | typeof bsonType;
 
 /**
- * @internal
+ * @public
  * 
  * Inherited `Long` members surfaced on `Timestamp`.
  * When `Long` gains a new member: add it here if it should pass through, or
--- a/src/timestamp.ts
+++ b/src/timestamp.ts
@@ -77,6 +77,9 @@ export class Timestamp extends LongWithoutOverridesClass {
     return 'Timestamp';
   }
 
+  /**
+   * @deprecated Use `Long.MAX_UNSIGNED_VALUE` instead.
+   */
   static readonly MAX_VALUE = Long.MAX_UNSIGNED_VALUE;
 
   /**
--- a/src/timestamp.ts
+++ b/src/timestamp.ts
@@ -25,6 +25,7 @@ export type TimestampOverrides =
  * the public type — preventing accidental behavioral bleed-through.
  */
 type TimestampKept =
+  | 'toBigInt'
   | 'toString'
   | 'equals'
   | 'eq'
@@ -288,8 +289,6 @@ export class Timestamp extends LongWithoutOverridesClass {
   declare toInt: Long['toInt'];
   /** @deprecated Not applicable to Timestamp; use `.t` for seconds since the Unix epoch, or `.i` for the increment ordinal. */
   declare toNumber: Long['toNumber'];
-  /** @deprecated Not applicable to Timestamp; use `.t` for seconds since the Unix epoch, or `.i` for the increment ordinal. */
-  declare toBigInt: Long['toBigInt'];
   /** @deprecated Not applicable to Timestamp; use `.t` for seconds since the Unix epoch, or `.i` for the increment ordinal. */
   declare toBytes: Long['toBytes'];
   /** @deprecated Not applicable to Timestamp; use `.t` for seconds since the Unix epoch, or `.i` for the increment ordinal. */
--- a/src/timestamp.ts
+++ b/src/timestamp.ts
@@ -183,6 +183,7 @@ export class Timestamp extends LongWithoutOverridesClass {
 
   /**
    * Returns a Timestamp from the given string, optionally using the given radix.
+   * @deprecated Interprets `str` as a uint64, splitting it across (t, i); the result rarely matches user intent. Use `new Timestamp({ t, i })` or `Timestamp.fromBits(lowBits, highBits)` for explicit construction.
    *
    * @param str - the textual representation of the Timestamp.
    * @param optRadix - the radix in which the text is written.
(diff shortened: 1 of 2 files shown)


src/timestamp.ts
```diff
@@ -11,13 +11,45 @@ export type TimestampOverrides =
...
+  | 'eqz';
+
 /** @public */
 export type LongWithoutOverrides = new (
   low: unknown,
   high?: number | boolean,
   unsigned?: boolean
 ) => {
-  [P in Exclude<keyof Long, TimestampOverrides>]: Long[P];
+  [P in TimestampKept]: Long[P];
```
PavelSafronov, line 53: This makes `TimestampOverrides` obsolete. Let's add a comment and a deprecation notice there as well?

src/timestamp.ts
PavelSafronov, line 191: This method should be deprecated as well, since it also constructs a Timestamp from a uint64.

PavelSafronov (commented):
Looking good so far.

As discussed earlier, this PR needs the `test/types` additions.
And I left a few small comments.

src/timestamp.ts
```diff
@@ -4,20 +4,55 @@ import type { Int32 } from './int_32';
...
   | typeof bsonType;
+
+/**
+ * Inherited `Long` members surfaced on `Timestamp`.
+ * When `Long` gains a new member: add it here if it should pass through, or
+ * add a `@deprecated declare` line on the `Timestamp` class if it should be
+ * surfaced as deprecated. Anything left unclassified is silently dropped from
+ * the public type — preventing accidental behavioral bleed-through.
+ */
+type TimestampKept =
```
PavelSafronov, line 27: A new warning appears now, https://spruce.corp.mongodb.com/task/node_bson_linux_web_tests_patch_5499890a54a7aacc64c50bd28834bf355fdaf9c4_6a15f3cbd9fe5d00073aa2f9_26_05_26_19_26_05/html-log?execution=0&origin=task#L112

>  Warning: src/timestamp.ts:50:1 - (ae-forgotten-export) The symbol "TimestampKept" needs to be exported by the entry point index.d.ts

We can fix this by adding `@internal` to this type.

src/timestamp.ts
```diff
@@ -173,4 +215,105 @@ export class Timestamp extends LongWithoutOverridesClass {
...
+  // Signing
+  /** @deprecated Not applicable to Timestamp as the underlying Long is always unsigned. */
+  declare toSigned: Long['toSigned'];
+  /** @deprecated Not applicable to Timestamp as the underlying Long is always unsigned (this is a no-op). */
+  declare toUnsigned: Long['toUnsigned'];
+  /** @deprecated Not applicable to Timestamp as the underlying Long is always unsigned (always returns false). */
+  declare isNegative: Long['isNegative'];
+  /** @deprecated Not applicable to Timestamp as the underlying Long is always unsigned (returns true except for the zero Timestamp). */
+  declare isPositive: Long['isPositive'];
+  /** @deprecated Not applicable to Timestamp as the underlying Long is always unsigned (always returns true). */
```
PavelSafronov: Nit: `always returns true` -> `is always true`

PavelSafronov (dismissed):
Changes look good, two nits.

Please add release notes, this feels like a large enough thing to note, even one sentence like this:
> Deprecated ~40 inherited Long methods on Timestamp that have no meaningful Timestamp semantics; use .t / .i instead.

src/timestamp.ts
johnmtll: TimestampKept is used in LongWithoutOverrides, which is marked `@public`, so I've marked it as public.

src/timestamp.ts
```diff
@@ -4,20 +4,57 @@ import type { Int32 } from './int_32';
...
+ *
+ * Inherited `Long` members surfaced on `Timestamp`.
+ * When `Long` gains a new member: add it here if it should pass through, or
+ * add a `@deprecated declare` line on the `Timestamp` class if it should be
+ * surfaced as deprecated. Anything left unclassified is silently dropped from
+ * the public type — preventing accidental behavioral bleed-through.
+ */
+type TimestampKept =
+  | 'high'
+  | 'low'
```
nbbeeken: Shouldn't these be deprecated too? if we're trying to remove the subclassing of long I would think it should leave behind the long's fields since `i` and `t` would be more appropriate to use.

src/timestamp.ts
```diff
@@ -4,20 +4,57 @@ import type { Int32 } from './int_32';
...
+  | 'le'
+  | 'greaterThan'
+  | 'gt'
+  | 'greaterThanOrEqual'
+  | 'gte'
+  | 'ge'
+  | 'compare'
+  | 'comp'
+  | 'isZero'
+  | 'eqz';
```
nbbeeken: Total outsider here so if the team already discussed the methods to keep ignore me! 

I would've expected no carry over of these methods. Rather the timestamp class would need to go through a type conversion or access the underlying storage if these comparisons were necessary.

I feel like it would make sense to allow code like this to be written instead:
```js
new Timestamp({t: 0, i: 0}) < new Timestamp({t: 0, i: 1})
```

We would have to introduce the hook to convert to a bigint in the next major as a breaking change.

src/timestamp.ts
```diff
@@ -142,6 +185,7 @@ export class Timestamp extends LongWithoutOverridesClass {
 
   /**
    * Returns a Timestamp from the given string, optionally using the given radix.
+   * @deprecated Interprets `str` as a uint64, splitting it across (t, i); the result rarely matches user intent. Use `new Timestamp({ t, i })` or `Timestamp.fromBits(lowBits, highBits)` for explicit construction.
```
nbbeeken: Given that a TS is transmitted around as a 64bit value, I think the likelihood of being in possession of a single sequence of digits that is meant to be split into the two 32bit values that make up a timestamp is high and that it would match my intent to make a Timestamp out of it. I think it would make sense to keep such a method. 

---

Separate from that decision 

> Use `new Timestamp({ t, i })` or `Timestamp.fromBits(lowBits, highBits)` for explicit construction.

The suggested alternative is not really viable: I am a user in possession of a string of digits and neither of the suggested methods accept a string so the implication here is that I must figure out how to do something first with that string to call these methods which seems like I'm getting really into the weeds of BSON type formats to get right. 

> Use `new Timestamp(BigInt(str))`

Would be a suggestion that works with the value I have on hand

src/timestamp.ts
```diff
@@ -173,4 +217,105 @@ export class Timestamp extends LongWithoutOverridesClass {
...
+  /** @deprecated Incompatible with Timestamp: returns a signed integer, but Timestamp is always unsigned. Use `.t` for seconds since the Unix epoch. */
+  declare getHighBits: Long['getHighBits'];
+  /** @deprecated Not applicable to Timestamp; use `.t` for seconds since the Unix epoch. */
+  declare getHighBitsUnsigned: Long['getHighBitsUnsigned'];
+  /** @deprecated Incompatible with Timestamp: returns a signed integer, but Timestamp is always unsigned. Use `.i` for the increment ordinal. */
+  declare getLowBits: Long['getLowBits'];
+  /** @deprecated Not applicable to Timestamp; use `.i` for the increment ordinal. */
+  declare getLowBitsUnsigned: Long['getLowBitsUnsigned'];
+
+  /** @deprecated Not applicable to Timestamp. Use `_bsontype === 'Timestamp'` to identify a Timestamp. */
```
nbbeeken: we should recommend the `bsonType` symbol instead of the _ field

src/timestamp.ts
```diff
@@ -173,4 +217,105 @@ export class Timestamp extends LongWithoutOverridesClass {
...
+  /** @deprecated Not applicable to Timestamp as the underlying Long is always unsigned (is always true). */
+  declare unsigned: Long['unsigned'];
+
+  // Conversion
+  /** @deprecated Not applicable to Timestamp; use `.t` for seconds since the Unix epoch, or `.i` for the increment ordinal. */
+  declare toInt: Long['toInt'];
+  /** @deprecated Not applicable to Timestamp; use `.t` for seconds since the Unix epoch, or `.i` for the increment ordinal. */
+  declare toNumber: Long['toNumber'];
+  /** @deprecated Not applicable to Timestamp; use `.t` for seconds since the Unix epoch, or `.i` for the increment ordinal. */
+  declare toBigInt: Long['toBigInt'];
```
nbbeeken: Since TS is transferred around as a 64bit value converting to a bigint seems like a reasonable operation. 

I still think it would be fine to deprecate this if the intention is to choose a more common path towards bigint type coercion than this bespoke method. But given the burden to maintain is low 🤷🏻 I'd leave it be

nbbeeken (changes requested):

src/timestamp.ts
tadjik1, line 84: This one is also not needed (as per AC of the ticket)

src/timestamp.ts
```diff
@@ -311,6 +307,15 @@ export class Timestamp extends LongWithoutOverridesClass {
   /** @deprecated Not applicable to Timestamp; use `.t` instead. */
   declare high: Long['high'];
 
+  /** @deprecated Use `.equals()`, `.lessThan()`, or `.greaterThan()` for comparisons. */
+  declare compare: Long['compare'];
```
nbbeeken: if we're keeping the greaterThan/lessThan methods then wouldn't compare still be useful? it's building block for sorting and the two aforementioned methods are simple `=== 1` or `=== -1` checks on the result of compare. I know we wanted to reduce the surface area but if comparisons in general are helpful then I'd see value in this one too. 

Loosely held opinion here though! Don't mind either way

nbbeeken (dismissed):

nbbeeken (dismissed):

src/timestamp.ts
PavelSafronov: New warning popped up: https://spruce.corp.mongodb.com/task/node_bson_linux_node_tests_v20_patch_2ea8dbb2f390291e21ee5e99f0a07d31b487b182_6a1f20030740e700071564cd_26_06_02_18_25_09/html-log?execution=0&origin=task

> Warning: src/timestamp.ts:48:1 - (ae-forgotten-export) The symbol "TimestampKept" needs to be exported by the entry point index.d.ts

src/timestamp.ts
```diff
@@ -173,4 +215,115 @@ export class Timestamp extends LongWithoutOverridesClass {
...
+  declare high: Long['high'];
+
+  /** @deprecated Use `.equals()`, `.lessThan()`, or `.greaterThan()` for comparisons. */
+  declare comp: Long['comp'];
+  /** @deprecated Compare `.t` and `.i` against `0` explicitly. */
+  declare isZero: Long['isZero'];
+  /** @deprecated Compare `.t` and `.i` against `0` explicitly. */
+  declare eqz: Long['eqz'];
+
+  /** @deprecated Not applicable to Timestamp. Use `bsontype === 'Timestamp'` to identify a Timestamp. */
```
PavelSafronov: Nit:
> Use the bsonType symbol or _bsontype === 'Timestamp' to identify a Timestamp.

src/timestamp.ts
```diff
@@ -173,4 +215,115 @@ export class Timestamp extends LongWithoutOverridesClass {
...
+  declare getLowBits: Long['getLowBits'];
+  /** @deprecated Not applicable to Timestamp; use `.i` for the increment ordinal. */
+  declare getLowBitsUnsigned: Long['getLowBitsUnsigned'];
+
+  /** @deprecated Not applicable to Timestamp; use `.i` instead. */
+  declare low: Long['low'];
+  /** @deprecated Not applicable to Timestamp; use `.t` instead. */
+  declare high: Long['high'];
+
+  /** @deprecated Use `.equals()`, `.lessThan()`, or `.greaterThan()` for comparisons. */
```
PavelSafronov: Nit:
> "Use .compare() for general comparison, or .equals(), .lessThan(), .greaterThan() for specific cases."

src/timestamp.ts
```diff
@@ -173,4 +215,115 @@ export class Timestamp extends LongWithoutOverridesClass {
...
+  declare shru: Long['shru'];
+
+  // Signing
+  /** @deprecated Not applicable to Timestamp as the underlying Long is always unsigned. */
+  declare toSigned: Long['toSigned'];
+  /** @deprecated Not applicable to Timestamp as the underlying Long is always unsigned (this is a no-op). */
+  declare toUnsigned: Long['toUnsigned'];
+  /** @deprecated Not applicable to Timestamp as the underlying Long is always unsigned (is always false). */
+  declare isNegative: Long['isNegative'];
+  /** @deprecated Not applicable to Timestamp as the underlying Long is always unsigned (is always true except for the zero Timestamp). */
```
PavelSafronov: > except for the zero Timestamp

Don't think this is accurate. Tried the below locally and isPositive always came back as true.

```
const zeroLong = new Timestamp(0n);
console.log(`zeroLong: ${zeroLong}, zeroLong isPositive: ${zeroLong.isPositive()}`);
const zeroBigint = new Timestamp(BigInt(0));
console.log(`zeroBigint: ${zeroBigint}, zeroBigint isPositive: ${zeroBigint.isPositive()}`);
const zeroObj = new Timestamp({ t: 0, i: 0 });
console.log(`zeroObj: ${zeroObj}, zeroObj isPositive: ${zeroObj.isPositive()}`);
```

src/timestamp.ts
PavelSafronov, line 90: Now that `low` is deprecated, `i` and `t` both produce deprecation warnings: 
Image

This is only a maintainer issue, so not blocking. But if we want to avoid these warnings, we can can do something like this:

```
  get i(): number {
    return (this as unknown as Long).low >>> 0;
  }

  get t(): number {
    return (this as unknown as Long).high >>> 0;
  }
```

johnmtll: If the goal is just suppression of deprecation strikethrough, I think we should keep it 'as is' since it's an (admittedtly small-value) breadcrumb for maintainers when the actual refactor development begins.

tadjik1 (approved):
thanks @johnmtll , great work! 👍

PavelSafronov (approved):

PR #890, opened 2026-06-09
fix(NODE-7606): restore serializer performance by removing generator-based iteration

tadjik1:
#### Summary of Changes

#886 (NODE-5937) made `serializeInto` iterative and, along the way, started producing each frame's key/value pairs from a generator (`toKvPairs`). Generators aren't free: every `.next()` saves and restores execution state and hands back a fresh `{value, done}` object, and `Object.entries()` builds a `[key, value]` array for each field on top of that. On a flat document with 145 fields that's hundreds of throwaway allocations per `serialize`, and it cost about 54% on `flat_bson._serialize_bson`.

This PR drops the generator. `makeFrame()` calls `Object.keys()` once when the frame is created and the loop walks them by index. Maps keep an `entries()` iterator since they have no random access, and arrays index straight into `keyIndex`.

A few smaller changes in the same hot loops:

- `regexp.test(key)` replaces `key.match(regexp)`, and the `ignoreKeys` Set lookup only runs when `key[0] === '$'`.
- `value._bsontype` is read once per BSON-typed value instead of several times down the dispatch chain.
- In the deserializer, the per-field `isArray` recompute becomes a maintained `currentIsArray`; the `setValue` closure becomes a top-level `assignValue(dest, …)` with `currentDest` tracked next to `currentFrame` so the destination is never recomputed per field; and a dead `globalUTFValidation` frame field is removed.
- Both parsers now thread their frame stack as a linked list (a `prev` pointer per frame) instead of a separate array, which drops the array push/pop per nested frame (suggested by @addaleax in the #886 review).

`flat_bson._serialize_bson` normalized throughput dropped from ~216 to ~99 after #886 merged. The perf alerting context on the `perf` build variant caught it.

## Benchmarks

`rhel90-dbx-perf-large`, Node v22.11.0, 10 000 iterations, 1 000 warmup. `normalized_throughput`, higher is better.

| Test | Baseline (`2ea8dbb`) | Regression (`0231603`) | This PR | vs Baseline | vs Regression |
|------|--------------------:|----------------------:|--------:|:-----------:|:-------------:|
| `flat_bson._serialize_bson`   | 215.58 |  98.87 | 228.42 | **+6.0%**  | **+131%** |
| `deep_bson._serialize_bson`   |  76.61 |  56.94 |  83.63 | **+9.2%**  | **+47%**  |
| `full_bson._serialize_bson`   |  92.33 |  85.76 | 132.01 | **+43%**   | **+54%**  |
| `flat_bson._deserialize_bson` |  93.62 |  90.76 |  92.23 |    −1.5%   |    +1.6%    |
| `deep_bson._deserialize_bson` |  69.00 |  63.87 |  64.19 |    −7.0%   |    +0.5%    |
| `full_bson._deserialize_bson` |  82.04 |  77.12 |  80.12 |    −2.3%   |    +3.9%    |

Serialize is back above baseline on all three document shapes. Flat and full deserialize sit within noise of baseline.

https://spruce.corp.mongodb.com/task/node_bson_perf_run_spec_benchmarks_patch_c5d1dd89d933556a9c9880152156d15caa0f293b_6a282aa58fce9500076b6fbc_26_06_09_15_00_55/trend-charts?execution=0

<table>
    <tr>
      <td>image</td>
<td>image</td></tr>
<td>image</td>
<td>image</td></tr>
<td>image</td>
<td>image</td></tr>
<td>image</td>
<td>image</td></tr>
<td>image</td>
<td>image</td></tr>
<td>image</td>
<td>image</td></tr>
<td>image</td>
</table>

The only metric still under baseline is `deep_bson._deserialize_bson` (−7.0%), and that's structural rather than something this PR introduced. [deep_bson](https://github.com/mongodb/js-bson/blob/main/test/bench/documents/deep_bson.json) is 63 documents nested six deep, so the iterative deserializer allocates a frame object for each nested document where the old recursive code just leaned on the call stack. Those allocations are most of the gap. This PR nudges the number up (63.87 to 64.19) and makes deep serialize a lot faster, but it doesn't close it. We can probably claw the rest back with lighter or pooled frames; I left that for a follow-up, since what remains is small and inside the benchmark's run-to-run noise.

#### Driver benchmarks

I also ran the driver's spec benchmark suite end to end with `node-mongodb-native` pointed at this branch (the `performance-tests` variant), comparing against the bson `^7.2.0` it currently ships, which predates #886 and is the known-good recursive code. Nothing regressed on the BSON-sensitive benchmarks. Serialize came out at parity or faster. Deserialize was at parity apart from a ~6% dip on the 1M-tiny-document aggregate, which is the same per-document setup cost from #886 noted above and not new here. The large-document bulk-insert benchmark bounced between −22% and +5% across runs, with the slow slot moving between it and the single-document variant; a micro-benchmark and the benchmark's own ~8% mainline CV confirm that's noise from the 2.73 MB document, not the serializer.
[1 run](https://spruce.corp.mongodb.com/version/6a28567c5f93d700079108da/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
[2 run](https://spruce.corp.mongodb.com/version/6a2865aa5883a70007e16084/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

### DIFF
--- a/src/parser/serializer.ts
+++ b/src/parser/serializer.ts
@@ -1,5 +1,6 @@
 import { Binary, validateBinaryVector } from '../binary';
 import type { BSONSymbol, DBRef, Document, MaxKey } from '../bson';
+import { bsonType } from '../bson_value';
 import type { Code } from '../code';
 import * as constants from '../constants';
 import type { Decimal128 } from '../decimal128';
@@ -206,7 +207,7 @@ function serializeMinMax(buffer: Uint8Array, key: string, value: MinKey | MaxKey
   // Write the type of either min or max key
   if (value === null) {
     buffer[index++] = constants.BSON_DATA_NULL;
-  } else if (value._bsontype === 'MinKey') {
+  } else if (value[bsonType] === 'MinKey') {
     buffer[index++] = constants.BSON_DATA_MIN_KEY;
   } else {
     buffer[index++] = constants.BSON_DATA_MAX_KEY;
@@ -276,7 +277,7 @@ function serializeDecimal128(buffer: Uint8Array, key: string, value: Decimal128,
 function serializeLong(buffer: Uint8Array, key: string, value: Long, index: number) {
   // Write the type
   buffer[index++] =
-    value._bsontype === 'Long' ? constants.BSON_DATA_LONG : constants.BSON_DATA_TIMESTAMP;
+    value[bsonType] === 'Long' ? constants.BSON_DATA_LONG : constants.BSON_DATA_TIMESTAMP;
   // Number of written bytes
   const numberOfWrittenBytes = ByteUtils.encodeUTF8Into(buffer, key, index);
   // Encode the name
@@ -640,16 +641,16 @@ export function serializeInto(
       if (value[constants.BSON_VERSION_SYMBOL] !== constants.BSON_MAJOR_VERSION) {
         throw new BSONVersionError();
       }
-      const bsontype = value._bsontype;
-      if (bsontype === 'ObjectId') {
+      const tag = value[bsonType];
+      if (tag === 'ObjectId') {
         index = serializeObjectId(buffer, key, value, index);
-      } else if (bsontype === 'Decimal128') {
+      } else if (tag === 'Decimal128') {
         index = serializeDecimal128(buffer, key, value, index);
-      } else if (bsontype === 'Long' || bsontype === 'Timestamp') {
+      } else if (tag === 'Long' || tag === 'Timestamp') {
         index = serializeLong(buffer, key, value, index);
-      } else if (bsontype === 'Double') {
+      } else if (tag === 'Double') {
         index = serializeDouble(buffer, key, value, index);
-      } else if (bsontype === 'Code') {
+      } else if (tag === 'Code') {
         const codeValue = value as Code;
         if (codeValue.scope && typeof codeValue.scope === 'object') {
           buffer[index++] = constants.BSON_DATA_CODE_W_SCOPE;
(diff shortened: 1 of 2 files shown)


src/parser/serializer.ts
```diff
@@ -571,27 +633,23 @@ export function serializeInto(
...
-        });
+        currentFrame = makeFrame(value, nestedStartIndex, null, frame);
         index += 4;
       }
     } else if (type === 'object') {
       if (value[constants.BSON_VERSION_SYMBOL] !== constants.BSON_MAJOR_VERSION) {
         throw new BSONVersionError();
-      } else if (value._bsontype === 'ObjectId') {
+      }
+      const bsontype = value._bsontype;
```
addaleax:
```suggestion
      const bsontype = value[bsonType];
```

Might be a really good opportunity for this drive-by cleanup

addaleax (dismissed):

src/parser/serializer.ts
tadjik1: Thanks! Replaced with `const tag = ...` to avoid naming problems.

nbbeeken (approved):
great work! linked lists are under utilized nice to see it lending a hand :)))

src/parser/serializer.ts
nbbeeken:
```
else if ('_bsontype' in object && typeof object._bsontype === 'string') {
      throw new BSONError(`BSON types cannot be serialized as a document`);
```

We have this error checking code, I don't think we can/should change it but it's an interesting limitation that's now lifted 😅 we can support `{ "_bsontype" : "str" }` documents for those who like to have fun with their data schema....

addaleax: It's not even just necessarily a limitation – I think it would be a reasonable invariant to expect that `bson.serialize()` never throws if its argument is a return value of `bson.deserialize()`. I'm pretty sure you this means that for example in the driver, right now there are documents that you can read from the database but not copy/replace back into it.

addaleax (approved):

src/parser/serializer.ts
nbbeeken: Yes another driver could totally insert a document that looks like the example I gave and our driver couldn't round trip it. so, bug ticket? 

Are there issues with backwards compat 🤔 like if we lift the prevention of this is there the chance that it could be more likely to _accidentally_ insert a document of this shape when you meant to just have a BSON primitive... need to think on that.

tadjik1: Yeah, seems like a pre-existing issue - it was there before the PR. Worth a ticket though, I'll file one, thanks!
https://jira.mongodb.org/browse/NODE-7608

PavelSafronov (approved):
Changes look good, approving. Thanks!

PR #893, opened 2026-06-16
perf(NODE-6246): store ObjectId bytes as four packed integers

tadjik1:
#### Summary of Changes

This replaces `ObjectId`'s per-instance `Uint8Array` with four 24-bit integer fields (`i0`, `i1`, `i2`, `i3`) holding bytes 0-2, 3-5, 6-8, and 9-11. With the bytes stored inline in the object instead of in a separate typed array, retained memory drops by about 63% per `ObjectId`, and serialization doesn't regress. It also keeps deep equality working (`assert.deepStrictEqual`, plus lodash `isEqual` and `cloneDeep`), which is what blocked the buffer pool in #707 and the `#`-private fields in #830.

Deep equality is the constraint that stalled the previous approaches:
- #830 moved the bytes into `#` private fields. lodash `cloneDeep` can't see private fields, so clones came out broken, and it was reverted for v7.
- The buffer pool in #707 stores a per-instance offset into a shared buffer. Two equal ObjectIds get different offsets, so lodash `isEqual` reports them as unequal.

Packed integers avoid both. The four fields are ordinary enumerable properties (not `#`), so `cloneDeep` copies them, and their values are a pure function of the 12 bytes, so two equal ObjectIds always have identical fields and `isEqual` returns true. I verified both against the built bundle with the real lodash package: `isEqual` is `true` for equal ids and `false` for different ones, and `cloneDeep` produces a working ObjectId that serializes to the same bytes. `equals` compares the four fields for two same-build ObjectIds and falls back to the public hex string for an `ObjectId` from a different copy of `bson`.

##### 24-bit chunks

Each field holds at most `0xffffff`, which is inside V8's small-integer (Smi) range, so all four stay inline in the object. Wider fields (`uint32`, `double`, `BigInt`) go past the Smi range and get boxed onto the heap, which uses more memory rather than less.
References:
https://v8.dev/blog/fast-properties
https://github.com/v8/v8/blob/main/src/objects/smi.h

#### Memory

Measured on Node v24 with forced GC, counting retained heapUsed + arrayBuffers.

| Per ObjectId | main | this PR |
|---|---|---|
| `new ObjectId()` / `(hex)` | 152 B | 56 B (63% less) |
| deserialized `{ _id }` (per doc) | 208 B | 112 B (46% less) |

`find().toArray()` 1_000_000 docs (3 ObjectIds each)
|  | main | this PR |
|---|---|---|
| total retained | 588.5 MB | 313.6 MB (47% less) |

On main the `new ObjectId()` cost depends on allocation: generated and deserialized ids use a pooled `Buffer` (152 B), while constructing from a standalone `Uint8Array` is 244 B, because each id gets its own `ArrayBuffer`. This PR stores the bytes inline, so all of them are a flat 56 B.

#### Representations considered

Before settling on packed integers, I compared the storage options in a microbenchmark. The packed mini-class measures the same 56 B as the shipped bundle, so the benchmark matches the real implementation. All of these options keep lodash `isEqual` and `cloneDeep` working, so the difference comes down to memory and serialization speed. Packed 24-bit integers are the smallest and the fastest:

| representation | memory/instance | serializeInto |
|---|---|---|
| packed, four 24-bit ints (this PR) | 56 B | 3.1 ns |
| latin1 string | 64 B | 14.5 ns |
| three uint32 | 80 B | 3.4 ns |
| two 48-bit doubles | 72 B | 89.5 ns |
| one BigInt | 64 B | 200.3 ns |

#### Performance

Per-operation timings, main vs this PR (ns/op, lower is better):

| operation | main | this PR |
|---|---|---|
| `new ObjectId()` | 43 | 49 |
| `new ObjectId(hex)` | 89 | 83 |
| `new ObjectId(buffer)` | 38 | 38 |
| `getTimestamp()` | 36 | 36 |
| `equals` | 31 | 4 |
| `toHexString` | 46 | 6.6 |
| `.id` getter | 0.4 | 25 |
| serialize (doc with one `_id`) | 128 | 128 |
| deserialize (per id, array of 1000) | 50 | 20 |

`.id` is the only regression: main returns the stored buffer (no allocation), while this PR rebuilds the bytes into a fresh buffer on each call (~25 ns, ~40M ops/sec). The library no longer uses `.id` internally, so it only affects callers reading raw `.id` bytes in a loop.

The only driver consumer of `ObjectId#id` is [`compareObjectId`](https://github.com/mongodb/node-mongodb-native/blob/main/src/utils.ts#L1105), used by SDAM to order replica-set `electionIds`. It runs only while processing `hello` responses - initial discovery, monitoring heartbeats (default every 10s per server), and topology changes such as elections - never during server selection, command execution, or document (de)serialization. Its call rate is bounded by topology activity, not by application throughput, so the extra allocation from `.id` has no effect on steady-state read/write performance. The load-scaling paths (serialize, deserialize, equals) don't read `.id`.

##### Notes for Reviewers

Full driver CI run with this branch (including perf) - everything is green, driver perf is not affected either:
https://spruce.corp.mongodb.com/version/6a327cabd8bc53000702c17d

Behavior change (not semver-breaking): `ObjectId#id` now returns a freshly allocated buffer on each access instead of the live internal buffer, so `new ObjectId(buf).id === buf` is `false` again. The bytes still compare equal and `.id` is still a `Uint8Array`. That `=== buf` identity was only ever true because #498 introduced it in a patch release for performance; it was never part of the semver API, so this restores the prior behavior rather than breaking a contract.

Deserializer: the `ObjectId` path reads the 12 bytes straight from the wire buffer via an `@internal (buffer, offset) constructor`. That is safe because `ObjectId` decodes the bytes into its fields and keeps no reference to the source.

Tests: three existing tests asserted the old internal layout (one checked that `.id` returned the same buffer instance; two reached into the `buffer` field through `equals`). I rewrote them to assert the public contract instead: byte equality for `.id`, and value equality against an ObjectId from a different `bson` build.

### Smaller ObjectIds

`ObjectId` now stores its 12 bytes as four inline integers instead of a `Uint8Array`, cutting per-`ObjectId` memory by about 63% (and roughly 45 to 47% across large result sets) with no serialization slowdown. Deep equality and cloning are preserved (`assert.deepStrictEqual`, and lodash `isEqual` / `cloneDeep`).

### DIFF
--- a/src/objectid.ts
+++ b/src/objectid.ts
@@ -71,22 +71,22 @@ export class ObjectId extends BSONValue {
    * inline, which keeps the object small. The fields are enumerable own properties (not `#`
    * private) so two ObjectIds with the same bytes stay equal under structural comparison such
    * as deepStrictEqual.
+   * @internal
    */
-  // @internal
   private i0!: number;
-  // @internal
+  /** @internal */
   private i1!: number;
-  // @internal
+  /** @internal */
   private i2!: number;
-  // @internal
+  /** @internal */
   private i3!: number;
 
   /** Pack 12 bytes into the four integer fields. @internal */
-  private setFromBytes(b: Uint8Array): void {
-    this.i0 = (b[0] << 16) | (b[1] << 8) | b[2];
-    this.i1 = (b[3] << 16) | (b[4] << 8) | b[5];
-    this.i2 = (b[6] << 16) | (b[7] << 8) | b[8];
-    this.i3 = (b[9] << 16) | (b[10] << 8) | b[11];
+  private setFromBytes(b: Uint8Array, offset = 0): void {
+    this.i0 = (b[offset] << 16) | (b[offset + 1] << 8) | b[offset + 2];
+    this.i1 = (b[offset + 3] << 16) | (b[offset + 4] << 8) | b[offset + 5];
+    this.i2 = (b[offset + 6] << 16) | (b[offset + 7] << 8) | b[offset + 8];
+    this.i3 = (b[offset + 9] << 16) | (b[offset + 10] << 8) | b[offset + 11];
   }
 
   /**
@@ -159,12 +159,23 @@ export class ObjectId extends BSONValue {
    */
   constructor(inputId?: string | ObjectId | ObjectIdLike | Uint8Array);
   /**
+   * Read 12 bytes from `source` starting at `offset` into a new ObjectId, without allocating an
+   * intermediate view. Used by the deserializer on a hot path.
+   * @internal
+   */
+  constructor(source: Uint8Array, offset: number);
+  /**
    * Create a new ObjectId.
    *
    * @param inputId - An input value to create a new ObjectId from.
    */
-  constructor(inputId?: string | ObjectId | ObjectIdLike | Uint8Array) {
+  constructor(inputId?: string | ObjectId | ObjectIdLike | Uint8Array, offset?: number) {
     super();
+    if (typeof offset === 'number') {
+      // Fast path used by the deserializer: read the 12 bytes directly from source at offset.
+      this.setFromBytes(inputId as Uint8Array, offset);
+      return;
+    }
     // workingId is set based on type of input and whether valid id exists for the input
     let workingId;
     if (typeof inputId === 'object' && inputId && 'id' in inputId) {
--- a/src/objectid.ts
+++ b/src/objectid.ts
@@ -68,9 +68,8 @@ export class ObjectId extends BSONValue {
   /**
    * The 12 ObjectId bytes packed as four 24-bit integers (bytes 0-2, 3-5, 6-8, 9-11).
    * The 24-bit width keeps each value inside V8's small-integer (Smi) range so it stays stored
-   * inline, which keeps the object small. The fields are enumerable own properties (not `#`
-   * private) so two ObjectIds with the same bytes stay equal under structural comparison such
-   * as deepStrictEqual.
+   * inline, which keeps the object small. The fields are enumerable own properties, so two
+   * ObjectIds with the same bytes stay equal under structural comparison such as deepStrictEqual.
    * @internal
    */
   private i0!: number;
@@ -159,8 +158,8 @@ export class ObjectId extends BSONValue {
    */
   constructor(inputId?: string | ObjectId | ObjectIdLike | Uint8Array);
   /**
-   * Read 12 bytes from `source` starting at `offset` into a new ObjectId, without allocating an
-   * intermediate view. Used by the deserializer on a hot path.
+   * Read 12 bytes from `source` starting at `offset` directly into the packed fields.
+   * Used by the deserializer on a hot path.
    * @internal
    */
   constructor(source: Uint8Array, offset: number);
@@ -203,7 +202,11 @@ export class ObjectId extends BSONValue {
       this.i2 = (pu[2] << 16) | (pu[3] << 8) | pu[4];
       this.i3 = inc & 0xffffff;
     } else if (ArrayBuffer.isView(workingId) && workingId.byteLength === 12) {
-      this.setFromBytes(workingId);
+      // Normalize a non-Uint8Array view (DataView, Int8Array, ...) to a byte buffer so its
+      // bytes are read correctly.
+      this.setFromBytes(
+        workingId instanceof Uint8Array ? workingId : ByteUtils.toLocalBufferType(workingId)
+      );
     } else if (typeof workingId === 'string') {
       if (ObjectId.validateHexString(workingId)) {
         this.setFromHex(workingId);
@@ -244,9 +247,10 @@ export class ObjectId extends BSONValue {
   }
 
   set id(value: Uint8Array) {
-    this.setFromBytes(value);
+    const bytes = value instanceof Uint8Array ? value : ByteUtils.toLocalBufferType(value);
+    this.setFromBytes(bytes);
     if (ObjectId.cacheHexString) {
-      __idCache.set(this, ByteUtils.toHex(value));
+      __idCache.set(this, ByteUtils.toHex(bytes));
     }
   }
 
@@ -381,9 +385,15 @@ export class ObjectId extends BSONValue {
       return false;
     }
 
-    if (ObjectId.is(otherId) && typeof otherId.i3 === 'number') {
-      // Same-build ObjectId: compare the packed integer fields directly. i3 (the counter /
-      // low bytes) differs most often between two ids, so checking it first fails fast.
+    if (
+      ObjectId.is(otherId) &&
+      typeof otherId.i0 === 'number' &&
+      typeof otherId.i1 === 'number' &&
+      typeof otherId.i2 === 'number' &&
+      typeof otherId.i3 === 'number'
+    ) {
+      // Same-build ObjectId (all four packed fields present): compare them directly. i3 (the
+      // counter / low bytes) differs most often between two ids, so checking it first fails fast.
       return (
         this.i3 === otherId.i3 &&
         this.i0 === otherId.i0 &&
--- a/src/objectid.ts
+++ b/src/objectid.ts
@@ -178,6 +178,21 @@ export class ObjectId extends BSONValue {
     // workingId is set based on type of input and whether valid id exists for the input
     let workingId;
     if (typeof inputId === 'object' && inputId && 'id' in inputId) {
+      if (
+        ObjectId.is(inputId) &&
+        typeof inputId.i0 === 'number' &&
+        typeof inputId.i1 === 'number' &&
+        typeof inputId.i2 === 'number' &&
+        typeof inputId.i3 === 'number'
+      ) {
+        // Same-build ObjectId: copy the packed fields directly. The general path below would
+        // hit the id getter (which allocates) and then re-decode a hex round trip.
+        this.i0 = inputId.i0;
+        this.i1 = inputId.i1;
+        this.i2 = inputId.i2;
+        this.i3 = inputId.i3;
+        return;
+      }
       if (typeof inputId.id !== 'string' && !ArrayBuffer.isView(inputId.id)) {
         throw new BSONError('Argument passed in must have an id that is of type string or Buffer');
       }
--- a/src/objectid.ts
+++ b/src/objectid.ts
@@ -246,17 +246,17 @@ export class ObjectId extends BSONValue {
    */
   get id(): Uint8Array {
     const b = ByteUtils.allocateUnsafe(12);
-    b[0] = (this.i0 >> 16) & 0xff;
-    b[1] = (this.i0 >> 8) & 0xff;
+    b[0] = (this.i0 >>> 16) & 0xff;
+    b[1] = (this.i0 >>> 8) & 0xff;
     b[2] = this.i0 & 0xff;
-    b[3] = (this.i1 >> 16) & 0xff;
-    b[4] = (this.i1 >> 8) & 0xff;
+    b[3] = (this.i1 >>> 16) & 0xff;
+    b[4] = (this.i1 >>> 8) & 0xff;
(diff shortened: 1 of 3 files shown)


nbbeeken:
> Note: ObjectId.prototype.id now returns a new buffer on each access rather than the instance's internal buffer, so new ObjectId(buf).id === buf is no longer true (the bytes are still equal).

I am sure there's a test that probably fails if you change this but I'm going to guess that was to catch someone from accidentally introducing copying bytes and regressing the ctor performance. 

I would be surprised by considereding this part of the semver API and not a quirk of saving time on copying bytes because in another world OID would've made deep copies of input bytes to be certain of lifetime and ownership of the input. Just from an intention point-of-view I'm curious what identity equality provide if the contents can change? 🤔 

Incredible work, hope this all works as intended :shipit:

src/parser/deserializer.ts
```diff
@@ -399,9 +399,9 @@ function deserializeObject(
       value = ByteUtils.toUTF8(buffer, index, index + stringSize - 1, shouldValidateKey);
       index = index + stringSize;
     } else if (elementType === constants.BSON_DATA_OID) {
-      const oid = ByteUtils.allocateUnsafe(12);
-      for (let i = 0; i < 12; i++) oid[i] = buffer[index + i];
-      value = new ObjectId(oid);
+      // ObjectId copies these bytes into its own fields and keeps no reference to the source,
+      // so it is safe to hand it a view onto the wire buffer.
+      value = new ObjectId(buffer.subarray(index, index + 12));
```
nbbeeken: Scanning this PR with and overly discerning eye for perf 👁️ could we do even better here by making an `@internal` ctor that's invoked with `(buffer, offset)` it could read across the 12 bytes itself and then we don't create a new view that is just collected anyway. 

(may have no visible impact 🤷🏻)

src/objectid.ts
```diff
@@ -7,6 +7,16 @@ import { NumberUtils } from './utils/number_utils';
...
 
+/** byte (0-255): its 2-character lowercase hex pair @internal */
+const byteToHex: string[] = [];
+for (let n = 0; n < 256; n++) byteToHex.push(n.toString(16).padStart(2, '0'));
+
+/** hex char code: nibble (0-15); indices are the codes of 0-9, a-f, A-F @internal */
+const hexCharCodeToNibble = new Int8Array(103);
+for (let c = 48; c <= 57; c++) hexCharCodeToNibble[c] = c - 48; // '0'-'9'
+for (let c = 65; c <= 70; c++) hexCharCodeToNibble[c] = c - 55; // 'A'-'F'
+for (let c = 97; c <= 102; c++) hexCharCodeToNibble[c] = c - 87; // 'a'-'f'
```
nbbeeken, line 18: This kind of code incurs some degree of start up cost, think like `mongosh` is going to need to do this before a user can be given a `>` prompt simply because this is in the import tree.  

While hideous it may be worth serializing the output here into code rather than calculation. 

But it is just a thought, I think in the end all things should be measured before making a decision.

src/objectid.ts
```diff
@@ -55,8 +65,66 @@ export class ObjectId extends BSONValue {
...
-  /** ObjectId Bytes @internal */
-  private buffer!: Uint8Array;
+  /**
+   * The 12 ObjectId bytes packed as four 24-bit integers (bytes 0-2, 3-5, 6-8, 9-11).
+   * The 24-bit width keeps each value inside V8's small-integer (Smi) range so it stays stored
+   * inline, which keeps the object small. The fields are enumerable own properties (not `#`
+   * private) so two ObjectIds with the same bytes stay equal under structural comparison such
+   * as deepStrictEqual.
+   */
+  // @internal
```
nbbeeken:
```suggestion
  /** @internal */
```

I think it needs to be tsdoc format for api-extractor to pick this up

src/parser/deserializer.ts
tadjik1: This change almost halved the operation time in my local microbenchmarks!

src/objectid.ts
tadjik1: Thanks! I got confused a bit, checked now with generated `.d.ts` - they are not exposed anymore (Excluded from this release type).

src/objectid.ts
tadjik1: I tested locally - it seems to have an impact of <100µs, I'd say negligible compared to library load.

I'd rather not serialize it into a literal table, though. A hardcoded 256-entry array of hex pairs is a magic blob - you can't tell it's correct by reading it, whereas the loop is effectively the spec so it's obviously right. If the import cost is worth removing, I'd make it lazy (build on the first `toHexString`) rather than inline the values.

tadjik1:
> I am sure there's a test that probably fails if you change this but I'm going to guess that was to catch someone from accidentally introducing copying bytes and regressing the ctor performance.
> 
> I would be surprised by considereding this part of the semver API and not a quirk of saving time on copying bytes because in another world OID would've made deep copies of input bytes to be certain of lifetime and ownership of the input. Just from an intention point-of-view I'm curious what identity equality provide if the contents can change? 🤔

Yeah, there were 3 tests I had to change for the representation switch - this one (`.id === buf`), plus two equals tests that reached into the old internal `buffer` field. I agree it didn't buy us anything and was arguably a footgun, because `.id` returned the live buffer, mutating it would change the actual `ObjectId`, which probably isn't what you'd want from an "identity."

I'm still not sure whether it counts as breaking, since `new ObjectId(buf).id === buf` goes from `true` to `false` with this PR.

src/objectid.ts
```diff
@@ -114,15 +182,20 @@ export class ObjectId extends BSONValue {
...
+      const pu = ObjectId.PROCESS_UNIQUE!;
+      this.i0 = (time >>> 8) & 0xffffff;
+      this.i1 = ((time & 0xff) << 16) | (pu[0] << 8) | pu[1];
+      this.i2 = (pu[2] << 16) | (pu[3] << 8) | pu[4];
+      this.i3 = inc & 0xffffff;
     } else if (ArrayBuffer.isView(workingId) && workingId.byteLength === 12) {
-      // If instanceof matches we can escape calling ensure buffer in Node.js environments
-      this.buffer = ByteUtils.toLocalBufferType(workingId);
+      this.setFromBytes(workingId);
     } else if (typeof workingId === 'string') {
```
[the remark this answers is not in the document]
tadjik1: addressed

src/objectid.ts
```diff
@@ -138,15 +211,29 @@ export class ObjectId extends BSONValue {
...
+    b[11] = this.i3 & 0xff;
+    return b;
   }
 
   set id(value: Uint8Array) {
-    this.buffer = value;
+    this.setFromBytes(value);
     if (ObjectId.cacheHexString) {
       __idCache.set(this, ByteUtils.toHex(value));
     }
```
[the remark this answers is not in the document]
tadjik1: addressed

src/objectid.ts
```diff
@@ -267,9 +370,14 @@ export class ObjectId extends BSONValue {
       return false;
     }
 
-    if (ObjectId.is(otherId)) {
+    if (ObjectId.is(otherId) && typeof otherId.i3 === 'number') {
```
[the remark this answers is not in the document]
tadjik1: addressed

addaleax (dismissed):

src/objectid.ts
```diff
@@ -114,15 +192,24 @@ export class ObjectId extends BSONValue {
...
+      this.setFromBytes(
+        workingId instanceof Uint8Array ? workingId : ByteUtils.toLocalBufferType(workingId)
+      );
     } else if (typeof workingId === 'string') {
       if (ObjectId.validateHexString(workingId)) {
-        this.buffer = ByteUtils.fromHex(workingId);
+        this.setFromHex(workingId);
         // If we are caching the hex string
         if (ObjectId.cacheHexString) {
           __idCache.set(this, workingId);
```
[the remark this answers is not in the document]
tadjik1: This is real, but it predates this PR, I've created separate ticket for this work: https://jira.mongodb.org/browse/NODE-7625

src/objectid.ts
nbbeeken: awesome, sounds good to me to leave as-is then thanks for checking!

src/parser/deserializer.ts
nbbeeken: wooohooo!!

nbbeeken:
Well looking back to when the `should return the same instance if a buffer is passed in` test was written (https://github.com/mongodb/js-bson/pull/498) (long ago by an engineer long forgotten) in a patch release we changed `new ObjectId(buf).id === buf` from `false` to `true` for the sake of performance, so we didn't consider that to be an expectation of the API then, in my humble onion, I don't think we need to now.

tadjik1:
Thanks @nbbeeken! Let's discuss this next week and then I'll open PR for review. I agree, this doesn't seem like a breaking change.

nbbeeken (approved):

PavelSafronov (dismissed):
Awesome approach! Approving and sending for team review.

nbbeeken (dismissed):

addaleax (dismissed):

src/objectid.ts
```diff
@@ -301,18 +448,18 @@ export class ObjectId extends BSONValue {
...
-    uint8array[index + 3] = this.buffer[3];
-    uint8array[index + 4] = this.buffer[4];
-    uint8array[index + 5] = this.buffer[5];
-    uint8array[index + 6] = this.buffer[6];
-    uint8array[index + 7] = this.buffer[7];
-    uint8array[index + 8] = this.buffer[8];
-    uint8array[index + 9] = this.buffer[9];
-    uint8array[index + 10] = this.buffer[10];
-    uint8array[index + 11] = this.buffer[11];
+    uint8array[index] = (this.i0 >> 16) & 0xff;
```
seanrmilligan: Just for my edification, what is this part handling? Sign extension during the bit shift?

seanrmilligan: I ask because it looks like you have an unsigned right shift above that can maybe be used here as well?

```
const time = this.i0 * 0x100 + (this.i1 >>> 16);
```

nbbeeken: Should never be any sign to extend in any case because for all three numbers there is only the lower 3 bytes being used. It would probably make sense to change the line you reference to just be `>>` to avoid the implication of sign mattering, but no functional change either way. Good call out

tadjik1: Yep, good catch. Updated code for consistency (lowest 3 bytes are unsigned anyway but code-wise it still looks cleaner). The main idea is to store each 3 bytes of 12-bytes ObjectId in lowest bytes of 4 4-bytes integers, that's why top byte of every integer is always zero.

nbbeeken (approved):

src/objectid.ts
```diff
@@ -289,8 +435,9 @@ export class ObjectId extends BSONValue {
   /** Returns the generation date (accurate up to the second) that this ID was generated. */
   getTimestamp(): Date {
     const timestamp = new Date();
-    const time = NumberUtils.getUint32BE(this.buffer, 0);
-    timestamp.setTime(Math.floor(time) * 1000);
+    // Bytes 0-3 (the big-endian timestamp) are i0's three bytes followed by i1's top byte.
+    const time = this.i0 * 0x100 + (this.i1 >>> 16);
```
seanrmilligan, line 439: This took me down a horrible path of exploring why you used `* 0x100` instead of `<< 8`. So thank you for that :)

YIL (yesterday I learned) that there is an implicit cast to int32 and so you can't just treat `number` as a dumb holder of 64 bits. I expected all the bits would just cha-cha one up or down but if the number holds a floating point value,the implicit to int32 can cast to 0 (as it did in my local testing, but really any int), shifted, and then returned as... 0.

https://tc39.es/ecma262/#sec-numeric-types-number-unsignedRightShift

seanrmilligan (approved):

PavelSafronov (approved):

PR #901, opened 2026-06-23
fix(NODE-7631): count Int32 and BSONSymbol in calculateObjectSize

spokodev:
`calculateObjectSize()` documents (and tests, for `bigint`/`Symbol`) the invariant that it returns the exact byte length `serialize()` will produce. It breaks for two BSON types: it **over-reports by 12 bytes per occurrence** (recursively) for documents containing `Int32` or `BSONSymbol` values.

```js
calculateObjectSize({ a: new Int32(2) })          // 24, serialize().byteLength = 12
calculateObjectSize({ a: new BSONSymbol('sym') })  // 28, serialize().byteLength = 16
calculateObjectSize({ a: [new Int32(1), new Int32(2)] }) // 51 vs 27
```

### Cause (`src/parser/calculate_size.ts`, `calculateElementSize`)
- **`Int32` has no branch.** It falls through to the final `else`, which treats the wrapper as a sub-document and walks its own enumerable `value` property (`Object.keys(new Int32(2))` → `['value']`), adding 5 bytes of fake-document overhead plus a phantom 7-byte `value` field = 12 extra bytes.
- **`BSONSymbol` guard is stale.** The branch tests `value._bsontype === 'Symbol'`, but the tag is `'BSONSymbol'` (`src/symbol.ts`, and the serializer's own `tag === 'BSONSymbol'`). It never matches, so `BSONSymbol` falls through the same way.

### Impact
`calculateObjectSize` is the documented way to pre-check the 16 MB `maxBsonObjectSize` limit and to size write batches before sending to the server. Over-counting causes documents/batches that are actually within the limit to be rejected client-side as too large, and over-allocation in any consumer that pre-sizes a buffer from this value. `Int32` is extremely common (any explicit 32-bit integer).

### Fix
Add an `Int32` branch returning the int32 element size (`4 + 1`, identical to the existing `number`/int32 path) and correct the `BSONSymbol` type guard. No other type is affected.

### Tests
Added cases asserting `calculateObjectSize(doc) === serialize(doc).byteLength` for `Int32` and `BSONSymbol`, mirroring the existing `bigint`/`Symbol` tests. They fail on the current code (24≠12, 28≠16) and pass with the fix. Full `test/node` suite green (4523 passing).

### `calculateObjectSize()` now returns accurate byte counts for `Int32` and `BSONSymbol`

Previously, each `Int32` or `BSONSymbol` value caused a 12-byte overcount. Thanks to @spokodev for the fix!

### DIFF
--- a/test/node/parser/calculate_size.test.ts
+++ b/test/node/parser/calculate_size.test.ts
@@ -57,9 +57,24 @@ describe('calculateSize()', () => {
     });
   });
 
-  describe('when given a BSONSymbol value with a single character key', function () {
+  describe('when given an array of Int32 values', function () {
     it('matches the serialized byte length', function () {
+      const doc = { a: [new BSON.Int32(1), new BSON.Int32(2)] };
+      expect(BSON.calculateObjectSize(doc)).to.equal(BSON.serialize(doc).byteLength);
+    });
+  });
+
+  describe('when given an Int32 value nested in a subdocument', function () {
+    it('matches the serialized byte length', function () {
+      const doc = { a: { b: new BSON.Int32(1) } };
+      expect(BSON.calculateObjectSize(doc)).to.equal(BSON.serialize(doc).byteLength);
+    });
+  });
+
+  describe('when given a BSONSymbol value with a single character key', function () {
+    it('returns the exact byte count (+4 bytes for document size + 1 type byte + 1 byte for "a" + 4 bytes string length + 4 bytes "sym\\0" + 2 null terminators)', function () {
       const doc = { a: new BSON.BSONSymbol('sym') };
+      expect(BSON.calculateObjectSize(doc)).to.equal(4 + 1 + 1 + 4 + 4 + 1 + 1);
       expect(BSON.calculateObjectSize(doc)).to.equal(BSON.serialize(doc).byteLength);
     });
   });
--- a/test/node/parser/calculate_size.test.ts
+++ b/test/node/parser/calculate_size.test.ts
@@ -48,4 +48,19 @@ describe('calculateSize()', () => {
       expect(BSON.calculateObjectSize(doc)).to.equal(BSON.serialize(doc).byteLength);
     });
   });
+
+  describe('when given an Int32 value with a single character key', function () {
+    it('returns 12 bytes (+4 bytes for document size + 1 type byte + 1 byte for "a" + 4 bytes int32 + 2 null terminators)', function () {
+      const doc = { a: new BSON.Int32(2) };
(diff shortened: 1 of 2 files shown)


test/node/parser/calculate_size.test.ts
```diff
@@ -48,4 +48,19 @@ describe('calculateSize()', () => {
       expect(BSON.calculateObjectSize(doc)).to.equal(BSON.serialize(doc).byteLength);
     });
   });
+
+  describe('when given an Int32 value with a single character key', function () {
```
tadjik1, line 52: Can you please also test case with arrays which was mentioned in the PR description, something like `{ a: [new Int32(1), new Int32(2)] }`?

test/node/parser/calculate_size.test.ts
```diff
@@ -48,4 +48,19 @@ describe('calculateSize()', () => {
...
+    it('returns 12 bytes (+4 bytes for document size + 1 type byte + 1 byte for "a" + 4 bytes int32 + 2 null terminators)', function () {
+      const doc = { a: new BSON.Int32(2) };
+      expect(BSON.calculateObjectSize(doc)).to.equal(4 + 1 + 1 + 4 + 1 + 1);
+      expect(BSON.calculateObjectSize(doc)).to.equal(BSON.serialize(doc).byteLength);
+    });
+  });
+
+  describe('when given a BSONSymbol value with a single character key', function () {
+    it('matches the serialized byte length', function () {
```
tadjik1, line 68: Can you please add similar test to the one from Int32, i.e. counting the exact number of bytes for the document with `BSONSymbol`? 
```
expect(BSON.calculateObjectSize(doc)).to.equal(4 + 1 + 1 + 4 + 1 + 1);
```

test/node/parser/calculate_size.test.ts
tadjik1, line 52: Can you please also add test for nested document: `{ a: { b: new Int32(1) } }` to make sure the right size is counted for subdocuments?

tadjik1 (changes requested):
Thank you @spokodev, great catch and great work! Please take a look at my comments and let me know what you think.

spokodev:
Thanks for the review. I pushed the three cases in `c1ccc4b`: an `Int32` array (`{ a: [Int32(1), Int32(2)] }`), a nested subdocument (`{ a: { b: Int32(1) } }`), and a `BSONSymbol` case asserting the exact byte count and cross-checking it against `serialize().byteLength`.

tadjik1 (approved):

PavelSafronov (approved):
Looks good, thanks for the PR @spokodev !

tadjik1:
hi @spokodev , thanks again! The pr is merged, and we going to to include it in the next minor release 7.4.0.

PR #903, opened 2026-06-29
fix(NODE-7648): guard startup snapshot detection

shmuelhizmi:
### Description
See bun issue at https://github.com/oven-sh/bun/issues/32501
will be fixed in next version of bun

#### Summary of Changes

Bun exposes `process.getBuiltinModule("v8").startupSnapshot.isBuildingSnapshot`, but calling it throws `NotImplementedError`. BSON calls this during `ObjectId` static initialization, so importing BSON in Bun fails before any API is used.

Tested with Bun 1.3.12.

Before this change:

```sh
npm i bson@7.3.1
bun -e "require('bson')"
```

```txt
NotImplementedError: node:v8 isBuildingSnapshot is not yet implemented in Bun.
```

After this change:

```sh
npm run build
bun -e "require('./lib/bson.cjs'); console.log('bson loaded')"
```

```txt
bson loaded
```

This moves the startup snapshot helpers into `src/utils/node_utils.ts`, skips the check when `globalThis.Bun` is present, and keeps a `try/catch` around the Node startup snapshot call for other partial implementations.

- [x] Lint is passing (`ESLINT_USE_FLAT_CONFIG=false npx eslint --ext '.js,.ts' --max-warnings=0 src/objectid.ts src/utils/node_utils.ts test/node/startup_snapshot.test.ts`)
- [x] PR title follows the conventional commit format: `type: description`
- [x] No new TODOs

Additional validation: `npm run build`, `npm exec -- mocha test/node/startup_snapshot.test.ts`, and `bun -e "require('./lib/bson.cjs')"`.


src/utils/node_utils.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export function getNodeStartupSnapshot(): NodeStartupSnapshot | undefined {
+  // @ts-expect-error Node.js types not present since this is an optional API
+  return globalThis?.process?.getBuiltinModule?.('v8')?.startupSnapshot;
+}
+
+export function isNodeProcessCurrentlyBuildingSnapshot(): boolean {
+  if ('Bun' in globalThis) {
+    return false;
+  }
```
addaleax: I'd rather not add a check like this since it should be possible for other runtimes (like bun) to implement support for startup snapshots

src/objectid.ts
```diff
@@ -46,9 +47,8 @@ export class ObjectId extends BSONValue {
   static {
     this.resetState();
     // https://nodejs.org/api/v8.html#startup-snapshot-api
-    // @ts-expect-error Node.js types not present since this is an optional API
-    const { startupSnapshot } = globalThis?.process?.getBuiltinModule?.('v8') ?? {};
-    if (startupSnapshot?.isBuildingSnapshot?.()) {
+    const startupSnapshot = getNodeStartupSnapshot();
+    if (isNodeProcessCurrentlyBuildingSnapshot()) {
       startupSnapshot?.addDeserializeCallback?.(this.resetState);
     }
```
addaleax: It seems like the simplest fix here would be to wrap these lines in a try/catch?

If we do want to externalize some logic into a separate file, I'd definitely suggest to export a single `addDeserializeCallback()` function and have that handle the conditional logic here

addaleax (commented):
> Bun exposes `process.getBuiltinModule("v8").startupSnapshot.isBuildingSnapshot`, but calling it throws `NotImplementedError`. 

This seems like an issue worth fixing in Bun as well – if startup snapshot support is not implemented, it is safe and correct to return `false` from `isBuildingSnapshot()`.

shmuelhizmi:
i made the refactor and linked bun issue https://github.com/oven-sh/bun/issues/32501 (merged)
tnx for the help @addaleax

src/utils/v8_snapshot_callback.ts
```diff
@@ -0,0 +1,23 @@
...
+declare global {
+  var process:
+    | { getBuiltinModule: (module: 'v8') => { startupSnapshot: NodeStartupSnapshot } }
+    | undefined;
+}
+
+export function addV8SnapshotDeserializeCallback(callback: () => void) {
+  try {
+    const startUpSnapshot = globalThis?.process?.getBuiltinModule?.('v8')?.startupSnapshot;
+    if (startUpSnapshot?.isBuildingSnapshot?.() ?? false) {
```
addaleax:
```suggestion
    if (startUpSnapshot?.isBuildingSnapshot?.()) {
```

(`if` already coerces to boolean)

addaleax (dismissed):
Looks good as far a I am concerned!

PavelSafronov:
Thank you for taking the time to create such a detailed PR. This PR will be tracked in NODE-7648 and the team will prioritize it in the next triage session.

johnmtll:
Thanks again for the contribution! We’re closing this PR because the change appears to address a runtime-specific edge case rather than a broader BSON bug, and we don’t want to add a new defensive pattern for optional runtime APIs without evidence that this is needed across other runtimes.

PR #909, opened 2026-07-02
fix(NODE-7645): use the first instant of year 10000 as the relaxed EJSON date bound

spokodev:
## Description

In relaxed Extended JSON mode, dates in the first five hours of year 10000 are emitted as ISO strings with a six-digit year, which strict Extended JSON consumers (mongosh, the server) reject.

The relaxed-range guard in `src/extended_json.ts` uses `253402318800000`, which is `+010000-01-01T05:00:00Z`, five hours *into* year 10000. So a `Date` in `[253402300800000, 253402318799999]` passes the "year 1970-9999" check and is serialized as a relaxed ISO string:

```js
EJSON.stringify({ a: new Date(253402300800000) }, { relaxed: true })
// {"a":{"$date":"+010000-01-01T00:00:00Z"}}   invalid 6-digit-year ISO
// should be {"a":{"$date":{"$numberLong":"253402300800000"}}}
```

The correct upper bound is the first instant of year 10000, `Date.UTC(10000, 0, 1) === 253402300800000`, which is exactly the value the repo's own `bson-corpus` `Y10K` fixture uses (`{"$date":{"$numberLong":"253402300800000"}}`).

## Fix

```diff
-      inRange = dateNum > -1 && dateNum < 253402318800000;
+      inRange = dateNum > -1 && dateNum < 253402300800000;
```

Dates from the first five hours of year 10000 now fall back to `$numberLong` in relaxed mode, matching canonical output and the Y10K corpus fixture.

## Test

Added a case to `test/node/extended_json.test.ts` asserting that `new Date(253402300800000)` relaxed-stringifies to `$numberLong`. It fails before the change (`+010000-01-01T00:00:00Z`) and passes after. The full node suite stays green.


PavelSafronov:
Thank you for taking the time to open this PR and provide detailed explanations. This PR will be tracked in NODE-7645 and the team will prioritize it in the next triage session.

src/extended_json.ts
```diff
@@ -264,7 +264,7 @@ function serializeValue(value: any, options: EJSONSerializeInternalOptions): any
   if (value instanceof Date || isDate(value)) {
     const dateNum = value.getTime(),
       // is it in year range 1970-9999?
-      inRange = dateNum > -1 && dateNum < 253402318800000;
+      inRange = dateNum > -1 && dateNum < 253402300800000;
```
PavelSafronov, line 270: A clear comment explaining where this value comes from would be helpful:
```suggestion
      inRange = dateNum > -1 && dateNum < 253402300800000; // Date.UTC(10_000)
```

PavelSafronov (changes requested):
Changes look good, only request is to add a comment to the new value of the magic integer.

spokodev:
Continued in #917. GitHub auto-closed this PR and won't let me reopen it because I force-pushed the branch while adding the comment you requested (my mistake; a shallow clone orphaned the branch history). #917 is the identical change plus that comment on the `253402300800000` bound. Thanks for the review.

PR #912, opened 2026-07-13
docs(NODE-7668): Add security warning to EJSON.parse documentation

mcmorisi:
## Summary
- Adds a warning to the `EJSON.parse` README section noting that using the parsed result in a query, update, or command can allow an attacker to inject operators or unexpected field values, especially when the input JSON comes from an untrusted source.

## Test plan
- Reviewed rendering of the GitHub alert syntax (`> [!WARNING]`) locally.


README.md
```diff
@@ -146,6 +146,16 @@ Only the following version combinations with the [MongoDB Node.js Driver](https:
 Parse an Extended JSON string, constructing the JavaScript value or object described by that
 string.
 
+> [!WARNING]
+> If you use the result of `EJSON.parse()` in a query, update, or command, an attacker
+> can inject operators or unexpected field values that change the meaning of the
+> operation. This risk is greatest when the input text originates from a user, an API
+> request, or another untrusted source. Validate untrusted input before parsing it.
```
PavelSafronov: > This risk is greatest when the input text originates from a user, an API request, or another untrusted source.

Users and APIs should be trusted at appropriate levels, I don't think we should be mentioning API in this specific warning.

mcmorisi: Updated!

PavelSafronov (approved):
Looks good from a tech POV, thanks!

johnmtll (approved):
LGTM! Thanks!

PR #915, opened 2026-07-23
docs(NODE-7682): add agents file

johnmtll:
#### Summary of Changes

Adds an AGENTS.md file to the project root, giving key context to ai coding agents on session creation. Conforms with agentskills.io. Features a sync script to support nested AGENTS.md files. Scaffolds the directory for shared skills.

##### Notes for Reviewers

scripts/symlink-claude-md.sh is vendored from an internal repo. Suggested changes upstream should begin there.

- [ ] Lint is passing (`npm run check:lint`)


README.md
```diff
@@ -95,6 +95,18 @@ If you are working directly in the browser without a bundler please use the `.mj
...
 
+## AI Agent Configuration
+
+This repository uses [agentskills.io](https://agentskills.io) conventions for AI coding agent
+instructions. `AGENTS.md` is the canonical source of truth — tool-specific files like `CLAUDE.md`
+are generated references.
+
+### Adding a nested AGENTS.md
+
+1. Create an `AGENTS.md` in the target directory.
```
PavelSafronov, line 106: If the user is creating an agents.md file, they should `git add` it first, otherwise the script won't see it.

scripts/symlink-claude-md.sh
PavelSafronov, line 1: Right now nothing is running this file, so there is not requirement to run this. Suggest adding a build step that fails if the running the script produces a difference.

scripts/symlink-claude-md.sh
```diff
@@ -0,0 +1,105 @@
...
+  else
+    printf '%s/CLAUDE.md\n' "$dir"
+  fi
+}
+
+remove_generated_file() {
+  local generated_file="$1"
+
+  if git ls-files --error-unmatch -- "$generated_file" > /dev/null 2>&1 || [ -e "$generated_file" ]; then
+    rm -f "$generated_file"
```
PavelSafronov, line 28: Generated file is removed without us checking its contents. If a user modifies this file, we'll be dropping all of their changes. Should we check that the file has expected contents before deleting it? This is something that shouldn't happen in CIs, but can happen on individual dev systems.

AGENTS.md
```diff
@@ -0,0 +1,79 @@
...
+- **Formatting** — Prettier: single quotes, 2-space indent, 100-char width, no trailing commas.
+
+## Other Important Information
+
+- Changes must be compliant with bson spec, and should conform to specification tests for bson (e.g: [bson-corpus](https://github.com/mongodb/specifications/blob/master/source/bson-corpus/bson-corpus.md)).
+- Before making key changes to anything within `src/parser/*`, consider performance implications and run benchmark test passes to ensure no significant performance degradation.
+
+## Commit Messages
+
+[Conventional Commits](https://www.conventionalcommits.org/) with a Jira ticket: `<type>(NODE-XXXX): <subject>` — types: `feat`, `fix`, `docs`, `style`, `refactor`, `perf`, `test`, `chore`; breaking changes use `!` (e.g. `feat(NODE-XXXX)!: …`).
```
PavelSafronov: This isn't strictly accurate: we use PR title as the commit title and squash on merge, so the PR commits don't have to follow the same format, they can be freeform.
Also, we sometimes have `chore: skip latest server tests` commits which don't have a tracking JIRA ticket. Not sure if we should call that out as an allowed practice or not.

PavelSafronov (changes requested):

scripts/symlink-claude-md.sh
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/env bash
+# 1. Ensure every staged AGENTS.md has a companion CLAUDE.md containing
+#    "@AGENTS.md" (a Claude Code import reference).
+# 2. Ensure .claude/skills is a symlink to .agents/skills.
+
+set -euo pipefail
```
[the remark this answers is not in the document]
johnmtll: this script is vendored from an internal repo. Suggested changes upstream should begin there.

scripts/symlink-claude-md.sh
```diff
@@ -0,0 +1,105 @@
...
+}
+
+remove_generated_file() {
+  local generated_file="$1"
+
+  if git ls-files --error-unmatch -- "$generated_file" > /dev/null 2>&1 || [ -e "$generated_file" ]; then
+    rm -f "$generated_file"
+    git add -u -- "$generated_file"
+    echo "auto-removed: $generated_file"
+  fi
```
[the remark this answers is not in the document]
johnmtll: this script is vendored from an internal repo. Suggested changes upstream should begin there.

scripts/symlink-claude-md.sh
PavelSafronov: This file is being copied from https://github.com/10gen/mongohouse/blob/master/scripts/symlink-claude-md.sh , we should include a comment in this script to point to its origin.

johnmtll: Since the link is internal to mongodb, it may be a bit strange to have a reference to it in the repo (OSS contributors cannot follow the link through). I agree with the idea though. Maybe something along the lines of a callout to 'review mongodb internal documentation' for more information on vendoring that script, and another comment not to edit the script directly. I'll workshop something. Thanks!

AGENTS.md
johnmtll: Since this drives agent actions, I think the agent may hallucinate tickets if one cannot be easily identified, leading to rewrites and force pushes--territory developers generally find tedious. For that reason I think it makes sense to specify that the practice is optional. Good call.

scripts/symlink-claude-md.sh
johnmtll: The requirement to run this is only in the case in which you want to add another AGENTS.md somewhere within the project and it is staged. It would be better served as a precommit hook. I've added that.

scripts/symlink-claude-md.sh
PavelSafronov: OSS contributors cannot follow the link, but that's also true of the .evergreen stuff as well: some aspects of the Driver and related libs are just internal, and I think that's fine.

AGENTS.md
```diff
@@ -59,8 +59,7 @@ Use these to validate `perf`-tagged changes — run before and after on the same
 
 ## Commit Messages
 
-[Conventional Commits](https://www.conventionalcommits.org/) with a Jira ticket: `<type>(NODE-XXXX): <subject>` — types: `feat`, `fix`, `docs`, `style`, `refactor`, `perf`, `test`, `chore`; breaking changes use `!` (e.g. `feat(NODE-XXXX)!: …`).
-
+[Conventional Commits](https://www.conventionalcommits.org/) optionally with a Jira ticket: `<type>(NODE-XXXX): <subject>` — types: `feat`, `fix`, `docs`, `style`, `refactor`, `perf`, `test`, `chore`; breaking changes use `!` (e.g. `feat(NODE-XXXX)!: …`). This rule is a *soft requirement*, though it is encouraged.
```
PavelSafronov: I wouldn't call the commit rule a *soft requirement*. The rule is mandatory for PR titles (we even have it part of the template), but not for the individual commits within a PR.

johnmtll: Alright, I'll change the verbiage here. Since this callout is to help agents create commit messages, I think we can leave the PR title portion out. LMK if you're aiming for something different. Thanks again

scripts/symlink-claude-md.sh
johnmtll: Fair. I'll make the callout more explicit and not so roundabout then!

AGENTS.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Other Important Information
+
+- Changes must be compliant with bson spec, and should conform to specification tests for bson (e.g: [bson-corpus](https://github.com/mongodb/specifications/blob/master/source/bson-corpus/bson-corpus.md)).
+- Before making key changes to anything within `src/parser/*`, consider performance implications and run benchmark test passes to ensure no significant performance degradation.
+
+## Commit Messages
+
+[Conventional Commits](https://www.conventionalcommits.org/) optionally with a Jira ticket: `<type>(NODE-XXXX): <subject>` — types: `feat`, `fix`, `docs`, `style`, `refactor`, `perf`, `test`, `chore`; breaking changes use `!` (e.g. `feat(NODE-XXXX)!: …`). Though this rule is optional, it is encouraged.
+## Further Reading
```
PavelSafronov, line 64: Nit: insert new line here

AGENTS.md
```diff
@@ -59,7 +59,7 @@ Use these to validate `perf`-tagged changes — run before and after on the same
 
 ## Commit Messages
 
-[Conventional Commits](https://www.conventionalcommits.org/) optionally with a Jira ticket: `<type>(NODE-XXXX): <subject>` — types: `feat`, `fix`, `docs`, `style`, `refactor`, `perf`, `test`, `chore`; breaking changes use `!` (e.g. `feat(NODE-XXXX)!: …`). This rule is a *soft requirement*, though it is encouraged.
+[Conventional Commits](https://www.conventionalcommits.org/) optionally with a Jira ticket: `<type>(NODE-XXXX): <subject>` — types: `feat`, `fix`, `docs`, `style`, `refactor`, `perf`, `test`, `chore`; breaking changes use `!` (e.g. `feat(NODE-XXXX)!: …`). Though this rule is optional, it is encouraged.
```
PavelSafronov: `Though this rule is optional, it is encouraged.` -> `This rule is mandatory for PR descriptions, because that is what ends up in the history. The individual commits inside a PR do not have to follow this convention, because we squash PR commits.`

etc/prepare.js
seanrmilligan, line 1: Is there a reason why this is not implemented as a `.gitconfig` in the root of the repo? This is the git-native way to specify a per-repo hooks folder.

```
[core]
    hooksPath = .githooks
```

Git tracks if the executable bit is set so we would `chmod` once and commit it instead of looping over the directory each time.

seanrmilligan: (If this has come up across any of the other repos feel free to link the conversation to that one)

seanrmilligan (dismissed):

etc/prepare.js
johnmtll: .gitconfig is ignored by git in the repo root by default: https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration. This means that the .gitconfig needs to be parsed via script and written to .git/config anyway. Since .gitconfig is generally considered to be a sensitive user file, and the only reason we'd want the .gitconfig in the repo root would be to support multiple .git/config mods into the future. I think we should save considering this for when we have more than one configuration to promote to .git/config.

"Git tracks if the executable bit is set so we would chmod once and commit it instead of looping over the directory each time." Good point! I'll fix that.

seanrmilligan: > .gitconfig is ignored by git in the repo root by default

TIL! 🫨

seanrmilligan (approved):

PavelSafronov (approved):

PR #923, opened 2026-08-26
fix(NODE-7612): reject subtype-9 vectors shorter than 2 metadata bytes

AbinMadathil-Celigo:
#### Summary of Changes

`validateBinaryVector()` read `vector.buffer[0]` (datatype) and `vector.buffer[1]` (padding) without first checking that those two bytes exist. A zero-length `Binary` created through the raw constructor — `new Binary(Buffer.alloc(0), Binary.SUBTYPE_VECTOR)` — makes both reads `undefined`, so every dtype-specific validation branch is skipped and the value passes validation. Both `BSON.serialize()` and `EJSON.stringify()` then emit a malformed subtype-9 vector with no dtype/padding metadata.

The typed helper constructors (`fromInt8Array`, `fromFloat32Array`, `fromPackedBits`) always allocate `byteLength + 2`, so the invariant "every subtype-9 value carries 2 metadata bytes" already holds everywhere except the raw constructor path. This PR enforces that invariant in `validateBinaryVector()` by rejecting vectors shorter than 2 bytes, which also closes the 1-byte PackedBit gap (`padding` reads `undefined`, `undefined > 7` is false, so it passed before).

##### Notes for Reviewers

- The existing spec test data (`test/node/specs/bson-binary-vector`) contains no valid sub-2-byte vectors, so no compliance tests are affected.
- The `size !== 0` condition in the existing Float32 check becomes unreachable after this guard; I left it in place to keep the diff minimal, happy to remove it if preferred.

Fixes NODE-7612 (part of NODE-7598).

### Zero-length `Binary` subtype-9 vectors are now rejected

`BSON.serialize()` and `EJSON.stringify()` now throw a `BSONError` for subtype-9 (Vector) values that do not contain the mandatory datatype and padding metadata bytes, instead of emitting malformed BSON.

### DIFF
--- a/src/binary.ts
+++ b/src/binary.ts
@@ -518,6 +518,14 @@ export function validateBinaryVector(vector: Binary): void {
 
   const size = vector.position;
 
+  // Every subtype-9 vector carries two metadata bytes (datatype and padding):
+  // the typed helpers always allocate them, but the raw constructor path may not (NODE-7612)
+  if (size < 2) {
+    throw new BSONError(
+      'Invalid Vector: must be at least 2 bytes to hold the datatype and padding bytes'
+    );
+  }
+
   // NOTE: Validation is only applied to **KNOWN** vector types
   // If a new datatype is introduced, a future version of the library will need to add validation
   const datatype = vector.buffer[0];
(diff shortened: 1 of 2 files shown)


johnmtll:
Thanks for your submission @AbinMadathil-Celigo!

We're going to close this. Accepting subtype-9 binaries without dtype/padding metadata is intentionally supported behavior that predates finalization of the binary vector spec. Data has already crossed the wire in this shape, so for backwards compatibility we need to keep the existing functionality. We appreciate the thorough writeup!

PR #924, opened 2026-08-26
fix(NODE-7611): bounds check findNull scan in on-demand parser

AbinMadathil-Celigo:
#### Summary of Changes

A malformed document with a string element declaring a zero length passes the on-demand parser's size/terminator guards and advances the cursor to the declared document end. The next loop iteration then calls `findNull` from past the end of the buffer: out-of-range `Uint8Array` reads return `undefined` (which is never `0x00`), so the scan never terminates — a synchronous CPU hang on caller-controlled bytes.

`findNull` now stops at the end of the input and throws a `BSONOffsetError('Null terminator not found')` instead of scanning out of range forever. The existing "reached the document terminator" throw is preserved by folding both conditions into `nullTerminatorOffset >= bytes.length - 1`.

Repro (hangs before this change, throws after):

```javascript
const BSON = require('bson');
// 10-byte document: string element with empty name declaring stringSize = 0
BSON.onDemand.parseToElements(
  new Uint8Array([0x0a, 0x00, 0x00, 0x00, 0x02, 0x00, 0x00, 0x00, 0x00, 0x00])
);
```

##### Notes for Reviewers

- Only the experimental `BSON.onDemand.parseToElements` path is affected; the standard `BSON.deserialize` path already rejects `stringSize <= 0`.
- The ticket also notes the `stringSize <= 0` check is not mirrored on the on-demand path. I kept this PR focused on the non-termination bug; mirroring that check would need per-type care (e.g. a zero length is valid for `binData`) and could be a follow-up.

Fixes NODE-7611 (part of NODE-7598).

### `BSON.onDemand.parseToElements` no longer hangs on malformed input

A malformed document containing a string element with a declared length of zero previously caused a non-terminating scan (synchronous CPU hang). The experimental on-demand parser now throws a `BSONOffsetError` instead.

Thanks to @AbinMadathil-Celigo for providing the fix!

### DIFF
--- a/test/node/parser/on_demand/parse_to_elements.test.ts
+++ b/test/node/parser/on_demand/parse_to_elements.test.ts
@@ -64,13 +64,47 @@ describe('parseToElements()', () => {
   context('when a string element declares a zero length (NODE-7611)', () => {
     it('throws instead of scanning past the end of the input', () => {
       // 10-byte document: string element with empty name declares a stringSize of 0,
-      // which advances the cursor to the declared document end and would otherwise
-      // make findNull scan out of range forever.
+      // which would advance the cursor to the declared document end and previously
+      // made findNull scan out of range forever.
       const test = () =>
         parseToElements(
           new Uint8Array([0x0a, 0x00, 0x00, 0x00, 0x02, 0x00, 0x00, 0x00, 0x00, 0x00])
         );
-      expect(test).to.throw(/Null terminator not found/i);
+      expect(test).to.throw(/value reports length larger than document/i);
+      expect(test).to.throw(BSONOffsetError);
+    });
+  });
+
+  context('when an element value extends past the end of an embedded document', () => {
+    it('throws an error indicating the value is larger than the document', () => {
+      const bytes = new Uint8Array([
+        0x17,
+        0x00,
+        0x00,
+        0x00, // outer documentSize
+        0x03,
+        0x61,
+        0x00, // 'a': embedded document
+        0x08,
+        0x00,
+        0x00,
+        0x00, // embedded documentSize
+        0x10,
+        0x62,
+        0x62,
+        0x00, // int32 named 'bb'
+        0x10,
+        0x63,
+        0x00,
+        0x2a,
+        0x00,
+        0x00,
+        0x00, // 'c': int32 sibling of 'a'
+        0x00 // outer terminator
+      ]);
+
+      const test = () => parseToElements(bytes, 7);
+      expect(test).to.throw(/value reports length larger than document/i);
       expect(test).to.throw(BSONOffsetError);
     });
   });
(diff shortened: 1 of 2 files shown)


test/node/parser/on_demand/parse_to_elements.test.ts
```diff
@@ -61,6 +61,20 @@ describe('parseToElements()', () => {
...
+      // make findNull scan out of range forever.
+      const test = () =>
+        parseToElements(
+          new Uint8Array([0x0a, 0x00, 0x00, 0x00, 0x02, 0x00, 0x00, 0x00, 0x00, 0x00])
+        );
+      expect(test).to.throw(/Null terminator not found/i);
+      expect(test).to.throw(BSONOffsetError);
+    });
+  });
+
```
tadjik1, line 111: Could you also add a test for an embedded document? Something like this:
```
context('when an element value extends past the end of an embedded document', () => {
  it('throws an error indicating the value is larger than the document', () => {
    const bytes = new Uint8Array([
      0x17, 0x00, 0x00, 0x00, // outer documentSize
      0x03, 0x61, 0x00, // 'a': embedded document
      0x08, 0x00, 0x00, 0x00, // embedded documentSize
      0x10, 0x62, 0x62, 0x00, // int32 named 'bb'
      0x10, 0x63, 0x00, 0x2a, 0x00, 0x00, 0x00, // 'c': int32 sibling of 'a'
      0x00 // outer terminator
    ]);

    const test = () => parseToElements(bytes, 7);
    expect(test).to.throw(/value reports length larger than document/i);
    expect(test).to.throw(BSONOffsetError);
  });
});
```

It shouldn't hang as well.

tadjik1 (changes requested):
hi @AbinMadathil-Celigo , thanks for your work! Please check one comment in the code, I believe there is another case we want to make sure works as expected.

test/node/parser/on_demand/parse_to_elements.test.ts
AbinMadathil-Celigo: Hi @tadjik1, thanks for the catch! You were right — that case was silently accepted. The length guard only checked length > documentSize, ignoring where the element starts, so a value could extend past its (embedded) document's end.

Fixed in https://github.com/mongodb/js-bson/commit/84ec4bb9f3779c57833eb5d6ab4e71fb8f44bb44: the value must now fit within the current document (offset + length > startOffset + documentSize - 1 throws value reports length larger than document). Added your test case verbatim — it throws as expected. Full suite (4,539 tests) and lint pass.

tadjik1:
Hi @AbinMadathil-Celigo, thanks a lot for taking care of this PR! Unfortunately the team is stretched thin at the moment and that's why there is a delay with the re-review. Sorry about that, I will be able to do it later this week. If all looks good the next step would be additional team member review and then we can merge it.

tadjik1 (approved):
Thank you @AbinMadathil-Celigo! I filled the follow-up ticket https://jira.mongodb.org/browse/NODE-7860 for the part you mentioned in the description (the `stringSize <= 0` check). Moving it into team review.

PavelSafronov (approved):
Looks good, thank you!

PR #926, opened 2026-08-27
fix(NODE-7696): calculateObjectSize hangs indefinitely on an object containing a circular reference

PavelSafronov:
#### Summary of Changes

Add a `path: Set<Document>` similar to [serializer's](https://github.com/mongodb/js-bson/blob/node-7696/src/parser/serializer.ts#L500-L508) so we can detect circular references when calculating BSON size. Also added a formal interface `SizeFrame` for an anonymous object we've been using.

### Fixed `BSON.calculateObjectSize` infinite hang

Fixed a bug where `BSON.calculateObjectSize` would hang indefinitely on an object containing a circular reference. The method now throws an exception when a circular reference is detected.


tadjik1 (approved):
Great work, thanks @PavelSafronov!

PR #936, opened 2026-09-17
fix(NODE-7784): try..catch wrapper around snapshot fn calls in static ObjectID initializer

johnmtll:
#### Summary of Changes

Adds a try..catch wrapper around calls to 2 node optional API's which are not intended to throw. This is to protect against runtimes misimplementing the optional API's with stubs that throw, creating a misleading scenario where import-time throws appear as if something has gone critically-wrong, when it has not.

### Fixed static ObjectID initializer throws at import-time

Guards against throws at import-time in the NextJS Edge & Bun runtimes, which occur when the static ObjectID initializer calls choice optional API's: `getBuiltinModule` and `isBuildingSnapshot`, which are implemented with a throwing stub.


src/objectid.ts
```diff
@@ -59,9 +59,17 @@ export class ObjectId extends BSONValue {
...
     this.resetState();
     // https://nodejs.org/api/v8.html#startup-snapshot-api
-    // @ts-expect-error Node.js types not present since this is an optional API
-    const { startupSnapshot } = globalThis?.process?.getBuiltinModule?.('v8') ?? {};
-    if (startupSnapshot?.isBuildingSnapshot?.()) {
+    let startupSnapshot;
+    let isBuildingSnapshot = false;
+    try {
+      // @ts-expect-error Node.js types not present since this is an optional API
+      const v8Module = globalThis?.process?.getBuiltinModule?.('v8') ?? {};
```
PavelSafronov: AC specifies:
> this try-catch should have a callout to the issue in question above it, so it's known that the try-catch depends on this external runtime bug

AC asks for two different try/catch blocks with two different links to the issues. We can consolidate those into one, but we should still provide links and an explanation for this pattern.

PavelSafronov (changes requested):
PR will need release notes before we can merge it.

src/objectid.ts
johnmtll: I opted to have 2. It just made things slightly easier to read. Let me know if it looks good to you or if you'd prefer consolidation.

Nice callout! Missed that completely. 🙇

addaleax (dismissed):

src/objectid.ts
```diff
@@ -58,10 +58,36 @@ export class ObjectId extends BSONValue {
 
   static {
     this.resetState();
+
+    let v8Module;
+    let isBuildingSnapshot = false;
+    // NODE-7784: getBuiltInModule is wrapped in a try..catch due to
```
PavelSafronov: Nit: there's no reason to mention the NODE item that adds these blocks, let's drop the two references

src/objectid.ts
```diff
@@ -58,10 +58,36 @@ export class ObjectId extends BSONValue {
...
+    // https://github.com/vercel/next.js/issues/98226.
+    //
+    // The NextJS Edge runtime
+    // implements getBuiltInModule with a throwable stub, which we protect
```
PavelSafronov, line 68: Nit: `getBuiltInModule` vs `getBuiltinModule`, go with the method casing

PavelSafronov (approved):

